PR #516, opened 2015-01-23
Add vectorItraversed for vectors with restricted polymorphism

klao:
I'm not sure if I might just missed this, but I wanted a traversal over unboxed vectors a few times now. Also, not sure if this is the right place for this.

It can probably be made more efficient by avoiding conversion to/from lists. (My simple benchmark shows a 5-10 fold slow-down compared to the direct use of `Vector.Unboxed.imap`.)

### DIFF
--- a/src/Data/Vector/Generic/Lens.hs
+++ b/src/Data/Vector/Generic/Lens.hs
@@ -29,6 +29,8 @@ module Data.Vector.Generic.Lens
   -- * Traversal of individual indices
   , ordinals
   , vectorIx
+  -- * Traversal with restricted polymorphism
+  , vectorItraversed
   ) where
 
 import Control.Applicative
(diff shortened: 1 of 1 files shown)


ekmett:
`each` can currently be used for this.

klao:
Thanks! I missed `each`, it's good to know about it!

Two quick questions:

1. `each` could be generic in vector, right? (This is how I missed it, I was grepping for Data.Vector.Generic)
2. Is there an indexed version of each?

But for most purposes `each` is perfect, it's very rare that I need an indexed traversal!

glguy:
My tests (this morning and before I saw this ticket) show that traversing via lists is faster than being clever about trying to construct build up an ST computation to construct a mutable vector and freeze it.

via list was running at 11us for a particular benchmark
via ST closure was taking 24us for the same

Not conclusive, but not promising for a new operation

klao:
> via list was running at 11us for a particular benchmark
> via ST closure was taking 24us for the same

Thanks, Eric! Good to know.

ekmett:
You can also (almost) get the behavior of what you wrote above by

`vector.traversed`

using `vector` from `Data.Vector.Generic.Lens`, but it is currently monomorphic. Probably because we had problems with it before we made the getter/fold operations monomorphic.

klao:
Well, `each` cannot be defined over generic vectors because of overlapping instances. Duh.

ekmett:
FWIW- if you want an indexing version of each that starts counting at 0, `indexing each` will work

klao:
Oh, right thanks!

PR #517, opened 2015-01-23
Make `vector` and `asStream` (more) polymorphic

klao:
A different solution to the same underlying problem as #516.

The advantage here is that `boxed.mapped` incurs no performance penalty compared to using `Data.Vector.Unboxed.map` directly! (At least that's what my simple benchmarks made me believe. The vector magic optimizes it all away, unlike when converting to and back from lists, as `each` and my previous solution does.)

### DIFF
--- a/src/Data/Vector/Generic/Lens.hs
+++ b/src/Data/Vector/Generic/Lens.hs
@@ -79,18 +79,18 @@ toVectorOf l s = fromList (toListOf l s)
 --
 -- >>> Vector.fromList [0,8,15] ^. from vector
 -- [0,8,15]
-vector :: Vector v a => Iso' [a] (v a)
+vector :: (Vector v a, Vector v b) => Iso [a] [b] (v a) (v b)
 vector = iso fromList V.toList
 {-# INLINE vector #-}
 
 -- | Convert a 'Vector' to a finite 'Stream' (or back.)
-asStream :: Vector v a => Iso' (v a) (Stream a)
+asStream :: (Vector v a, Vector v b) => Iso (v a) (v b) (Stream a) (Stream b)
 asStream = iso stream unstream
 {-# INLINE asStream #-}
 
 -- | Convert a 'Vector' to a finite 'Stream' from right to left (or
 -- back.)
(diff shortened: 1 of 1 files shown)


glguy:
The vector package defines

````haskell
convert :: (Vector v a, Vector w a) => v a -> w a
convert = unstream . stream
````

which would probably be nicer to implement `boxed` it in terms of.

My benchmark attempt on a simple case confirms that `Data.Vector.Unboxed.map` and `over (boxed.mapped)` both perform nearly identically and in this limited case do optimize to about the same core.

**Comment by klao** [comment]
Huh, I missed `convert`, nice catch!

I added your suggestion for `converted`.

**Comment by ekmett** [comment]
`from vector.vector`

?

**Comment by glguy** [comment]
I think that the most direct implementation using existing optics is: `asStream . from asStream`

**Comment by ekmett** [comment]
Makes sense. Not really a huge fan of the names of any of the optics in that module -- despite having written them ;) -- but that implementation looks as good as any.

**Comment by klao** [comment]
Yes, `asStream` would be perfect for me if it were polymorphic. For common cases `asStream.mapped` is enough, and it has no performance overhead.

Ed, you mentioned that there was some issue with `vector` being monomorphic in values:

> ... `vector` from `Data.Vector.Generic.Lens`, but it is currently monomorphic. Probably because we had problems with it before we made the getter/fold operations monomorphic.

I looked through the history, but couldn't figure out what was the issue and/or if it's still the case. Do you remember? Because that might really be best solution.

**Comment by klao** [comment]
OK, I am re-purposing this pull request to making the `vector` and `asStream` isos polymorphic in values.

With a change like this it's harder to know if/how it will affect the users, but if there are no hidden pitfalls, it seems to be more useful like this.

**Comment by glguy** [comment]
I think this should be fine; when used for updating and viewing these get reasonable types:

```haskell
over asStream :: (Vector v b, Vector v a) => (Stream a -> Stream b) -> v a -> v b
view asStream :: Vector v a => v a -> Stream a
```

ekmett:
I think its fine to make these polymorphic now. Back when Getter had all 4 parameters you'd get inference problems.

PR #523, opened 2015-02-12
Unconditionally add Zoom ExceptT instance using mtl-compat

bergmark:
This is needed to allow reverse-dependencies of this package to migrate to `ExceptT` while keeping backwards compatibility for older versions of `mtl`.

Since the `MonadState ExceptT` instance is defined in `Control.Monad.State.Class` in `mtl` and `mtl-compat` can't override this we can import `Control.Monad.Except` instead of `Control.Monad.Trans.Except` to always get the instance.

Also see fpco/stackage#439

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -196,7 +196,6 @@ library
     kan-extensions            >= 4.2.1    && < 5,
     exceptions                >= 0.1.1    && < 1,
     mtl                       >= 2.0.1    && < 2.3,
-    mtl-compat                == 0.1.*,
     parallel                  >= 3.1.0.1  && < 3.3,
     primitive                 >= 0.4.0.1  && < 0.6,
     profunctors               >= 4        && < 5,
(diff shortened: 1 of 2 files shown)


ekmett:
`mtl-compat` as it is currently implemented has the same problem that `transformers-compat` did when it was first designed.

It relies on flags to backtrack between versions rather than on having 2-3 different "current" versions with different flags set.

This means that what will happen is the cabal backtracker will work some of the time, but not all of the time and we'll get horribly broken build plans, and then we'll start hitting backtracking timeouts a few packages up from lens, just like we did originally with `transformers-compat`. =(

ekmett:
Another option might be for me to give in and add `mtl` as a dependency for `transformers-compat`. This would enable `transformers-compat` to supply the instances itself, and avoid the very need for an entire package full of orphan instances.

bergmark:
Good catch, I didn't consider that `mtl-compat` might cause the same issue.

I've updated this PR to use `transformers-compat 0.4.*` instead of `mtl-compat`.

bergmark:
Thanks!

PR #528, opened 2015-03-07
fix haddock markup errors

isomorphism:
Haddock's markup handling is kinda fishy sometimes. I think it was choking on the lack of spaces before/after `.` in one case and the fact that the @-delimited text contained more @ symbols in the other.


ekmett:
Thanks!

PR #532, opened 2015-03-29
PrimMonad is redundant for primitive >0.6 if it's constrained by PrimBase

dredozubov:
I was skimming through changes made for GHC 7.10.1/base-4.8.0.0 and i noticed you can leave the PrimMonad out of the signature in `prim`. Just a nitpick, but i'll leave a PR anyway. 

PrimBase is defined like this:
```haskell
class PrimMonad m => PrimBase m where
  -- | Expose the internal structure of the monad
  internal :: ...
```
https://github.com/haskell/primitive/blob/master/Control/Monad/Primitive.hs#L66

btw can someone please say `cabal upload` for lens 4.9?


glguy:
Good catch

PR #535, opened 2015-04-01
4.7 maintenance

snoyberg:
Pinging @glguy. Discussed with @ekmett, this would be really nice to get out right away for the LTS 2 release, it's the only thing holding back primitive-0.6 now. I've just cherry picked the commits over from HEAD. Let me know if you have any questions.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,3 +1,7 @@
+4.7.0.1 [maintenance release]
+---
+* `primitive` 0.6 support
+
 4.7
 ---
 * Migrated `Control.Lens.Action` to `lens-action`.
--- a/lens.cabal
+++ b/lens.cabal
@@ -196,7 +196,7 @@ library
     exceptions                >= 0.1.1    && < 1,
     mtl                       >= 2.0.1    && < 2.3,
     parallel                  >= 3.1.0.1  && < 3.3,
-    primitive                 >= 0.4.0.1  && < 0.6,
+    primitive                 >= 0.4.0.1  && < 0.7,
     profunctors               >= 4        && < 5,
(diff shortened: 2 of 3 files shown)


snoyberg:
I've kicked off the beginning of the LTS 2 build now. That will likely take 3 hours. I'd still prefer to get primitive 0.6 in than not. If you're able to merge this and release before then, I'll restart the LTS build for the newer primitive.

ekmett:
I'm on the fence on this one.

It is technically a major version bump requiring change as the signature changes, so I can empathize with glguy's reservations.

What locks us until lens 4.7 on the platform? 

We pretty much consider that version dead here at `lens` HQ, and we're 99% of the way to shipping 4.9. =/

ekmett:
4.8 can be much much faster, and removes all uses of `unsafeCoerce`, making folks much feel much safer.

http://hackage.haskell.org/package/lens-4.8/changelog

snoyberg:
diagrams holds us back.

But I disagree with it being a major version bump. The API remains identical for all previously compiling use cases. The only change is that a new use case is opened up when compiling against primitive-0.6, in which case the API change reflects the underlying package's API change.

If the answer to this is going to be no, please let me know, and I'll move ahead with the LTS 2 release using primitive 0.5.

snoyberg:
Here's the latest update I did on lens:

https://github.com/fpco/stackage/issues/467#issuecomment-87378359

For the record, this set of packages is a constant source of delayed lens releases to Stackage.

ekmett:
4.7.0.1 is shipped. I don't like it, but I'd rather not hold up `primitive` adoption.

snoyberg:
Awesome, thank you!

PR #539, opened 2015-04-08
Depend on base-compat for orphan Foldable/Traversable instances

RyanGlScott:
At the moment, both the `base-compat` and `lens` libraries expose orphan `Foldable` and `Traversable` instances for `Either`, `Const`, and `(,)`, which means they can't be used together. Since `base-compat` was created for the purpose of backporting new features of `base` (including orphan instances), I feel like it makes more sense for them to be defined in `base-compat`. This commit adds a dependency on `base-compat` and reexports the orphan instances from `Control.Lens.Internal.Instances`.

See also [this pull request](https://github.com/ekmett/semigroupoids/pull/27) for `semigroupoids`.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,6 +1,8 @@
 4.9.1
 -----
 * Added `_Wrapped` support for `NonEmpty`.
+* Added dependency on `base-compat`, which contains orphan `Foldable` and `Traversable` instances for `Either`, `Const`, and `(,)`.
+* The `Control.Lens.Internal.Instances` module now reexports the orphan instances from `base-compat`.
 
 4.9
 -------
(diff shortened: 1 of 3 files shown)


RyanGlScott:
Oops, I forgot that the orphan instances from the current version `semigroupoids` will percolate into `lens` and clash with `base-compat`'s, so the Travis-CI builds won't all work unless `semigroupoids` is updated.

RyanGlScott:
At the very least, @sol and I have agreed that the orphan instances (which is all that `semigroupoids` needs from `base-compat`) should be split off into a separate package (probably named `base-orphans` or something similar), so if we do go through with this, it won't be via this pull request. I'll close for now.

PR #540, opened 2015-04-24
Switch conventions when no fields have a _ prefix

neongreen:
`makeLenses` now adds underscores to lenses if no fields have underscores. (`makeClassy` does the same – so perhaps `makeClassy_` and `classyRules_` can be deprecated now?)

### DIFF
--- a/src/Control/Lens/TH.hs
+++ b/src/Control/Lens/TH.hs
@@ -140,29 +140,32 @@ createClass f r =
 
 -- | 'Lens'' to access the convention for naming fields in our 'LensRules'.
 --
--- Defaults to stripping the _ off of the field name, lowercasing the name, and
--- skipping the field if it doesn't start with an '_'. The field naming rule
--- provides the names of all fields in the type as well as the current field.
--- This extra generality enables field naming conventions that depend on the
--- full set of names in a type.
+-- Defaults to stripping the @_@ off of the field name, lowercasing the first
+-- letter, and skipping the field if it doesn't start with an @_@ (however,
+-- if /no/ fields start with an @_@, it will do the opposite thing and add
+-- @_@ to all of them).
 --
--- The field naming rule has access to the type name, the names of all the field
--- of that type (including the field being named), and the name of the field
--- being named.
+-- The field naming rule has access to the type name, the names of all the
+-- field of that type (including the field being named), and the name of the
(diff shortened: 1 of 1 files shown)


ekmett:
I'm not in a hurry to deprecate them as they are much less magic, so if you want to document that that is your convention I could see using them.

neongreen:
@ekmett: Would it be better to create a separate `makeLenses_`, then, to avoid excessive magic? Then `makeLenses` could remain unchanged.

ekmett:
I think the "magic" `makeLenses` is the right default, but I can understand that other folks might disagree. 

I'm neutral on adding `makeLenses_` in addition though for symmetry. That said, I'm also fairly neutral on just removing the makeFoo_ versions and am happy to be swayed by strong opinion. Given that the latter yields a simpler end-state, that might not be a bad idea.

neongreen:
I added `makeLenses_` and `lensRules_`, because I think some people would want to rely on magic and some people would want their code to be more self-documenting (for instance, if there was only magical `makeLenses`, I can imagine myself wanting to add a comment explaining what's going on – and with `makeLenses_` that might not be necessary).

(Also, somewhat unrelatedly, I exported `classyRulesFor`, because I noticed that it was written and documented but not exported.)

neongreen:
Okay, I see 3 ways forward with this:

  1. export `makeLenses_` and `lensRules_` (i.e. this pull request)
  2. change `makeLensesFor` to take `String -> Maybe String` instead of `[(String, String)]`
  3. leave `makeLensesFor` as it is and create a new `makeLenses*` function

I don't like 1. much, and I don't like 2. because it would require people to write something like this:

~~~ haskell
makeLenses? (flip lookup
    [ ("_barX", "bar")
    , ("_barY", "bar") ]
    ''Bar
~~~

which doesn't look nice. So, it seems to me that the best thing would be either 3. (in which case I've no idea how to name this function), or 2. + flipping argument order of `makeLensesFor` to be able to write

~~~ haskell
makeLensesFor ''Bar $ \n -> lookup n
    [ ("_barX", "bar")
    , ("_barY", "bar") ]
~~~

@glguy, what do you think?

ekmett:
Alternately we could do nothing, and just learn to love the magic taking no new names.

PR #546, opened 2015-05-24
Traversal rewrite rules

cchalmers:
These rewrite rules specialise a traversal when it's used as a (indexed) setter or getter. This can provide a significant performance boost, especially for indexed variants and functions over unboxed containers.

This is done by rewriting the traversal to use `sets` or `isets` when it's used as `ASetter` or `AnIndexedSetter` or the new `foldring` and `ifoldring` (not great names) with a `Getting` or `IndexedGetting`.

`foldring` and `ifoldring` take a `foldr` like function and turns it into a `Fold`. These are useful for defining rewrite rules. I've also redefined `folded` in terms of these because indexed folds perform better without using `indexing`. In my tests the new `folded` performs at least as well as the old version but there may be some issue's I haven't thought of.

Because the rewrite rules only depend on the context the traversal is used, not their arguments/what they're applied to, they have a good fire rate. Rules still fire when they're composed / defined as part of another function.

I've also added benchmarks for common traversals and folds.

### DIFF
--- a/benchmarks/folds.hs
+++ b/benchmarks/folds.hs
@@ -1,0 +1,110 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE BangPatterns #-}
+
+import qualified Data.ByteString as BS
+import qualified Data.HashMap.Lazy as HM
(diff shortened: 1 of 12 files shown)


ion1:
Nice. I wonder how people would feel about “foldingr”?

ekmett:
We may want to supply a flag to turn off these RULES for SafeHaskell consumers

Issue #183, opened 2012-12-07
Move arguments to the right side of the =?


shachaf:
We have definitions like

``` haskell
over l f = runMutator# (l (mutator# f))
```

If we turn this into

``` haskell
over l = \f -> runMutator# (l (mutator# f))
```

GHC can inline it better in a call like

``` haskell
foo :: (a -> b) -> [a] -> [b]
foo = over mapped
```

Since the application is saturated, even if foo doesn't end up getting inlined.

Our strategy in lens has mostly been "inline everything, and let GHC simplify it down to nothing at the use site", so this probably wouldn't cause any harm.

raymoo:
How is 8ae6bec9ff37efc2823b81abca4447b86fea4889?

ekmett:
Looks pretty good. I really do hate let notation, but the only alternative there is

```haskell
case foo of
  mv -> ...
```

ekmett:
Potential issue that may affect this: https://ghc.haskell.org/trac/ghc/ticket/11029#comment:1

glguy:
With our aggressive inlining I think that this is less of an issue. Inlining happens before optimizations. Additionally we aren't using this argument trick to share pattern matching; we're using it to influence the inliner.

I did a test of the code attached to that ticket and an inline pragma on `test1` allows the test to run as quickly as the version with `trace`

ekmett:
Good to hear. =)

PR #551, opened 2015-06-14
Moving arguments to the right in Control.Lens.At

raymoo:
I've tried applying #183 to the instances in Control.Lens.At. One thing I'm unsure about is the translation from guards to if-then-else in cases where the lambda-ized arguments are part of the guard expression.


ekmett:
Could somebody try to concoct a benchmark to see if this ever actually helps?

PR #558, opened 2015-07-08
Add Indexed variants of at, ix, and contains (iat, iix, icontains)

andremarianiello:
Variants have not been added to the respective classes. Instead they were defined in terms of the non-indexed version.

Also, some outdated documentation was changed in the Contains class.


glguy:
Should these new combinators be marked for inlining?

ekmett:
Makes sense.

PR #561, opened 2015-07-09
GHC.Generics related comment and TH related stuff.

pacak:
Comment about Field1/Field2 instances of :*:

HasName instance for VarStrictType (sadly adds FlexibleInstances)

HasType class that provides traversals into types


ekmett:
Merged all save that one instance.

PR #566, opened 2015-07-20
Plated instance for CofreeT.

danidiaz:
I added a Plated instance for CofreeT (there is already one for Cofree).

### DIFF
--- a/src/Control/Lens/Plated.hs
+++ b/src/Control/Lens/Plated.hs
@@ -98,6 +98,7 @@ module Control.Lens.Plated
 
 import Control.Applicative
 import Control.Comonad.Cofree
+import qualified Control.Comonad.Trans.Cofree as CoTrans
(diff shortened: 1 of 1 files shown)


glguy:
I don't use `CofreeT` enough to know if it's reasonable to expect a `Traversable` instance for the comonad `w` (I just don't know), but assuming that it's reasonable we might also consider just using `traverse` in the implementation. `bitraverse pure` seems unnecessary.

    plate f (CoTrans.CofreeT xs) = CoTrans.CofreeT <$> traverse (traverse f) xs

ekmett:
The Traversable cases will look like some form of f-branching stream or environment comonad. They all do the right thing. Using `Traversable` is probably better than `Plated` though.

ekmett:
Anyways, yeah, traverse is better than bitraverse pure here.

PR #567, opened 2015-07-20
Add to (^..) documentation

Icelandjack:
0,1,2 traverses show the result of toListOf incrementally


bitemyapp:
:+1: good thinking.

PR #593, opened 2015-08-13
Make use of Travis' folder cache for build-deps

hvr:
See also https://github.com/hvr/multi-ghc-travis/issues/28


hvr:
seems to work... my work is done... :-)

with empty cache:

foo0

on cache hit:

foo

PR #606, opened 2015-09-30
Make Equality poly-kinded

### LINKED ISSUE #605: Equality could be polykinded
This would be a *mildly* breaking change, unless some new names were added. I doubt it would break any actual code. Is there something wrong with this idea that I'm not seeing?

```haskell
{-# LANGUAGE PolyKinds #-}
-- These kinds will be inferred; they're annotated for clarity.
type Equality (s :: k1) (t :: k2) (a :: k1) (b :: k2) = forall p f . p a (f b) -> p s (f t)
type Equality' s a = Simple Equality s a

simple :: Equality' a a
simple f = f

lessSimple :: Equality a b a b
lessSimple f = f

data Identical :: k1 -> k2 -> k1 -> k2 -> * where
  Identical :: Identical a b a b

-- The potentially-breaking change: use Proxy instead of Identity
type AnEquality s t a b = Identical a (Proxy b) a (Proxy b) -> Identical a (Proxy b) s (Proxy t)
type AnEquality' s a = AnEquality s s a a

runEq :: forall (s :: k1) (t :: k2) (a :: k1) (b :: k2) . AnEquality s t a b -> Identical s t a b
runEq l = case l Identical of Identical -> Identical

substEq :: forall (s :: k1) (t :: k2) (a :: k1) (b :: k2) (r :: *) . AnEquality s t a b -> ((s ~ a, t ~ b) => r) -> r
substEq l = case runEq l of
  Identical -> \r -> r

mapEq :: forall (s :: k1) (t :: k2) (a :: k1) (b :: k2) f . AnEquality s t a b -> f s -> f a
mapEq l r = substEq l r

fromEq :: forall (s :: k1) (t :: k2) (a :: k1) (b :: k2) . AnEquality s t a b -> Equality b a t s
fromEq l = substEq l id
```

treeowl:
Allow `Equality s t a b` to prove `s ~ a` and `t ~ b` whenever
`s` and `a` have the same kind and `t` and `b` have the same kind.
Fixes  #605 

Specifically, change the definition of `Equality` to

```haskell
type Equality (s :: k1) (t :: k2) (a :: k1) (b :: k2) =
    forall (p :: k1 -> k3 -> *) (f :: k2 -> k3) . p a (f b) -> p s (f t)
```
for `GHC >= 7.8` and to
```haskell
type Equality (s :: k1) (t :: k2) (a :: k1) (b :: k2) =
    forall (p :: k1 -> * -> *) (f :: k2 -> *) . p a (f b) -> p s (f t)
```
for `GHC==7.6`,

and the definition of `AnEquality` to

```haskell
type AnEquality s t a b =
    Identical a (Proxy b) a (Proxy b) -> Identical a (Proxy b) s (Proxy t)
```

for all versions (with appropriate kind signatures for 7.6+).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ wip
 cabal.sandbox.config
 .stack-work/
 codex.tags
+.stack-work/
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ wip
 cabal.sandbox.config
 .stack-work/
 codex.tags
-.stack-work/
--- a/src/Control/Lens/Equality.hs
+++ b/src/Control/Lens/Equality.hs
@@ -2,6 +2,9 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE TypeFamilies #-}
+#if __GLASGOW_HASKELL__ >= 708
+{-# LANGUAGE PolyKinds #-}
+#endif
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Control.Lens.Equality
@@ -29,7 +32,11 @@ module Control.Lens.Equality
   ) where
 
 import Control.Lens.Type
-import Data.Functor.Identity
+#if __GLASGOW_HASKELL__ >= 708
+import Data.Proxy (Proxy)
+#else
+import Data.Functor.Identity (Identity)
+#endif
 
 #ifdef HLINT
 {-# ANN module "HLint: ignore Use id" #-}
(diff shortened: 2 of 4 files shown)


treeowl:
An alternative:

```haskell
type Equality (s :: k1) (t :: k2) (a :: k1) (b :: k2) = forall (p :: k1 -> * -> *) (f :: k2 -> *) . p a (f b) -> p s (f t)
```

I'm not sure how to choose between those. Either seems to get the job done. The alternative version does not require as poly-kinded a function, but I don't know that it makes any practical difference, and I kind of like the way the version it the pull request shows clearly exactly which kinds have to be fixed and which don't.

treeowl:
I guess the more poly-kinded one can be applied to more different things, if anyone cares.

treeowl:
Whatever the tradeoffs are, the "alternative" version works with 7.6.3 and the other doesn't; that seems to settle it unless there's a better compatible choice for the kind of the result of `f`.

treeowl:
Is anything else needed here?

PR #609, opened 2015-10-02
Poly2

treeowl:
Let's give it a shot this way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ wip
 cabal.sandbox.config
 .stack-work/
 codex.tags
+.stack-work/
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ wip
 cabal.sandbox.config
 .stack-work/
 codex.tags
-.stack-work/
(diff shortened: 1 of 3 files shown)


ekmett:
We should probably just uniformly use Proxy rather than Identity if we're going to do this, that way the only thing that changes is things upgrade to PolyKinds on 7.6+

treeowl:
That sounds utterly sensible. The only potential wibble of concern I can think of is in 7.6; was there ever a versin of `tagged` for 7.6 that had non-poly-kinded `Proxy`?

treeowl:
Grrrrr! No, no good. Am I missing something obvious, or does this poly-kinded stuff not work so reliably on 7.6?

```haskell
tests/properties.hs:65:47:
    Couldn't match kind `AnyK' against `*'
    Kind incompatibility when matching types:
      p0 :: * -> AnyK -> *
      p :: * -> * -> *
    In the first argument of `isIso', namely `(simple :: Iso' Int Int)'
    In the expression: isIso (simple :: Iso' Int Int)
```

treeowl:
Merged back to first branch, hopefully for good.

PR #612, opened 2015-11-07
Canonicalise Monad instances

hvr:
For the record, `Hs2010ToHs201x` was able to perform these simple
refactorings automatically.


Taneb:
How will this affect the inlining? Is GHC clever enough to transform it to `pure` and then inline accordingly?

glguy:
GHC inlines the source definition (not optimized) and then proceeds to inline the things from that definition that needed to be inlined. This wouldn't be the first place we relied on that behavior.

PR #613, opened 2015-11-10
Added indexed versions of minimumOf etc.

meditans:
These functions are the analogue of `minimumOf`, `minimumByOf`, `maximumOf` and `maximumByOf` which also return the position of the element in the indexed fold.

This could be useful, for example, when trying to modify the minimum element in a fold, when one doesn't want to expose a bidirectional optics for the bad compositional properties, eg `modifyMin f . modifyMin g /= modifyMin (f . g)`.

### DIFF
--- a/src/Control/Lens/Fold.hs
+++ b/src/Control/Lens/Fold.hs
@@ -95,6 +95,8 @@ module Control.Lens.Fold
   , firstOf, lastOf
   , maximumOf, minimumOf
   , maximumByOf, minimumByOf
+  , imaximumOf, iminimumOf
+  , imaximumByOf, iminimumByOf
   , findOf
   , findMOf
   , foldrOf', foldlOf'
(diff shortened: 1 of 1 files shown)


ekmett:
I'd usually expect an

```haskell
iminimumByOf :: IndexedGetting i (Endo (Maybe (i,a))) s a -> (a -> a -> Ordering) -> s -> Maybe (i,a)
```

to have a signature more like

```haskell
iminimumByOf :: IndexedGetting i (Endo (Maybe (i,a))) s a -> (i -> a -> a -> Ordering) -> s -> Maybe (i,a)
```

all of the other `ifoo` variants give the index to the helper function.

drquicksilver:
maybe the right name is `minimumWithIndexByOf`?

meditans:
@ekmett if the function is used for comparison shouldn't it be `i -> i -> a -> a -> Ordering`, or something similar (id est, using also the second index for comparison)?

ekmett:
@meditans Indeed.

meditans:
@ekmett updated

ekmett:
Does anybody else want to weigh in here?

glguy:
I think these functions are mostly unnecessary and will see very little use. They don't directly correspond to an existing Prelude function like a lot of the `Of` generalizations in the lens library and a reader will inevitably have to consult the lens documentation to understand what's going on. For something this specialized I'd prefer to see a local definition so that the reader could see what was happening directly.

The library certainly has other confusing definitions in it, but not necessarily in the `Of` family of things.

ekmett:
True, plus even here there is a bit of an open design landscape. e.g `foldMap` versions might be more efficient.

drquicksilver:
I think @ekmett is right to suggest that under the name `iminimumByOf` it should logically have the type which involves the parameter `(i -> a -> a -> Ordering)`, and I am not convinced that will see a lot of use.

However I think that the original type signatures - with a different set of names, probably `minimumWithIndex`, `minimumWithIndexOf` etc - are useful. Note that `Data.Map`, arguably the standard container with the most interesting indexing behaviour has `findMin` returning the index as well as the value, If you have a useful/non-trivial index then it is quite normal to want the index as well as the value of something you have searched for

drquicksilver:
Haha my example is invalid. That is of course key-minimum not value-minimum. Sorry!

ekmett:
I think I'm going to bow to @glguy's intuition here on the relative usefulness, and my own about concerns about multiple available implementation strategies and nearby points in the design space and decline to merge this.

Thank you, everyone for all of your efforts here.

meditans:
Hi all, I've been a bit busy these days, and I'd still like to complete the analysis, regardless of the merging.

I think the real usecase here is where you want to modify the minimum element in a structure. You better not make a lens, for the composability problems I hinted before. So the only viable route remains to be semantically honest: learn where the minimum is and then modify it as you please. Moreover, the various possibilities granted by using the combinators (`<.>`, `<.`, `.>`) when talking about nested containers, make a nice language for locating elements in nested data structures.

It is true, as @glguy suggested, that these functions don't correspond to any `Prelude` function, but I think that's because the `Prelude` doesn't export utilities for working with the indices of the containers. In the same vein `iany`, in `Control.Lens.Indexed`, doesn't correspond per se to any prelude function, but it's there because `any` does, and now we are caring about indexes, too. Moreover, the the similarity with `minimumOf` should clear most confusion for new users on what is going on. That said, it's probably true that a prefixed `i` suggests someting along the type signature `i -> a -> a`, as @ekmett said. In fact, I now like more @drquicksilver's names suggestion (`minimumWithIndexOf` etc).

Regarding the the design variations @kmett suggested, I find that the implementation of, eg, `minimumByWithIndexOf`, which would be (with some strictness added):

```haskell
minimumByWithIndexOf :: IndexedGetting i (Endo (Maybe (i,a))) s a -> (i -> a -> i -> a -> Ordering) -> s -> Maybe (i,a)
minimumByWithIndexOf l cmp = ifoldrOf' l step Nothing where
  step i a Nothing      = Just $! (i,a)
  step i a (Just (j,b)) = Just $! (if cmp i a j b == GT then (j,b) else (i,a))
```

is not too different from the `minimumByOf`, which is:

```haskell
minimumByOf :: Getting (Endo (Endo (Maybe a))) s a -> (a -> a -> Ordering) -> s -> Maybe a
minimumByOf l cmp = foldlOf' l mf Nothing where
  mf Nothing y = Just $! y
  mf (Just x) y = Just $! if cmp x y == GT then y else x
```

so, a possible improvement in the design of the new functions should involve also the function that are already in lens; keeping them together could mean a better control on both implementation. But maybe I understood poorly what the issue is, here, so please explain if that's the case. 

What motivated me to write these function here is that an user acquainted with the indexed part of lens will notice that the `Prelude` functions come also in the indexed version here in `lens`; then, in a usecase like the one I mentioned before, he will wonder why `minimum` doesn't follow that pattern, and eventually write some sort of variation of the code above. 

This ended up longer then I expected, thank you all for your time and keep up the good work :)

ekmett:
@meditans: `minimumOf` would compare values, not keys. `iminimumOf` would still compare values, not keys. None of the ifoo's modify the result type, just tweak the input functions to take an extra `i` from the elements involved. Next, maximum has the annoying property of being partial, which we decided (against pattern) to fix with `minimumOf`.

So assuming we went to `iminimumByOf` I'd expect the signature to be something like `(i -> a -> i -> a) -> IndexedGetting i ... s a -> s -> Maybe a` and not to return the pair.

On the other hand, some sort of `argMinimumOf` would likely give back the `Maybe i`, etc.

There isn't a very good way to fit the existing combinator you've designed above into our existing naming conventions and once they start getting as verbose as @drquicksilver's naming suggestion and as rarely used as @glguy suggested, and as non-canonical as I worried above, I think leaving them out is the right call.

meditans:
@ekmett thanks, now I understand better the naming pattern in the indexed functions, and agree with you. Is it worth to write the `argMinimumOf` combinators, or are they still too marginal in your opinion? What would be the signature?

ekmett:
I think I'd just let them go for now.

Issue #618, opened 2015-12-08
Add Bifunctor analogues of `lmapping` and `rmapping`


treeowl:
There was a [StackOverflow question](http://stackoverflow.com/questions/34154211/lifting-an-iso-into-the-first-argument-of-a-bifunctor) about implementing

```haskell
firsting
  :: Bifunctor f
  => AnIso s t a b
  -> Iso (f s x) (f t y) (f a x) (f b y)
```

This is easily done with

```haskell
firsting p = bimapping p id
```

similarly,

```haskell
seconding = bimapping id
```

However, these do not take advantage of any `Bifunctor` instances' optimized `first` or `second` implementations. To do so, we'd need

```haskell
firsting p = withIso p $ \ sa bt ->
             iso (first sa) (first bt)

seconding q = withIso q $ \sa bt ->
             iso (second sa) (second bt)
```

I suspect these are just enough trouble to be worth adding to `Control.Lens.Iso`.

ekmett:
Merged.

PR #619, opened 2015-12-09
Add firsting and seconding

treeowl:
Add `firsting` and `seconding` for `Bifunctor` to match
`Profunctor`'s `lmapping` and `rmapping`. This allows use
of potentially optimized `first` and `second` implementations,
whereas `firsting p = bimapping p id` and
`seconding = bimapping id` each use `bimap`.


treeowl:
Fixes #618

Icelandjack:
It follows a pattern but I have my reservations about `firsting`...

ekmett:
The naming sucks, but what can you do?

PR #622, opened 2015-12-28
Equality.hs: minore reduction

Heather:


ekmett:
This actually doesn't work on older GHCs. Reverting.

PR #623, opened 2016-01-02
Control.Lens.TH: more documentation and exports

debug-ito:
This patch is mainly for `Control.Lens.TH`.

- Introduce subsections to the document.
- Add and export type-synonyms `FieldNamer` and `ClassyNamer`.
- Export the namers that were hidden.
- Add and export `mappingNamer`.

I just wanted `mappingNamer` to configure `LensRules` easily (See also: http://stackoverflow.com/questions/17132514/generating-lenses-for-a-lens-library-with-a-custom-name-processor-instead-of-t ) However, I think exporting namers helps users understand how LensRules work.

### DIFF
--- a/src/Control/Lens/Internal/FieldTH.hs
+++ b/src/Control/Lens/Internal/FieldTH.hs
@@ -541,7 +541,7 @@ applyTypeSubst sub = rewrite aux
 -- Field generation parameters
 ------------------------------------------------------------------------
 
-
+-- | Rules to construct lenses for data fields.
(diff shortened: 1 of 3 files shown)


ekmett:
@glguy: Thoughts? This is more your end of the sandbox than mine.

glguy:
This seems like an improvement. Additionally, having the `mappingNamer` is going to be handy for people to avoid having to deal with TH `Name` values.

debug-ito:
Thank you!

Issue #545, opened 2015-05-24
Combinator to apply an `Iso` to a `Lens`


ocharles:
```
1:15 PM <ocharles_> is there a way to write '(_zy . iso (over _x negate) (over _x negate))' using negated?
1:15 PM <ocharles_> I want a lens into V2 a from a V3 a by taking the negated z and normal y components. the above works, but I'm sure there's a better way to write it
1:22 PM <ocharles_> It's a bit like 'mapping', but I only want to apply it one thing, rather than all things under the functor
1:23 PM <ocharles_> I guess _zy . involuted (over _x negate) isn't so bad
2:34 PM <xplat> hm, a combinator to expand an iso using a lens can exist, therefore it should ...
```

fumieval:
```haskell
altering :: ASetter s t a b -> AnIso a b b a -> Iso s t t s
altering l i = withIso i $ \f g -> iso (over l f) (over l g)
```

I couldn't make this more general...

ekmett:
I think I'm going to reject this one. The type we can get out in greatest generality isn't that great, so the benefit of having it vs. the cognitive overhead of tracking its limitations doesn't really weigh out in its favor.

PR #639, opened 2016-03-08
Add chimping and chimping'

fumieval:
This addresses #545.


glguy:
An alternative solution to #545 might be to close it. It doesn't seem obvious that naming this operation in the lens package is better than locally defining it nearer to the use site.

fumieval:
I think it makes sense to have a variant of `mapping` that takes a Setter as a generalized fmap. But there are few combinator that does this kind of generalization so it may be too hasty

ekmett:
Going to take @glguy's suggestion here as I commented in #545

PR #642, opened 2016-04-01
Arg type prisms

duog:
Hi,

This PR expands the capability of prism-generating template haskell. Constructors that take multiple arguments have data types for those arguments generated, and the prisms for those constructors target the generated data types instead of tuples. The generated data types also have lenses generated.  

Two new functions, declareArgTypePrisms and makeArgTypePrisms, are added that take a Parameter of type ``Name -> Maybe (Name, LensRules)``.  This function is applied to each constructor name. A result of Nothing will cause tuple prisms to be generated as usual, while a result of Just (n, lr) will cause a data type n to be generated, and lenses on that data type to be generated using lr.

e.g.
```haskell
data A = A1 {_a :: Int, _b :: Int } | A2 {_c :: Int, _d :: Int}
makeArgTypePrisms argTypesRulesSelector ''A
```

Will generate:
```haskell
data A1Args = A1Args Int Int
data A2Args = A2Args Int Int
_A1 :: Prism' A A1Args
a :: Lens' A1Args Int
b :: Lens' A1Args Int
_A2 :: Prism' A A2Args
c :: Lens' A2Args Int
d :: Lens' A2Args Int
```

I have added some tests to templates.hs to exercise the new functionality.
I'm not confident that the names of chosen are any good.

Please let me know if you are interested in merging this.


ekmett:
I confess to being a bit conflicted about this.

duog:
I did this work when I was working with the ghc api, and it's plethora of data types with many constructors of many arguments. In general I don't find the existing prisms particularly useful for constuctors with more than 2 or 3 arguments, and these arg type prisms were useful then.

I wasn't able to find a way to build this without copying a lot of the internal lens code

Feel free to close if you're not interested, or I'm happy to take direction on helping to turn this into something acceptable.

PR #644, opened 2016-04-09
Documentation: Getters mappend into a Fold

hlian:
This pops up intermittently on Stack Overflow and I thought it'd be nice to enshrine the answer into the `Getter` module docs.


ekmett:
Technically you can glue multiple folds together into a fold, not just getters. It may be better to expand the language to show that this is simply a special case of a more general concept.

hlian:
👍 Good idea. Language expanded.

PR #645, opened 2016-04-11
Free Zoom

scshunt:
This adds an instance of Zoom for the free monad transformer. There are a few things of note:

This removes the `Zoomed m ~ Zoomed n` constraint. It is a nice thing to have in theory, but for a recursive instance like this, it can't be satisfied, or at least can't be satisfied without some gymnastics I haven't worked out.

The behaviour could get kind of unintuitive when applied to multiple layers of free monad. A single value of type `FreeT f m a` can have many different `m (...)` values in it, and these will each get zoomed separately. While this is totally find with a `Lens`, it could cause weirdness for a `Traversal` if you expect it to coalesce all the values into one.

I haven't really thought about the behaviour of the `Traversal` here and/or whether it makes sense, I've just defined the obvious instance. It's quite possible that it doesn't make sense and that there is no sensible way to implement it, or perhaps there is a sensible way to implement it and this boring one wasn't it. If this is an issue, it's easy enough to remove the `Apply` and `Applicative` instances so that you can only zoom on it with a proper `Lens`.

I haven't touched Magnify, primarily because I have yaks of my own to shave.

### DIFF
--- a/src/Control/Lens/Internal/Zoom.hs
+++ b/src/Control/Lens/Internal/Zoom.hs
@@ -232,6 +232,20 @@ instance Applicative (k (Err e s)) => Applicative (FocusingErr e k s) where
 -- 'Control.Monad.Trans.FreeT'
 newtype FocusingFree f m k s a = FocusingFree { unfocusingFree :: k (FreeF f s (FreeT f m s)) a }
 
+instance Functor (k (FreeF f s (FreeT f m s))) => Functor (FocusingFree f m k s) where
+  fmap f (FocusingFree as) = FocusingFree (fmap f as)
+  {-# INLINE fmap #-}
+
+instance Apply (k (FreeF f s (FreeT f m s))) => Apply (FocusingFree f m k s) where
+  FocusingFree kf <.> FocusingFree ka = FocusingFree (kf <.> ka)
+  {-# INLINE (<.>) #-}
+
+instance Applicative (k (FreeF f s (FreeT f m s))) => Applicative (FocusingFree f m k s) where
+  pure = FocusingFree . pure
+  {-# INLINE pure #-}
+  FocusingFree kf <*> FocusingFree ka = FocusingFree (kf <*> ka)
+  {-# INLINE (<*>) #-}
+
 -----------------------------------------------------------------------------
 --- Effect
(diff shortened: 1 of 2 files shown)


ekmett:
I'm not 100% sure abound the Applicative instance, but we'll see how it goes.

glguy:
These changes fail to build on older GHCs. @scshunt Do you have time to fix this or should I revert these changes for now?

ekmett:
Is it just a matter of importing Control.Applicative conditionally?

glguy:
No, there was an error about something not being a Functor and another about an unsafe import. I already fixed the hlint error, there might be more.

scshunt:
I don't have time to fix it right now, but if you could paste a build log, that would be appreciated and I'll try to look at it tonight. The unsafe stuff @ekmett noticed, but I got the impression he was going to fix that in free?

glguy:
The build logs are available on travis and are linked in the little red "X" icons next to the commit in this PR

ekmett:
@scshunt I gave you commit access to `free` to chase it down. ;)

scshunt:
Ahh, ok. I'll try to get to it tonight.

scshunt:
Looks like the conflict with `Functor` should be resolved by using `liftM` rather than `fmap`. Not sure about the safe haskell thing.

scshunt:
Do I need to implement free an trustworthy flags in `free`, as is done in `lens`?

ekmett:
That would probably be the best solution, the trick is finding precisely the imports that trigger unsafety and what compiler and import versions screw everything up.

scshunt:
Ok. I'll try to do something and get Travis to test it, then you can tell
me how I'm horribly wrong.

Sean

ekmett:
Travis usually does a pretty good job of catching mistakes of this sort.

RyanGlScott:
I attempted to fix the aforementioned problems with [these](https://github.com/ekmett/lens/commit/e26c84f1f832c577d296da7cb685471569c083bd) two [commits](https://github.com/ekmett/lens/commit/b28e8cccdeec8f24bb83b153bfd0e1e227758d8b). Feel free to revise.

PR #648, opened 2016-04-18
ghc 8.0 TH changes and warning flag changes

cartazio:
I believe these are all the changes needed for lens master to be clean building on ghc 8.0 tip as of today onwards

@ekmett  @hvr  @glguy  @bgamari

### DIFF
--- a/src/Language/Haskell/TH/Lens.hs
+++ b/src/Language/Haskell/TH/Lens.hs
@@ -125,6 +125,13 @@ module Language.Haskell.TH.Lens
   , _RecGadtC
 #endif
 #if MIN_VERSION_template_haskell(2,11,0)
+  -- ** Overlap Prisms
+  ,_Overlappable
+  ,_Overlapping
+  ,_Overlaps
+  ,_Incoherent
+#endif
+#if MIN_VERSION_template_haskell(2,11,0)
   -- ** SourceUnpackedness Prisms
   , _NoSourceUnpackedness
   , _SourceNoUnpack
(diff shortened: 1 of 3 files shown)


cartazio:
i'm not sure if the CPP macros expose enough info to write this in a way that could support the release candidates that have come out so far at the same time as the tip of ghc-8.0 :(

cartazio:
nb: the build failure seems to indicate that the current hvr ppa for 8.0.1 isn't the current tip of that branch

hvr:
@cartazio yeah, @thoughtpolice suggested to keep the 8.0.1 package at RC3 state until RC4 is cut, that's why I'm holding back updating the 8.0.1 package.

RyanGlScott:
@cartazio, can you add `Prism`s for `Overlap` in `Language.Haskell.TH.Lens` while you're at it?

cartazio:
Good idea

phadej:
I made a revisions for `lens-4.13.1` and `lens-4.13.2` to restrict `template-haskell-2.11`, I didn't touch `lens-4.13.2.1` as someone is probably is using older `ghc-8.0.1` RC and the package still works for them.

Please remember to make a revision after this is merged & released.

http://matrix.hackage.haskell.org/package/lens

hvr:
fwiw, I restarted https://travis-ci.org/ekmett/lens/jobs/123983821 and now it passes

phadej:
Now this is broken for `!= 8.0.*` :)

src/Language/Haskell/TH/Lens.hs
```diff
@@ -848,12 +848,21 @@ _ClassD
       remitter (ClassD x y z w u) = Just (x, y, z, w, u)
       remitter _ = Nothing
 
-_InstanceD :: Prism' Dec (Cxt, Type, [Dec])
+#if MIN_VERSION_template_haskell(2,11,0)
+_InstanceD :: Prism' Dec (Maybe Overlap,Cxt, Type, [Dec])
+#else
+_InstanceD :: Prism' Dec (Maybe Overlap,Cxt, Type, [Dec])
```
hvr:
```
src/Language/Haskell/TH/Lens.hs:854:33:
    Not in scope: type constructor or class ‘Overlap’
```

cartazio:
I'll do the cleanup this morning.   Thanks for being patient :)

src/Language/Haskell/TH/Lens.hs
cartazio: Derp, good catch

cartazio:
@RyanGlScott  what do you want the Prism to be between?  Just exposing the Sum Structure of the `Overlaps` data type?
```haskell
-- | Varieties of allowed instance overlap.
data Overlap = Overlappable   -- ^ May be overlapped by more specific instances
             | Overlapping    -- ^ May overlap a more general instance
             | Overlaps       -- ^ Both 'Overlapping' and 'Overlappable'
             | Incoherent     -- ^ Both 'Overlappable' and 'Overlappable', and
                              -- pick an arbitrary one if multiple choices are
                              -- available.
  deriving( Show, Eq, Ord, Data, Typeable, Generic )
```
or what?

src/Language/Haskell/TH/Lens.hs
cartazio: should be fixed now i hope :)

RyanGlScott:
The convention in `Language.Haskell.TH.Lens` is to create prisms of type `Prism Foo ()` for datatypes `Foo` which have all nullary constructors. [Here's an example](https://github.com/ekmett/lens/blob/b28e8cccdeec8f24bb83b153bfd0e1e227758d8b/src/Language/Haskell/TH/Lens.hs#L1272-L1294) for `data Safety = Safe | Unsafe | Interruptible`.

cartazio:
@RyanGlScott should work now :)

RyanGlScott:
Thanks, @cartazio!

Issue #657, opened 2016-05-25
Restore Generics.Deriving.Lens


RyanGlScott:
A couple of things in `lens` pertaining to GHC generics are less than satisfactory at the moment:
- @Icelandjack [points out](https://ghc.haskell.org/trac/ghc/ticket/10512#comment:12) that the [doctests for `tinplate`](https://github.com/ekmett/lens/blob/8fb89aeaec0be52c042ed971cb5793b45b818ce5/src/Generics/Deriving/Lens.hs#L73-L78) no longer compile since the `Generic Char` instance was removed in 8.0 (for good reason!) After discussing with @glguy, it seems that the `Generic.Deriving.Lens` module is not recommended for general use anyway, so perhaps it ought to be removed.
- The `GHC.Generics.Lens` module doesn't contain `Iso`s for the new representation types for unlifted types (the [`URec` data family](http://hackage.haskell.org/package/base-4.9.0.0/docs/GHC-Generics.html#t:URec)) introduced in GHC 8.0. We'd need to box their underlying values, incurring a slight performance penalty, but `lens` programmers might not mind this.
  
  If we wanted that code to be backwards-compatible with old versions of GHC, we'd need to depend on the `generic-deriving` library. Interestingly, the [documentation for `GHC.Generics.Lens`](https://github.com/ekmett/lens/blob/8fb89aeaec0be52c042ed971cb5793b45b818ce5/src/GHC/Generics/Lens.hs#L27-29) claims `lens` already depends on `generic-deriving`, though this is not currently the case. I suppose we could sneakily retcon our dependencies so that this is still truthful :)

RyanGlScott:
Officially, we've "fixed" both of the items above via #658 and #659. However, @ekmett indicated in https://github.com/ekmett/lens/pull/659#issuecomment-230505985 that removing `Generic.Deriving.Lens` was a step too far, and that he'd like to see it revived in a different form, so I'm leaving this issue up.

PR #659, opened 2016-06-03
Remove Generics.Deriving.Lens

RyanGlScott:
`Generics.Deriving.Lens` was never really recommended for general use, and now it's messing up `doctests` on GHC 8.0, so it's high time to remove it. This also reenables `doctest` when build with GHC 8.0.

Fixes one part of #657 (#658 fixes the other part).


ekmett:
I'm kind of sad to see these combinators go. They may not perform as well as the `Data` variants, but they respect the 'intended' structure much better. 

Can we reimplement them as part of `GHC.Generics.Lens`?

RyanGlScott:
My objection to `Generics.Deriving.Lens` wasn't about performance, but rather correctness. The example [given in the `doctests`](https://github.com/ekmett/lens/blob/8fb89aeaec0be52c042ed971cb5793b45b818ce5/src/Generics/Deriving/Lens.hs#L73-L78), for instance, relied on `Char` having a `Generic` instance, which isn't going to work post-`base-4.9` (as well it shouldn't, since the ["leaves" of a datatype](https://ghc.haskell.org/trac/ghc/ticket/10512#comment:3) shouldn't have to be `Generic` instances).

Is it just a matter of the example being faulty, or do the combinators need to be rewritten?

glguy:
That module was a weird mix of generics and typeable. If we could find a
properly typed approach I'd be in favor of the module. The previous
implementation relying on both typeable and Generic leaves wasn't fitting
into the shape of a typical Generics approach, though.

ekmett:
For many use-cases the thing it does is the more 'correct' thing, even if it is badly documented. It is just different than what Data does. I'm more than happy to have us document clearly what the differences are in behavior, though. What it says definitely isn't sufficiently explanatory. This is often much better for things like removing 'one layer' of container from a thing.

ekmett:
That said, the current implementation is somewhat boring. I'd hoped to eventually get around to computing a sort of hit-map like the Data.Lens code does, which would enable this to drastically improve performance.

ekmett:
(That was what I was originally addressing with the issue.)

PR #660, opened 2016-06-05
using Data.Coerce and GHC.Exts needs to be trusted

int-e:
`Control.Lens.Iso` uses `Data.Coerce` and `GHC.Exts` (for `IsList`) so it's not inferred safe, but still trustworthy.

### DIFF
--- a/src/Control/Lens/Iso.hs
+++ b/src/Control/Lens/Iso.hs
@@ -13,13 +13,7 @@
 #define MIN_VERSION_bytestring(x,y,z) 1
 #endif
 
-#ifndef MIN_VERSION_profunctors
-#define MIN_VERSION_profunctors(x,y,z) 1
-#endif
-
-#if __GLASGOW_HASKELL__ < 708 || !(MIN_VERSION_profunctors(4,4,0))
 {-# LANGUAGE Trustworthy #-}
-#endif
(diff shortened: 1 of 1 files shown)


glguy:
It looks like we don't need the Data.Coerce import on 7.8. Does that affect the need to mark this module TRUSTWORTHY on 7.8?

https://github.com/ekmett/lens/tree/import-fix

int-e:
I'm confused... `coerce` is used in the definition of `auf`, so where is it coming from with ghc-7.8? Will have a look...

glguy:
import Control.Lens.Internal.Coerce

int-e:
Okay, then the import of `Data.Coerce` should never be needed, but we need the `Trustworthy` pragma anyway.

int-e:
Sorry, my last comment here is wrong, because `Control.Lens.Internal.Coerce` is itself `Trustworthy` `#if (MIN_VERSION_profunctors(4,4,0)) && __GLASGOW_HASKELL__ >= 708`. I'll have to think about whether that's okay.

But it means that currently we only need the `Trustworthy` pragma `#if __GLASGOW_HASKELL__ < 708 || !(MIN_VERSION_profunctors(4,4,0))`, which is precisely the check that I removed.

PR #663, opened 2016-07-07
Loosen GHC version bound for coerced

treeowl:
GHC has had `Coercible` since version 7.8.


treeowl:
Why does 7.8 need more manipulations to work? I don't understand that.

treeowl:
The failure for HEAD does not look like my fault.

ekmett:
(Note: The reason why it needs a manual flip in GHC 7.8 is that `Coercible a b` doesn't automatically imply `Coercible b a` in that version. This was a bug I filed with GHC HQ that was fixed in 7.10.

PR #665, opened 2016-07-20
Add (?.) operator

cblp:
Shortcut for `. _Just .` in a chain of optics.

Writing and semantics resembles "null-safe" access operator in languages such as Kotlin or Groovy.

```haskell
data Event = Event { _theDay :: Maybe Day {- ^ event may have no date scheduled -}, ... }
makeLenses ''Event

>>> Event Nothing ^? theDay ?. year
Nothing
>>> Event (Just $ fromGregorian 2004 5 6) ^? theDay ?. year
Just 2004
```

### DIFF
--- a/src/Control/Lens/Prism.hs
+++ b/src/Control/Lens/Prism.hs
@@ -45,6 +45,7 @@ module Control.Lens.Prism
   , _Show
   , only
   , nearly
+  , (?.)
   -- * Prismatic profunctors
   , Choice(..)
   ) where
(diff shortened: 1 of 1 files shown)


cblp:
I don't know which type should be. What about `(?.) :: Traversal s t (Maybe a) (Maybe b) -> Traversal a b c d -> Traversal s t c d`?

ekmett:
This doesn't really fit into the pattern of any of our other operators. Given the amount of crap we get for every operator we add, the shortness of the implementation and the Fairbairn threshold, I'm rather inclined to say you should define this locally in your project.

xplat:
actually this kind of fits into a similar pattern as the operators for composing indexed optics ...

ekmett:
Clearly the operator should just be `?`. Er `?`, er `?`

PR #666, opened 2016-07-26
Add *WithIndex ZipList instances

phadej:


RyanGlScott:
Hm, that [`hlint` test error](https://travis-ci.org/ekmett/lens/jobs/147531493#L1003) is really odd... I don't remember that happening in previous builds. Might it be a regression introduced in [`haskell-src-exts-1.18.0`](http://hackage.haskell.org/package/haskell-src-exts-1.18.0)?

RyanGlScott:
OK, building with `haskell-src-exts-1.18.1` at least fixes the issue with `#.` being parsed. However, a new issue [has sprung up](https://travis-ci.org/ekmett/lens/jobs/147531493#L975):

```
src/Control/Lens/Reified.hs:128:6: Error: Parse error: >
Found:
      m <@ _ = m
      {-# INLINE (<@) #-}
  >   _ @> m = m
      {-# INLINE (@>) #-}
```

cc @mpickering

mpickering:
I just released a new version which should fix this problem.

PR #671, opened 2016-08-22
Added documentation of the Iso laws

benjamin-hodgson:
Per [a discussion on Reddit](https://www.reddit.com/r/haskell/comments/4yzima/applied_traversals_and_isos_a_2048_clone/), I added a clarification about the `Iso` laws.


ekmett:
You don't even need the `view` around it.

```haskell
f . from f = id
from f . f = id
```

suffices.

benjamin-hodgson:
@ekmett Thanks for the feedback. I've updated the branch

ekmett:
Merged. Thanks!

PR #673, opened 2016-08-31
Use more efficient functions from containers-0.5.8.1 whenever possible

### LINKED ISSUE #629: Improve TraversableWithIndex for Data.Sequence
It's not possible to do anything about this here yet, but assuming https://github.com/haskell/containers/pull/179 is accepted, the `TraversableWithIndex Int Seq` instance should be changed to `itraverse = Data.Sequence.traverseWithIndex`.

RyanGlScott:
With `containers-0.5.8.1` or later, we can use the more efficient `traverseWithIndex` function for the `TraversableWithIndex` instance for `Seq`.

cc'ing @treeowl in case there are other optimizations of this ilk I overlooked.

Fixes #629.

### DIFF
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -21,6 +21,10 @@
 #ifndef MIN_VERSION_base
 #define MIN_VERSION_base(x,y,z) 1
 #endif
+
+#ifndef MIN_VERSION_containers
+#define MIN_VERSION_containers(x,y,z) 1
+#endif
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Control.Lens.At
@@ -454,17 +458,25 @@ instance At (Maybe a) where
   {-# INLINE at #-}
 
 instance At (IntMap a) where
+#if MIN_VERSION_containers(0,5,8)
+  at k f m = IntMap.alterF f k m
+#else
   at k f m = f mv <&> \r -> case r of
     Nothing -> maybe m (const (IntMap.delete k m)) mv
     Just v' -> IntMap.insert k v' m
     where mv = IntMap.lookup k m
+#endif
   {-# INLINE at #-}
 
 instance Ord k => At (Map k a) where
+#if MIN_VERSION_containers(0,5,8)
+  at k f m = Map.alterF f k m
+#else
   at k f m = f mv <&> \r -> case r of
     Nothing -> maybe m (const (Map.delete k m)) mv
     Just v' -> Map.insert k v' m
     where mv = Map.lookup k m
+#endif
   {-# INLINE at #-}
 
 instance (Eq k, Hashable k) => At (HashMap k a) where
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -459,7 +459,7 @@ instance At (Maybe a) where
 
 instance At (IntMap a) where
 #if MIN_VERSION_containers(0,5,8)
-  at k f m = IntMap.alterF f k m
+  at k f = IntMap.alterF f k
 #else
   at k f m = f mv <&> \r -> case r of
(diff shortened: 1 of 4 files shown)


treeowl:
Yes, we now have Data.Sequence.foldMapWithIndex (ifoldMap) and
Data.Map.Lazy.alterF (at). Some flipping required.

src/Control/Lens/At.hs
```diff
@@ -454,17 +458,25 @@ instance At (Maybe a) where
   {-# INLINE at #-}
 
 instance At (IntMap a) where
+#if MIN_VERSION_containers(0,5,8)
```
treeowl: I'm not sure there's much point to this one right now. The definion of `Data.IntMap.alterF` was lifted straight from here! Someone (either `containers` or `lens`) should add some `RULES` to specialize to `Identity` and `Const`. I suppose that probably should be `containers`, at which point this change is justified. I don't really care *too* much either way. Do you have strong feelings about it?

src/Control/Lens/At.hs
```diff
@@ -454,17 +458,25 @@ instance At (Maybe a) where
...
+#else
   at k f m = f mv <&> \r -> case r of
     Nothing -> maybe m (const (IntMap.delete k m)) mv
     Just v' -> IntMap.insert k v' m
     where mv = IntMap.lookup k m
+#endif
   {-# INLINE at #-}
 
 instance Ord k => At (Map k a) where
+#if MIN_VERSION_containers(0,5,8)
```
treeowl: The above comment does *not* apply here; `Map` lookups have the potential to be expensive, which `IntMap` lookups don't, justifying the custom implementation of `Data.Map.alterF`.

src/Control/Lens/TH.hs
```diff
@@ -99,10 +99,12 @@ import Control.Lens.Type () -- haddocks
 import Data.Char (toLower, isUpper)
 import Data.Foldable hiding (concat, any)
 import Data.List as List
-import Data.Map as Map hiding (toList,map,filter)
+import qualified Data.Map as Map hiding (toList,map,filter)
```
treeowl: Qualified imports only need `hiding` clauses when importing multiple modules `as` the same name.

src/Control/Lens/TH.hs
```diff
@@ -99,10 +99,12 @@ import Control.Lens.Type () -- haddocks
...
+import Data.Map (Map)
 import Data.Maybe (maybeToList)
 import Data.Monoid
-import Data.Set as Set hiding (toList,map,filter)
+import qualified Data.Set as Set hiding (toList,map,filter)
```
treeowl: No hiding needed here either.

src/Control/Lens/At.hs
RyanGlScott: > Do you have strong feelings about it?

No, certainly not! I only added it since you mentioned `IntMap`, and I was naïvely trying to catch all of the cases where `containers-0.5.8.1` added something that could be used in `lens`. I'll revert this particular change for now.

treeowl: Eh... Probably better to leave it as you have it actually. I'll see if there's a ticket for `containers` to add any appropriate `RULES`.

treeowl: Better to keep as much as possible in `containers`, probably.

src/Control/Lens/TH.hs
RyanGlScott: Good catch, fixed.

src/Control/Lens/At.hs
RyanGlScott: It sounds like there are plans (https://github.com/haskell/containers/issues/325) to add `RULES` for `Data.IntMap.alterF`, so I'll opt to keep this definition, then.

PR #676, opened 2016-09-22
Add Data.Map.Lens.toMapOf

phadej:
I tried hard to find whether there is generic version of `toMapOf`, without luck.

If this is OK, I'd like to add `HashMap` and `IntMap` versions, also strict.

Also a monoidal version `(Monoid m, ...) => ... -> Map i m`, might be useful, but I don't know a good name for it


glguy:
Is this implementation with singleton better than (performance) itoListOf and fromList?

phadej:
I haven't checked the performance, just copied how `setOf` is implemented. The benefit of singleton approach, is that `Getter` and `Lens` variants don't need `Ord` instance. That's minor, but nice anyway.

hlian:
i think this could warrant an entry in `CHANGELOG`

ekmett:
No objection here. I'm also okay with the monoidal form it has taken. That can yield better asymptotics in the presence of sharing -- of course the linear scan form can yield better asymptotics in the presence of strictly monotone increasing keys.

PR #681, opened 2016-10-09
Restore the generic(1) functions in GHC.Generics.Lens

RyanGlScott:
#659, which removed `Generics.Deriving.Lens`, had the inadvertent effect of removing the `generic` and `generic1` combinators, which the documentation of `GHC.Generics.Lens` references (as that module previously reexported them from `Generics.Deriving.Lens`). We can at least make it better by reintroducing them back to `GHC.Generics.Lens`.

A stop-gap measure for #657.


glguy:
whoops!

PR #682, opened 2016-10-10
Add a Generic default implementation for Wrapped

RyanGlScott:
Inspired by the [`newtype-generics`](http://hackage.haskell.org/package/newtype-generics) library, we can quite easily define a `Generic`-based default implementation for `Wrapped` that simply wraps and unwraps the generic representation of a newtype (or datatype with one constructor and field).

Open question: should we use this to eliminate some of the boilerplate in `Control.Lens.Wrapped` for its `Wrapped` instances? The generics code here uses exclusively `newtype`s, so it's extremely easy to optimize away and thus the performance should be the same. In fact, I tried compiling `main = print $ ala Sum foldMap [1,2,3,4]` before and after these changes, and in both cases they produced the same Core.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,3 +1,7 @@
+next
+----
+* Add a `Generic` default implementation for `Wrapped`
+
 4.15.1
 ----
 * Restore the `generic` and `generic1` functions in `GHC.Generics.Lens`
--- a/src/Control/Lens/Wrapped.hs
+++ b/src/Control/Lens/Wrapped.hs
@@ -80,7 +80,6 @@ module Control.Lens.Wrapped
 #endif
   -- * Generics
   , _GWrapped'
-  , GWrapped
   ) where
 
 #include "HsBaseConfig.h"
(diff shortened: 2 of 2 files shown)


ekmett:
Great idea! I do admit the GWrapped flexible instance kinda bugs me, because it feels like the errors users get will be pretty obtuse, but I think i can get over it.

RyanGlScott:
@ekmett: I could just as well split up that flexible instance into four or so individual, non-flexible instances, and I think it would work out the same. Would that assuage your concerns?

ekmett:
Nah. It is probably good as it is. 

The one alternative that comes to mind is more about using one instance with equality constraints, so that when you incur the dependency you get channeled to the right instance, but with the extra floating variables that may well be 4 instances as well.

glguy:
We don't even need a class, we can embed the implementation in the default instance!

```haskell
class Wrapped s where
  type Unwrapped s :: *
  type Unwrapped s = GUnwrapped (Rep s)

  -- | An isomorphism between @s@ and @a@.
  --
  -- If your type has a 'Generic' instance, '_Wrapped'' will default to '_GWrapped'',
  -- and you can choose to not override it with your own definition.
  _Wrapped' :: Iso' s (Unwrapped s)
  default _Wrapped' :: (Generic s, D1 d (C1 c (S1 e (Rec0 a))) ~ Rep s, Unwrapped s ~ GUnwrapped (Rep s))
                    => Iso' s (Unwrapped s)
  _Wrapped' = iso Generic.from Generic.to . _GWrapped'
    where
      _GWrapped' = iso remitter reviewer
      remitter (M1 (M1 (M1 (K1 x)))) = x
      reviewer x = M1 (M1 (M1 (K1 x)))
  {-# INLINE _Wrapped' #-}

type family GUnwrapped rep where
  GUnwrapped (D1 d (C1 c (S1 s (Rec0 a)))) = a
```

ekmett:
That works too.

RyanGlScott:
@glguy, great idea! Unfortunately, we can't use a closed type family for `GUnwrapped` for backwards-compat reasons, but oh well.

RyanGlScott:
I'm debating whether it's worthwhile to export `_GWrapped'` from `Control.Lens.Wrapped`. On one hand, it'd let you use it as a standalone `Iso` in your own code. On the other hand, it would lead to the `GUnwrapped` class (which, ostensibly, is an internal detail) being leaked into the Haddocks...

ekmett:
I don't particularly mind the leaking of extra information into the haddocks.

We wind up commenting the heck out of everything in `lens` anyways because the types are pretty opaque to begin with.

Issue #684, opened 2016-11-02
TH generation of prisms for GADTs in GHC 8 generates uncompilable prisms which are too general


Dridus:
reproduction repo: https://github.com/Dridus/lens-gadt-prisms-issue - that one is set to LTS-6.23 with lens 4.13 and GHC 7.10.3 and compiles correctly. The stack.yaml has a commented out line to switch to LTS-7.7, lens 4.14, GHC 8.0.1.

this code:

```haskell
import Control.Lens.TH (makePrisms)

data A = A
data B = B

data Gadt a where
  GadtA :: A -> Gadt A
  GadtB :: B -> Gadt B

makePrisms ''Gadt
```

compiles fine with LTS-6.x, GHC 7.10.3, lens 4.13, but fails to compile with LTS-7.x, GHC 8.0.1, lens 4.14:

```
/Users/ross/1st/lens-gadt-prisms-issue/src/Main.hs:12:1: error:
    • Couldn't match type ‘a0’ with ‘B’
      ‘a0’ is a rigid type variable bound by
        the type signature for:
          _GadtB :: forall a0 a1 (p :: * -> * -> *) (f :: * -> *).
                    (profunctors-5.2:Data.Profunctor.Choice.Choice p, Applicative f) =>
                    p B (f B) -> p (Gadt a1) (f (Gadt a0))
        at src/Main.hs:12:1
      Expected type: p B (f B) -> p (Gadt a1) (f (Gadt a0))
        Actual type: p B (f B) -> p (Gadt a1) (f (Gadt B))
    • In the expression:
        Control.Lens.Prism.prism
          (\ x_a6c0 -> GadtB x_a6c0)
          (\ x_a6c1
             -> case x_a6c1 of {
                  GadtA y1_a6c2 -> Left (GadtA y1_a6c2)
                  GadtB y1_a6c3 -> Right y1_a6c3 })
      In an equation for ‘_GadtB’:
          _GadtB
            = Control.Lens.Prism.prism
                (\ x_a6c0 -> GadtB x_a6c0)
                (\ x_a6c1
                   -> case x_a6c1 of {
                        GadtA y1_a6c2 -> Left (GadtA y1_a6c2)
                        GadtB y1_a6c3 -> Right y1_a6c3 })
    • Relevant bindings include
        _GadtB :: p B (f B) -> p (Gadt a1) (f (Gadt a0))
          (bound at src/Main.hs:12:1)

[...]

PR #687, opened 2016-11-14
Fix `makePrisms` to work with GADTs

pseudonom:
It looks like GADTs used to specify the return type via the context.

```haskell
data Foo a where
  Foo :: Foo Bool
```

would desugar to something like

```haskell
data Foo a where
  Foo :: (a ~ Bool) => Foo a
```

Apparently, this representation has changed and the `GadtC` constructor now
embeds the return type directly.

This commit uses the new source of information and thus partially address #684. Partially, because this commit does not fix the generation of classy GADT prisms.

### DIFF
--- a/src/Control/Lens/Internal/PrismTH.hs
+++ b/src/Control/Lens/Internal/PrismTH.hs
@@ -33,6 +33,7 @@ import Control.Lens.Tuple
 import Control.Monad
 import Data.Char (isUpper)
 import Data.List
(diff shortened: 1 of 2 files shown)


pseudonom:
Also, rather than tupling `(NCon, Type)`, would we be better off adding `Type` to `NCon`?

glguy:
GADT support has been off-loaded to the th-abstraction package.

PR #688, opened 2016-11-14
Fix order of :< and :> in Control.Lens.Cons

axman6:
The order of these patterns in the docs seems wrong - the Snoc one is after the Cons functions and the Cons one is after the Snoc functions.


axman6:
Apparently the 7.6.3 build is failing, not sure why but doesn't appear to be my changes.

RyanGlScott:
I believe I've fixed the build issues on `master`. Can you rebase your changes on top of that?

axman6:
I've probably made a big mess of this, I'm not used to rebasing let me know if I've messed everything up and I'll just make a new PR.

PR #692, opened 2016-11-28
Reverse the order of builds in travis - newer compilers run sooner.

axman6:
It seems later GHCs run the travis build up to 50% slower than earlier GHCs; reordering these should mean slower builds start earlier, reducing overall latency.


ekmett:
No objection, unfortunately conflicts have merged since this pull request.

Issue #694, opened 2016-11-29
Add SurelyIxed class


treeowl:
```haskell
class Ixed m => SurelyIxed m where
  ixSure :: Index m -> Lens' m (IxValue m)
  default ixSure ::
    (Representable f, Eq (Rep f),
     Rep f ~ Index (f a),
     a ~ IxValue (f a)) => Rep f -> Lens' (f a) a
```

The superclass constraint is trivially justified by `ix = ixSure`. The default signature is mostly to make a point; it's terribly inefficient.

PR #695, opened 2016-11-29
Add SurelyIxed class

treeowl:
This refines `Ixed` for total indexing, such as for representable
functors.

Implements #694

### DIFF
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -41,6 +41,8 @@ module Control.Lens.At
     At(at)
     , sans
     , iat
+  -- * SurelyIxed
+  , SurelyIxed(ixSure)
   -- * Ixed
   , Index
   , IxValue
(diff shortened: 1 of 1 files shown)


treeowl:
Should I add the `Representable` "default"? This requires adding a direct `adjunctions` dependency (there's already an indirect one). Using this as a `Getter` should be just fine, but it's a thoroughly lousy `Setter`.

```haskell
ixSureRep :: (Representable f
             , Eq (Rep f)
             , Index (f a) ~ Rep f
             , IxValue (f a) ~ a)
          => Rep f -> Lens' (f a) a
ixSureRep e p f = p (Rep.index f e) <&> \a ->
   Rep.tabulate (\e' -> if e == e' then a else Rep.index f e')
```

treeowl:
Another option would be to use a different `SureIndex` type family and remove the `Ixed` superclass. This would allow one type of index for partial lookups and another for total ones.

fumieval:
I don't see how this could be useful.

phadej:
I use similar class at work (I call it `Att` :P). It's useful for `a -> b` instance. Also I have a date `TotalMap k v { tmDef :: v, tmMap :: Map k v }`, where `Att` is useful.

I don't like `Att` neither `SurelyIxed` names. something short yet descriptive would be great.

ekmett:
Given a general dearth of instances in the things lens depends on, I think I'm going to close this as an out of scope sort of issue. Feel free to package it somewhere independently.

PR #699, opened 2016-12-23
Cleaner `deepOf` implementation

fosskers:
This leverages `failing` to provide a cleaner implementation.

**Note:** `deepOf id other` will enter an infinite loop if `other` fails, but that is true for the current implementation as well (so Don't-do-That™).

I can see this being rejected if Traversal laws are a concern when involving `failure`, as mentioned in [this Stackoverflow post](http://stackoverflow.com/questions/27138856/why-does-failing-from-lens-produce-invalid-traversals).


fosskers:
The Travis failure is:
```
The command "cabal install --only-dependencies --enable-tests --enable-benchmarks --dry -v > installplan.txt" failed and exited with 1 during .
```
which only occurred with `ghc-8.0.2`.

PR #700, opened 2017-01-09
Add firstOf1, lastOf1, both1

phadej:
If this is ok, I'll add `maximumOf1`, `minimumOf1` etc `Semigroup` folds.

### DIFF
--- a/src/Control/Lens/Fold.hs
+++ b/src/Control/Lens/Fold.hs
@@ -77,7 +77,7 @@ module Control.Lens.Fold
   -- ** Folding
   , foldMapOf, foldOf
   , foldrOf, foldlOf
-  , toListOf
+  , toListOf, toNonEmptyOf
   , anyOf, allOf, noneOf
   , andOf, orOf
   , productOf, sumOf
@@ -88,7 +88,7 @@ module Control.Lens.Fold
   , elemOf, notElemOf
   , lengthOf
   , nullOf, notNullOf
-  , firstOf, lastOf
+  , firstOf, firstOf1, lastOf, lastOf1
   , maximumOf, minimumOf
   , maximumByOf, minimumByOf
   , findOf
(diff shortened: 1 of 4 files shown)


src/Control/Lens/Fold.hs
```diff
@@ -599,6 +604,22 @@ toListOf :: Getting (Endo [a]) s a -> s -> [a]
...
+-- @
+-- 'toListOf' :: 'Getter' s a      -> s -> NonEmpty a
+-- 'toListOf' :: 'Fold1' s a       -> s -> NonEmpty a
+-- 'toListOf' :: 'Lens'' s a       -> s -> NonEmpty a
+-- 'toListOf' :: 'Iso'' s a        -> s -> NonEmpty a
+-- 'toListOf' :: 'Traversal1'' s a -> s -> NonEmpty a
+-- 'toListOf' :: 'Prism'' s a      -> s -> NonEmpty a
+-- @
+toNonEmptyOf :: Getting (NonEmpty a) s a -> s -> NonEmpty a
+toNonEmptyOf l = foldMapOf l (\x -> x :| [])
```
phadej: I'm not sure how to make dlist trick to avoid O(n^2) so went with this for now.

phadej:
Side comment/question: would be cool to have `Apply` as superclass of `Applicative`, were there any related discussions somewhere?

phadej:
travis breakage will be hopefully resolved via https://github.com/hvr/cassava/pull/125

ekmett:
A large part of the reason for using the old cabals was to ensure we could still build with them without being caught by surprise. I'm happy to merge the first part.

RyanGlScott:
Actually, it appears that `.travis.yml` is _currently_ using `cabal-install-head`, after 4b8751488fd1520a71f22ab910ca2542fb49dcf6.

ekmett:
Then I guess we already took the plunge and will have to deal with these issues if we find them post-deployment.

PR #646, opened 2016-04-13
WIP: Migrate to nix-local-build based .travis.yml

hvr:
DO NOT MERGE YET

PR #701, opened 2017-01-11
Doctests new build

phadej:


ekmett:
Can you add a preamble to the Setup.lhs like:

```
#ifndef MIN_VERSION_Cabal
#define MIN_VERSION_Cabal(X,Y,Z) 0
#endif
```

This way when run on older cabal versions it'll be more robust.

phadej:
@ekmett, I did. Let's see what travis says. I guess `setup-depends: Cabal >= 1.24` is ok, as it really helps if people move to recent `cabal`.

RyanGlScott:
Thanks for doing this, @phadej! Since there's a number of other packages that use a similar `Setup.hs` script (and since we don't yet have [`cabal doctest`](https://github.com/haskell/cabal/issues/2327)), I'd like to adapt this for other packages. Some questions I have after skimming the changes:

1. Does this really require `cabal-install-head` as the minimum?
2. Is there any logic here that's `lens`-specific that couldn't just be copy-pasted into something like, say, `distributive`'s `Setup.hs`?

phadej:
@RyanGlScott For 2: The lens diagram copying code is something none of other packages need: https://github.com/ekmett/lens/blob/master/Setup.lhs#L42-L44

For 1: I'm not sure, it might work with `cabal-install-1.24 new-build` too, too lazy to check what https://github.com/haskell/cabal/blame/675d4f1625edc6e3d3a68448cbc72a7931a590ea/cabal-install/Distribution/Client/SetupWrapper.hs#L810-L821 looks like.

But anyhow, `cabal-install-1.24 new-build` is/was a technology preview, and I wouldn't use it if I can pick `HEAD` (with HVR's ppa repo it's easy, if you are on Ubuntu; I'm also planning to provide nightly OSX builds of `cabal`, but run out of time setting that up: now I said it publicly, so I guess it will happen sooner ;) )

RyanGlScott:
1. Oh, I didn't read the `new-build` part of the title. But `lens`'s Travis script [isn't currently using `new-build`](https://github.com/ekmett/lens/pull/646), right? It's still using `configure`/`build`/`test` for the moment. So as long as you stick to the old-style build process, does it matter what version of `cabal-install` you use?

2. Good point, I'll make sure not to use the diagram-copying part elsewhere.

bennofs:
@phadej does this work with older cabal-install versions? I'd assume it requires a cabal-install that knows about `custom-setup`, or parsing of the cabal file will fail.

phadej:
parsing doesn't fail, cabal just says "unknown section". You can see it e.g. from https://travis-ci.org/ekmett/distributive/builds/197339532

bennofs:
Oh cool, I had assumed that this was less backwards-compatible than it actually is. This is nice, thanks!

PR #703, opened 2017-01-11
Try Setup.lhs with cabal-doctest

phadej:
Let's see if this work. Thanks @ezyang for the `display` idea. Ping @RyanGlScott


Setup.lhs
```diff
@@ -72,33 +53,10 @@ generateBuildModule verbosity pkg lbi = do
         , ""
         , "compiler :: String"
         , "compiler = " ++ (show $ showCompilerId $ compiler lbi)
-        , ""
-        , "depsFlag :: String"
-#if MIN_VERSION_Cabal(1,24,0)
-        , "depsFlag = " ++ show "-package-id="
-#else
-        , "depsFlag = " ++ show "-package="
-#endif
```
RyanGlScott: Why don't we need to be picky about choosing `-package=` vs. `=package-id` anymore? (We seem to be unconditionally using  `=package-id` in `doctests.hsc`.)

phadej:
@RyanGlScott see https://github.com/ekmett/bytes/pull/32, I guess we have to work bottom-up for GHC-8.2.

Setup.lhs
phadej: `package-id` is more precise (you can have single version of a package installed multiple times, which happens with new-build more often than you might think at first). Before realising that `display` did *the right* thing, even `InstalledPackageId` representation changed in between, I didn't know how to extract the package-id string without CPP magic (which, unfortunately, wasn't reliable)

PR #708, opened 2017-01-30
Add 'recons' and 'unfoldr' to Control.Lens.Cons

frasertweedale:
recons :: (Cons s1 s1 a a, Cons s2 s2 a a, AsEmpty s2) => Getter s1 s2
  unfoldr :: (Cons s2 s2 a a, AsEmpty s2) => (s1 -> Maybe (a, s1)) -> s1 -> s2

For efficieny, rewrite rules are provided for common conversions.

### DIFF
--- a/src/Control/Lens/Cons.hs
+++ b/src/Control/Lens/Cons.hs
@@ -29,6 +29,8 @@ module Control.Lens.Cons
   , cons
   , uncons
   , _head, _tail
+  , recons
+  , unfoldr
 #if __GLASGOW_HASKELL__ >= 710
   , pattern (:<)
 #endif
@@ -44,8 +46,10 @@ module Control.Lens.Cons
 
   ) where
 
+import Control.Lens.Empty (AsEmpty(..))
 import Control.Lens.Equality (simply)
 import Control.Lens.Fold
+import Control.Lens.Getter (to)
 import Control.Lens.Prism
 import Control.Lens.Review
 import Control.Lens.Tuple
@@ -54,7 +58,7 @@ import qualified Data.ByteString      as StrictB
 import qualified Data.ByteString.Lazy as LazyB
 import           Data.Monoid
 import qualified Data.Sequence as Seq
-import           Data.Sequence hiding ((<|), (|>), (:<), (:>))
+import           Data.Sequence hiding ((<|), (|>), (:<), (:>), unfoldr)
 import qualified Data.Text      as StrictT
 import qualified Data.Text.Lazy as LazyT
 import           Data.Vector (Vector)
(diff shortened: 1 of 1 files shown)


src/Control/Lens/Cons.hs
```diff
@@ -319,6 +323,42 @@ _tail :: Cons s s a a => Traversal' s s
...
+-- | Convert one type with a 'Cons' instance into the other.
+--
+-- Rewrite rules are provided for efficient conversion between
+-- 'String' and 'Text', @['Word8']@ and 'ByteString', and lazy and
+-- strict 'Text' and 'ByteString'.  Programs must be compiled
+-- with @-O@ to use them.
+--
+-- If @(recons . recons)@ exists it should obey:
+--
+-- > recons . recons ≡ id
```
phadej: or `recons . recons = recons`?

src/Control/Lens/Cons.hs
```diff
@@ -319,6 +323,42 @@ _tail :: Cons s s a a => Traversal' s s
...
+"recons/bs-list" recons = to StrictB.unpack
+"recons/list-lazybs" recons = to LazyB.pack
+"recons/lazybs-list" recons = to LazyB.unpack
+"recons/bs-strict" recons = to LazyB.toStrict
+"recons/bs-lazy" recons = to LazyB.fromStrict
+ #-}
+
+-- | > unfoldr f = foldrOf (unfolded f) cons (_Empty # ())
+--
+unfoldr :: (Cons s2 s2 a a, AsEmpty s2) => (s1 -> Maybe (a, s1)) -> s1 -> s2
```
phadej: I'm not sure we want export this name. OTOH we already export `cons`, so maybe it isn't an real issue.

src/Control/Lens/Cons.hs
```diff
@@ -319,6 +323,42 @@ _tail :: Cons s s a a => Traversal' s s
...
+-- Rewrite rules are provided for efficient conversion between
+-- 'String' and 'Text', @['Word8']@ and 'ByteString', and lazy and
+-- strict 'Text' and 'ByteString'.  Programs must be compiled
+-- with @-O@ to use them.
+--
+-- If @(recons . recons)@ exists it should obey:
+--
+-- > recons . recons ≡ id
+--
+recons :: (Cons s1 s1 a a, Cons s2 s2 a a, AsEmpty s2) => Getter s1 s2
```
phadej: Why not `recons ;: (Cons s1 s1 a a, Cons s2 s2 a a, AsEmpty s1, AsEmpty s2) => Iso' s1 s2`?

src/Control/Lens/Cons.hs
frasertweedale: Maybe both?  Sensible implementations should be involutive and idempotent.

src/Control/Lens/Cons.hs
frasertweedale: I originally had that.  @tonymorris urged the change because it would be possible to write 'Cons' instances that result in an optic that does not obey `Iso` laws.  Such an instance would be daft, but still, the types do not guarantee it.

That said, there is precedent in *lens* of unlawful values.  I defer to the *lens* maintainer(s) wishes on this.

src/Control/Lens/Cons.hs
frasertweedale: I discussed this change with @ekmett who seemed more keen on `unfoldr` than `recons` :)

frasertweedale: IRC log: https://gist.github.com/tonymorris/1b219366a8796890b82569f85e80dd4d

src/Control/Lens/Cons.hs
phadej: @frasertweedale `recons = id` if `s1 = s2`, isn't it? Maybe that can be written explcitly (And instances added).

src/Control/Lens/Cons.hs
```diff
@@ -319,6 +323,42 @@ _tail :: Cons s s a a => Traversal' s s
 _tail = _Cons._2
 {-# INLINE _tail #-}
 
+-- | Convert one type with a 'Cons' instance into the other.
+--
+-- Rewrite rules are provided for efficient conversion between
+-- 'String' and 'Text', @['Word8']@ and 'ByteString', and lazy and
+-- strict 'Text' and 'ByteString'.  Programs must be compiled
+-- with @-O@ to use them.
```
phadej: This text is misleading, you can understand that you can convert between all of those. AFAICS you can convert `ByteString` to `[Word8]` and `Text` to `[Char]`, and lazy to strict variants. How this is better than `strict` / `lazy` and `packed` / `unpacked` ?

src/Control/Lens/Cons.hs
frasertweedale: "instances added" <- what do you mean by this, @phadej?

src/Control/Lens/Cons.hs
phadej: Ok. The `unfoldr` is :+1:. I'm not sure about `recons` either, let @ekmett say the final word.

src/Control/Lens/Cons.hs
phadej: Sorry, my bad, there is `recons = id` rewrite rule. Missed that.

src/Control/Lens/Cons.hs
frasertweedale: I will improve the wording.

How is it better?  More polymorphic type aids readability and reuse.

src/Control/Lens/Cons.hs
phadej: Anyhow, writing few example instances

```haskell
recons :: Getter Text String
recons :: Getter [a] [a]
recons :: Getter (Vector a) [a]
recons :: Getter ByteString [Word8]
```

would highlight the generality of this combinator, what you think?

frasertweedale: Sure; happy to add this.

src/Control/Lens/Cons.hs
phadej: I agree. But I just don't like this solution. There is e.g. `https://hackage.haskell.org/package/string-conversions-0.4.0.1/docs/Data-String-Conversions.html#v:convertString` which though assumes the encoding of `ByteString`, so isn't problematic either.

phadej: Anyhow, I'm not forced to use this, so if Edward is ok with this, let him merge this PR.

frasertweedale:
@glguy `Empty` constraints could be replaced with `Monoid`.  Does that taste less bitter?

ekmett:
I think I'm going to reject this on the grounds that it takes a module that is already borderline for inclusion and pushes it over the edge of messy pseudo-laws.

frasertweedale:
No worries @ekmett.  It's published on Hackage anyhow: http://hackage.haskell.org/package/concise 

Thanks for considering :)

PR #711, opened 2017-02-10
Compatibility with GHC 8.2

bgamari:
Data.Dynamic now exposes its constructor due to type-indexed Typeable,
so it is necessary to qualify the export of the Dyanmic pattern in
Data.Dynamic.Lens.

Also, the default signature of itraverse did not match its class'
definition, triggering GHC's new default method compatibility check.


glguy:
Does line 508 need to have two `=>`?

PR #717, opened 2017-03-23
Fix broken link in haddock

janpath:
Also `cloneLens` and alike don't really use `Context` (anymore?) but `Pretext (->)`, so maybe this should be changed anyway?


phadej:
> The indexed store **can be** used to characterize a `Lens`

That's true, `Pretext` is an optimisation.

janpath:
Ah thanks, I missed that.

PR #728, opened 2017-04-25
Add examples to withIndex

phadej:


glguy:
I'm in favor of more examples. I think there's a bit of an issue with the first example in that we have better ways to do that than to use withIndex: `[10,20,30] ^@.. ifolded`

Of course on the other hand this wouldn't be the first suboptimal example of using `withIndex`. There are examples for `(.>)` and `(<.)` that do the same thing.

phadej:
@glguy I don't remember my motivation for the first example, but IIRC it was to make the second one  easier to comprehend.

ekmett:
I can see it. It does help with reading the second example.

PR #730, opened 2017-04-25
Add minimum1Of, maximum1Of; rename firstOf1 to first1Of etc.

### LINKED ISSUE #727: maximum1Of and minimum1Of
Since we have `firstOf1`, `lastOf1` (why are they not called `head1Of`, `last1Of`?) it may be a good idea to give a `maximum1Of`, `minimum1Of` that do not return a `Maybe` using a `Getting` over `Data.Semigroup.Max`, `Data.Semigroup.Min`.

I don't have *lens* atm but they should be easy enough to write.

We could also add total versions of `foldr1Of`, `foldl1Of`.

phadej:
Resolves #727


glguy:
It's unfortunate that O and 0 look so similar. I first read "minimum-ten-f". (I'm still in favor of having "Of" last.)

PR #732, opened 2017-05-06
Improve error messages by using LensLike, Over and Optic

Gurkenglas:
Hopefully, that is. I don't have a benchmark file that generates notorious lens errors.

### DIFF
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -43,7 +43,10 @@ module Control.Lens.At
   , icontains
   ) where
 
+import Prelude ()
+
 import Control.Lens.Each
+import Control.Lens.Internal.Prelude
 import Control.Lens.Traversal
 import Control.Lens.Lens
 import Control.Lens.Setter
@@ -60,7 +63,6 @@ import Data.HashSet as HashSet
 import Data.Int
 import Data.IntMap as IntMap
 import Data.IntSet as IntSet
-import Data.List.NonEmpty as NonEmpty
 import Data.Map as Map
 import Data.Set as Set
 import Data.Sequence as Seq
@@ -72,10 +74,6 @@ import Data.Vector.Primitive as Prim
 import Data.Vector.Storable as Storable
 import Data.Vector.Unboxed as Unboxed hiding (indexed)
 import Data.Word
-
-#if !MIN_VERSION_base(4,8,0)
-import Control.Applicative
-#endif
 
 type family Index (s :: *) :: *
 type instance Index (e -> a) = e
--- a/src/Control/Lens/Each.hs
+++ b/src/Control/Lens/Each.hs
@@ -28,17 +28,18 @@ module Control.Lens.Each
     Each(..)
   ) where
 
+import Prelude ()
+
 import Control.Lens.Traversal
 import Control.Lens.Internal.ByteString
+import Control.Lens.Internal.Prelude
 import Data.Array.Unboxed as Unboxed
 import Data.Array.IArray as IArray
 import Data.ByteString as StrictB
 import Data.ByteString.Lazy as LazyB
 import Data.Complex
-import Data.Functor.Identity
 import Data.HashMap.Lazy as HashMap
 import Data.IntMap as IntMap
-import Data.List.NonEmpty
 import Data.Map as Map
 import Data.Sequence as Seq
 import Data.Text.Lens (text)
@@ -54,10 +55,6 @@ import Data.Vector.Storable (Storable)
 import qualified Data.Vector.Unboxed as Unboxed
 import Data.Vector.Unboxed (Unbox)
 import Data.Word
-
-#if !MIN_VERSION_base(4,8,0)
-import Control.Applicative
-#endif
 
 -- $setup
 -- >>> :set -XNoOverloadedStrings
--- a/src/Control/Lens/Empty.hs
+++ b/src/Control/Lens/Empty.hs
@@ -28,12 +28,14 @@ module Control.Lens.Empty
 #endif
   ) where
 
+import Prelude ()
+
 import Control.Lens.Iso
 #if __GLASGOW_HASKELL__ >= 710
 import Control.Lens.Fold
 #endif
-import Control.Applicative (ZipList(..))
 import Control.Lens.Prism
+import Control.Lens.Internal.Prelude as Prelude
 import Control.Lens.Review
 import Data.ByteString as StrictB
 import Data.ByteString.Lazy as LazyB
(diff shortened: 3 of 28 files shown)


Gurkenglas:
Perhaps an updating of the documentation of LensLike, Over and Optic is also in order to tell people what `(Conjoined p, Functor f) => Optic p f` in their error messages means.

glguy:
I'd want to see a bunch of examples of how this change makes better GHC message output before churning so many synonyms.

duplode:
(Chiming in from the [/r/haskell thread](https://www.reddit.com/r/haskell/comments/69ots1/notorious_lens_errors_thread/). Here are some fence-straddling, not fleshed out thoughts about potential implications of the patch.)

As things stand, these changes would add an extra level of indirection to the documentation in case e.g. one wants to see what a `Lens` consists of. They would also increase the visibility of synonyms such as `Over` and `Optic`, which aren't exactly self-explanatory and, under the status quo, seem unlikely to be met by beginners during their first steps with *lens*. The flip side is that the synonyms could be something of an anchor when looking for help after seeing an error (as @Gurkenglas implies by suggesting "an updating of the documentation"), even though the generality of the synonyms means any remarks about them have to be similarly general (cf. the [current documentation for `Optic`](https://hackage.haskell.org/package/lens-4.15.1/docs/Control-Lens-Type.html#t:Optic): "This gives rise to the laws for `Equality`, `Iso`, `Prism`, `Lens`, `Traversal`, `Traversal1`, `Setter`, `Fold`, `Fold1`, and `Getter` as well along with their index-preserving variants").

That said, here is a tiny mockup of the sort of change in the patch, meant more as a conversation starter than as a representative of any realistic scenario.

``` haskell
type MyPrism s t a b = forall p f. (Choice p, Applicative f) => Optic p f s t a b

_MyJust :: MyPrism (Maybe a) (Maybe b) a b
_MyJust = _Just
```

(I considered throwing in redefinitions and misuses of `below` and `inside` -- the sort of thing that begins to make shortened signatures welcome -- but I felt the signal-to-noise ratio would be too low for this comment.)

The synonym shows up alright on `:t`...

```
GHCi> :t _Just
_Just
  :: (Choice p, Applicative f) =>
     p a (f b) -> p (Maybe a) (f (Maybe b))
GHCi> :t _MyJust
_MyJust
  :: (Choice p, Applicative f) => Optic p f (Maybe a) (Maybe b) a b
```

... even though it is still necessary to know the significance of the `Choice` and `Applicative` constraints in order to realise this is a `Prism`.

The synonym also shows up in this arbitrary example of a type error:

```
GHCi> [1..3] & _Just %%~ \x -> [-x..x]

<interactive>:34:10: error:
    • Couldn't match type ‘Maybe a’ with ‘[Integer]’
      Expected type: LensLike [] [Integer] (Maybe a) a a
        Actual type: (a -> [a]) -> Maybe a -> [Maybe a]
    • In the first argument of ‘(%%~)’, namely ‘_Just’
      In the second argument of ‘(&)’, namely
        ‘_Just %%~ \ x -> [- x .. x]’
      In the expression: [1 .. 3] & _Just %%~ \ x -> [- x .. x]
    • Relevant bindings include
        it :: [Maybe a] (bound at <interactive>:34:1)
GHCi> [1..3] & _MyJust %%~ \x -> [-x..x]

<interactive>:35:10: error:
    • Couldn't match type ‘Maybe a’ with ‘[Integer]’
      Expected type: LensLike [] [Integer] (Maybe a) a a
        Actual type: Optic (->) [] (Maybe a) (Maybe a) a a
    • In the first argument of ‘(%%~)’, namely ‘_MyJust’
      In the second argument of ‘(&)’, namely
        ‘_MyJust %%~ \ x -> [- x .. x]’
      In the expression: [1 .. 3] & _MyJust %%~ \ x -> [- x .. x]
    • Relevant bindings include
        it :: [Maybe a] (bound at <interactive>:35:1)
```

(The `LensLike` is a courtesy of `(%%~)`.)

However, if I rewrite this without using `(%%~)`, the synonym disappears:

```
GHCi> _Just (\x -> [-x..x]) [1..3]

<interactive>:36:1: error:
    • Couldn't match type ‘Maybe a’ with ‘[Integer]’
      Expected type: [Integer] -> [Maybe a]
        Actual type: Maybe a -> [Maybe a]
    • The function ‘_Just’ is applied to two arguments,
      its type is ‘p0 a (f0 b0) -> p0 (Maybe a) (f0 (Maybe b0))’,
      it is specialized to ‘(a -> [a]) -> Maybe a -> [Maybe a]’
      In the expression: _Just (\ x -> [- x .. x]) [1 .. 3]
      In an equation for ‘it’: it = _Just (\ x -> [- x .. x]) [1 .. 3]
    • Relevant bindings include
        it :: [Maybe a] (bound at <interactive>:36:1)
GHCi> _MyJust (\x -> [-x..x]) [1..3]

<interactive>:37:1: error:
    • Couldn't match type ‘Maybe a’ with ‘[Integer]’
      Expected type: [Integer] -> [Maybe a]
        Actual type: Maybe a -> [Maybe a]
    • The function ‘_MyJust’ is applied to two arguments,
      its type is ‘p0 a (f0 b0) -> p0 (Maybe a) (f0 (Maybe b0))’,
      it is specialized to ‘(a -> [a]) -> Maybe a -> [Maybe a]’
      In the expression: _MyJust (\ x -> [- x .. x]) [1 .. 3]
      In an equation for ‘it’: it = _MyJust (\ x -> [- x .. x]) [1 .. 3]
    • Relevant bindings include
        it :: [Maybe a] (bound at <interactive>:37:1)
```

This suggests a bunch of examples might be necessary to survey where the sweetening breaks down.

(By the way, is there any executive summary around explaining what GHC does with synonyms in messages? A quick search has only led me to this [remark by Gabriel Gonzales](http://stackoverflow.com/questions/27157890/could-ghc-be-enhanced-to-print-type-aliases-in-error-messages#comment43866689_27157890): "Basically if you do anything non-trivial to the value the type alias is lost.")

ekmett:
I definitely like the notion of using LensLike for Traversal and Lens as it doesnt get "exploded" when we drag Applicative (or Functor) out front during typechecking. or when unifying with f, that leads to cocrrete wins in terms of type error legibility.

ekmett:
Optic is less clear of a win, you avoid duping 'f' in the type error, but it is still prety painfullu opaque.

Gurkenglas:
If we want to degenerate completely to form typechecker behavior, we can introduce aliases for `Optic`, `Over` and `LensLike` for each type synonym touched by this patch just for type error readability:

```
type Traversal s t a b = forall f. Applicative f => MonomorphicTraversal f s t a b
type MonomorphicTraversal f s t a b = LensLike f s t a b
```

These would be used nowhere except that line, and the documentation would say that `MonomorphicX f` is always a specialized `X`.

ekmett:
I don't think we want to proliferate more than a couple of names this way.

LensLike vs Over vs Optic vs Optical is bad enough.

ekmett:
If we do make this change (which I'm leaning towards) I'd like to include the fully expanded version of the types in the documentation above each type. It is important to me that people be able to figure out what something elaborates to, otherwise everything becomes an opaque tower of type synonyms, and nobody ever notices things like `traverseOf = id`.

PR #737, opened 2017-06-10
Small documentation changes in Control.Lens.Fold

janpath:


src/Control/Lens/Fold.hs
```diff
@@ -424,8 +424,8 @@ takingWhile p l pafb = fmap runMagma . traverse (cosieve pafb) . runTakingWhile
 --
 -- Note: Many uses of this combinator will yield something that meets the types, but not the laws of a valid
 -- 'Traversal' or 'IndexedTraversal'. The 'Traversal' and 'IndexedTraversal' laws are only satisfied if the
--- new values you assign also pass the predicate! Otherwise subsequent traversals will visit fewer elements
--- and 'Traversal' fusion is not sound.
+-- new value you assign to the first target also does not pass the predicate! Otherwise subsequent traversals
```
phadej: This would need an example expecially where changing the second target don't invalidate the laws. Hard to believe.

janpath: The reason why you must not change the targets in a way that changes the predicate is that it would change the elements you traverse over. But as long as the first element for which the predicate fails stays the same the elements traversed over stay the same.

janpath: `dropping 1 . droppingWhile p $ t` should always be lawful for any traversal `t` and any predicate `p`. Eg:

    t  = droppingWhile (<= 1) traverse
    t' = dropping 1 t

    -- over t f . over t g ≠ over t (f . g)
    >>> [1,2,3] & t .~ 0 & t .~ 1
    [1,0,0]
    >>> [1,2,3] & t .~ 1
    [1,1,1]

    -- t' on the other hand behaves lawful
    >>> [1,2,3] & t' .~ 0 & t' .~ 1
    [1,2,1]
    >>> [1,2,3] & t' .~ 1
    [1,2,1]

phadej: I see. Could you add the examples as `doctest`s ?

phadej: I mean, the *values* variant is much more obvious, the quite specific *the first element* needs explanation. Maybe even an advice that not-changing for all values is "better". 

Also plural *new values* could be understood as that the resulting `Traversal` could be used multiple times.

src/Control/Lens/Fold.hs
```diff
@@ -424,8 +424,29 @@ takingWhile p l pafb = fmap runMagma . traverse (cosieve pafb) . runTakingWhile
...
+--
+-- @l@ is not a lawful setter because @`Control.Lens.Setter.over` l f .
+-- `Control.Lens.Setter.over` l g ≢ `Control.Lens.Setter.over` l (f . g)@:
+--
+-- >>> [1,2,3] & l .~ 0 & l .~ 4
+-- [1,0,0]
+-- >>> [1,2,3] & l .~ 4
+-- [1,4,4]
+--
+-- @l'@ on the other hand behaves lawful:
```
RyanGlScott: s/lawful/lawfully

RyanGlScott:
You might wish to rebase on top of master to fix the Travis build.

PR #739, opened 2017-06-12
Fix #643 by threading through more state

### LINKED ISSUE #643: declareFields can generate duplicate Has* classes
``` haskell
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE FunctionalDependencies #-}
{-# LANGUAGE TemplateHaskell #-}
import Control.Lens

declareFields [d|
  data A = A { aA :: Int }
  data B = B { bA :: Int }
 |]
```

Two `class HasA s a` are generated. Separate calls to makeFields / declareFields for each data type don't run into this problem. Using lens-4.13.2.1.

RyanGlScott:
#643 happens because we aren't keeping track of which field classes we've created during an invocation of `declareFields`. This accomplishes that by adding more `State`. In particular, I keep track of a `Set` of field class `Names` which are consulted before creating a new field class, and propagate this information where necessary.

### DIFF
--- a/src/Control/Lens/TH.hs
+++ b/src/Control/Lens/TH.hs
@@ -86,8 +86,8 @@ import Control.Applicative
 #if !(MIN_VERSION_template_haskell(2,7,0))
 import Control.Monad (ap)
 #endif
-import qualified Control.Monad.Trans as Trans
-import Control.Monad.Trans.Writer
+import Control.Monad.State
+import Control.Monad.Writer
 import Control.Lens.Fold
 import Control.Lens.Getter
 import Control.Lens.Lens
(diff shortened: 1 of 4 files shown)


src/Control/Lens/TH.hs
```diff
@@ -86,7 +86,8 @@ import Control.Applicative
 #if !(MIN_VERSION_template_haskell(2,7,0))
 import Control.Monad (ap)
 #endif
-import qualified Control.Monad.Trans as Trans
+import Control.Monad.Trans.Class
+import Control.Monad.Trans.State
```
RyanGlScott: NB: I used lazy `State` here simply because that appeared to be the convention when I arrived here (it already uses lazy `Writer`). I'd have no objections to making it strict, however.

PR #740, opened 2017-06-12
Use travis_wait as a workaround for long test times

RyanGlScott:
Until https://github.com/sol/doctest/pull/165 is released, it'd be nice to have `travis_wait` in place to avoid Travis timing out on `cabal test`, which tends to take slightly over 10 minutes.


RyanGlScott:
For #740.

glguy:
Couldn't we just vendor doctests into lens via cabal.project?

RyanGlScott:
Yes, in hindsight that's a much better idea.

Issue #738, opened 2017-06-10
Travis builds are timing out


RyanGlScott:
The `lens` test suites take longer than you'd expect on Travis CI's build machines, sometimes going for periods of longer than 10 minutes (Travis's timeout limit) without any output. See [here](https://travis-ci.org/ekmett/lens/jobs/241541282#L748) for an example.

We should try using [`travis_wait`](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) to see if that eliminates the problem.

glguy:
Maybe we can make doctests less bad? It seems like it's doing way more work than it should be.

RyanGlScott:
Perhaps, but I think this is to a large degree a symptom of Travis's build machines. For instance, all tests run to completion on my machine in about 3 minutes, whereas it drags on much longer on Travis.

Any tweaks to make the tests faster would also be welcome, of course.

glguy:
To be clear I don't want to change the tests, I want to fix the doctests driver. Most of the test time is spent recompiling things

glguy:
On my computer the tests take over 13 minutes. If I fixed doctests by removing this line linked below they take 1.5 minutes. This line appears to be a workaround for a bug in an old version of GHC.

https://github.com/sol/doctest/blob/master/src/Runner.hs#L129

RyanGlScott:
Nice find, @glguy! Hopefully @sol is open to merging https://github.com/sol/doctest/pull/165, which implements your suggestion.

RyanGlScott:
After talking with @glguy, there are two ideas for bringing down the time on Travis:

1. Use `cabal new-build` and `cabal.project` to vendor in [this fork of `doctests`](https://github.com/RyanGlScott/doctest/commit/72abefc9244fe4be123e541cd75540e2bafecbd2) which is substantially faster. Of course, we'd need to actually port the Travis testing infrastructure over to `cabal new-build` in the first place (see #646 for an attempt at this).
2. `apt-get install hlint` instead of installing `hlint` the library. This will avoid the need to install `haskell-src-exts` and a bunch of other dependencies.

glguy:
I forgot that we weren't using new-build for the lens builds. We don't need to use new-build to vendor doctest

PR #741, opened 2017-06-13
Port .travis.yml over to new-build, and vendor in doctest

RyanGlScott:
This should make the build more robust and (hopefully) prevent the tests from timing out on Travis.

Supersedes #646. Addresses one point of #738.


RyanGlScott:
Bah. Either I did something wrong, or something fishy is happening, since the [doctests are timing out](https://travis-ci.org/ekmett/lens/jobs/242258715#L1324) even after this change...

glguy:
The default hvr .travis.yml overwrites your cabal.project

https://github.com/RyanGlScott/lens/blob/baea18f3bb2915c656526ce4bf81b99dfcf1eb38/.travis.yml#L84

RyanGlScott:
Ack, I deleted one occurrence of that in the default script, but I didn't see that other one! Let's see what happens after I delete _that_ one...

glguy:
I believe that the problem is that the build script tests building an sdist, which isn't including the vendored doctest!

RyanGlScott:
Oy vey. This is what I get for blindly trusting an autogenerated script...

Take three.

glguy:
Is it necessary to build lens twice, we're doing that once with tests disabled and again with them enabled. It just takes 2x as long to do that, right?

glguy:
In my tests GHC 7.6.3 appears to tickle the bug that doctest was working around https://travis-ci.org/ekmett/lens/builds/242294540

RyanGlScott:
We're not actually building the entirety of `lens` twice, since it caches the result of the previous build (without tests and benchmarks).

That being said, it does seem a bit odd to build it in stages like that. I do prefer the old behavior better, where we build the library, tests, and benchmarks all at once, and then run the tests immediately afterwards. If this build succeeds, I'll do that consolidation step next.

glguy:
I watched it slowly build it again, (configuration changed) was cited as the reason, and it was just as slow the second time

RyanGlScott:
Oh, this must be a GHC-specific thing. On [this GHC 8.2.1 run](https://travis-ci.org/ekmett/lens/jobs/242294541), it doesn't exhibit that behavior, but on [this GHC 7.6.3 run](https://travis-ci.org/ekmett/lens/jobs/242294545), it does. Disturbing...

RyanGlScott:
> In my tests GHC 7.6.3 appears to tickle the bug that doctest was working around https://travis-ci.org/ekmett/lens/builds/242294540

Argh, and I had boldly proclaimed that this bug wasn't present anymore in https://github.com/sol/doctest/issues/164#issuecomment-307654699, too...

I suppose that this won't be the cure-all we'd hoped for, then. Any ideas for avoiding timeouts on GHC 7.6 and older?

glguy:
Exactly two tests fail, we can delete them, change them, or CPP them

RyanGlScott:
>Exactly two tests fail, we can delete them, change them, or CPP them

This makes me leery. If that `doctest` patch really is broken on GHC 7.4 and 7.6, then any attempts to conditionally fix tests on our end are inherently fragile. There's no telling when the bug would crop up during a modification of another test, or when introducing new ones.

I've opted to simply guard that `:reload` behind CPP for now (and fix the double-build bug you mentioned in https://github.com/ekmett/lens/pull/741#issuecomment-308017376). Let's see if the timeout happens on old GHCs.

phadej:
The script builds also without tests enabled, because test-suites and bencmarks might (and often do) affect the exact dependencies of the library, e.g. in the case test-dependency has restrictive bound like "it depends on old optparse-applicative, and only newer optparse-applicative allows newest transformers-compat.

RyanGlScott:
That explanation makes sense, @phadej, thanks.

Still, there's a problem that old GHCs (for instance, 7.6) are rebuilding the entire library when you run `cabal new-test` after `cabal new-build` because it thinks the configuration changed. That's pretty bad - I'd almost consider that a deal-breaker for switching over to `new-build`, even. It seems like it might be a `Cabal` bug... does anyone know what's going on here?

RyanGlScott:
Hrm. Even with `cabal-install-head`, the duplicate-build bug is still present. That's frustrating.

RyanGlScott:
An update. After much trial-and-error (many thanks to @hvr for the assistance!), I managed to fix (or at least work around) the double-build bug. So I can confidently state that `new-build` is `lens`-ready, which is nice.

Unfortunately, the ultimate goal of making the `doctest`s not time out is still a problem. I'm using a modified version of `doctest` here, but due to https://github.com/ekmett/lens/pull/741#issuecomment-308017962, we can't apply the speed hack to GHC 7.6 and earlier. As you might expect, this means the GHC 7.4 and 7.6 builds can time out, and this actually happened in the most recent build: https://travis-ci.org/ekmett/lens/jobs/243353190

I don't know how to work around that old `doctest` bug yet, so until I do, I propose that we run `cabal new-test` with `travis_wait`. This means that the GHC 7.4 and 7.6 builds will take longer than the others, but I think we'd still fall under the 60-minute window, which is the maximum amount of time that you can run a single Travis job for. Does that sound agreeable, @glguy?

RyanGlScott:
Actually, it turns out that one can work around the aforementioned GHC 7.4 and 7.6 bug [quite easily](https://github.com/sol/doctest/pull/165#issuecomment-308953288), so after incorporating the fix, all builds are green! Hooray!

PR #747, opened 2017-06-22
whitespace

gwils:


glguy:
Should we do a release now?

RyanGlScott:
This is a pretty important change, so sure! ;)

PR #751, opened 2017-07-15
Add .@~ to Control.Lens.Operators

Peaker:


RyanGlScott:
Thanks!

It looks like `(.@=)` isn't exported from `Control.Lens.Operators` either - do you mind adding that as well while you're in the neighborhood?

Peaker:
Feel free to squash these commits.

PR #755, opened 2017-07-28
Change makeFields instances for fields with type families in their types

### LINKED ISSUE #754: Change instances generated by makeFields to avoid “Illegal type synonym family application” errors
Currently, it’s possible to make `makeFields` generate an instance which is not accepted by GHC if a record field’s type is an application of a type synonym family. For example, this code will produce an error:

```haskell
type family Fam a
data Rec a = Rec { _recFam :: Fam a }
makeFields ''Rec
```
```
• Illegal type synonym family application in instance: Fam a_an7y
• In the instance declaration for
    ‘HasFam (Rec a_an7y) (Fam a_an7y)’
```

I *think* this instance really ought to be legal due to the fundep, but GHC does not allow it. I’ve opened [GHC ticket #14046](https://ghc.haskell.org/trac/ghc/ticket/14046) to track that problem, but (as mentioned in the ticket) there is actually a workaround. Currently, `makeFields` generates an instance like this:

```haskell
instance HasFam (Rec a) (Fam a) where
```

However, the following instance is equivalent, and it is accepted by GHC:

```haskell
instance (b ~ Fam a) => HasFam (Rec a) b where
```

It seems like there would be no reason to not *always* generate instances like that, though it might be theoretically possible to statically check if there are type synonym family applications in the field’s type and only generate the constraint in those situations.

lexi-lambda:
This adjusts the implementation of `makeFields` to resolve #754. The implementation is perhaps not the cleanest, so I would be happy to make suggested improvements. It also adds a new dependency on `th-expand-syns`, and I’m not sure if that’s considered worth the extra dependency.

### DIFF
--- a/src/Control/Lens/Internal/FieldTH.hs
+++ b/src/Control/Lens/Internal/FieldTH.hs
@@ -369,10 +369,31 @@ makeFieldClass defType className methodName =
   s = mkName "s"
   a = mkName "a"
 
+-- | Build an instance for a field. If the field’s type contains any type
+-- families, will produce an equality constraint to avoid a type family
+-- application in the instance head.
 makeFieldInstance :: OpticStab -> Name -> [DecQ] -> DecQ
-makeFieldInstance defType className =
-  instanceD (cxt [])
-    (return (className `conAppsT` [stabToS defType, stabToA defType]))
+makeFieldInstance defType className decs =
+  containsTypeFamilies a >>= pickInstanceDec
+  where
+  s = stabToS defType
+  a = stabToA defType
+
+  containsTypeFamilies = go <=< D.resolveTypeSynonyms
+    where
+    go (ConT nm) = has _FamilyI <$> reify nm
+    go ty = or <$> traverse go (ty ^.. plate)
+
+  pickInstanceDec hasFamilies
+    | hasFamilies = do
+        placeholder <- VarT <$> newName "a"
+        mkInstanceDec
+          [return (D.equalPred placeholder a)]
+          [s, placeholder]
+    | otherwise = mkInstanceDec [] [s, a]
+
+  mkInstanceDec context headTys =
+    instanceD (cxt context) (return (className `conAppsT` headTys)) decs
 
 ------------------------------------------------------------------------
 -- Optic clause generators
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Thanks, @lexi-lambda!

Note that we already depend on `th-abstraction` which has its own [`resolveTypeSynonyms`](http://hackage.haskell.org/package/th-abstraction-0.2.3.0/docs/Language-Haskell-TH-Datatype.html#v:resolveTypeSynonyms) function, so I would recommend using that instead of bringing in the `th-expand-syns` dependency.

lexi-lambda:
Aha, I peeked very briefly at `th-abstraction`, but I missed the `resolveTypeSynonyms` function. I’ve updated the commit to use that instead. Also, it looks like the Travis build failed due to a lack of `EqualityT` pre-ConstraintKinds, so I’ve pushed a fix that adds a compatibility binding using CPP to `Control.Lens.Internal.TH`.

glguy:
@lexi-lambda `th-abstraction` has `equalPred :: Type -> Type -> Pred`

src/Control/Lens/Internal/TH.hs
```diff
@@ -96,6 +96,14 @@ fromSet f x = Map.fromDistinctAscList [ (k,f k) | k <- Set.toAscList x ]
 newNames :: String {- ^ base name -} -> Int {- ^ count -} -> Q [Name]
 newNames base n = sequence [ newName (base++show i) | i <- [1..n] ]
 
+-- | Compatibility shim for pre-ConstraintKinds equality predicates.
+equalT :: Type -> Type -> Pred
```
RyanGlScott: Don't do this. Use [`equalPred`](http://hackage.haskell.org/package/th-abstraction-0.2.3.0/docs/Language-Haskell-TH-Datatype.html#v:equalPred) from `th-abstraction` instead, which is fully backwards compatible.

lexi-lambda:
Thanks, that’s what I get for not looking at all carefully at th-abstraction.

lexi-lambda:
Okay, I’ve updated the commit with the relevant changes, and CI appears to have passed.

RyanGlScott:
Thanks!

Coincidentally, I'm currently preparing a new release, so we should have a version of `lens` out soon that supports this.

lexi-lambda:
Awesome, thanks so much for your work maintaining lens.

PR #757, opened 2017-08-01
Fix `Setter'`'s unintended document code

aiya000:
The document of `Setter'` says `type Setter' = Setter'`,  
but I think it is wrong of `type Setter' = Simple Setter`.

Is this right ? 🤔


RyanGlScott:
Good catch.

Issue #764, opened 2017-09-15
Add support for Semigroup=>Monoid superclass


hvr:
It would be useful to have this sorted out rather sooner than later so we can have a version of `lens` (even if unreleased or it only lives on the new head.hackage repo) which is GHC HEAD ready as we'd like to be able improve "early" continuous integration of GHC HEAD w/ popular Hackage packages; from the IRC channel:

```
hvr:       the problems start already w/
hvr:        instance (Contravariant f, Apply f) => Semigroup (Folding f a) where
hvr:        instance (Contravariant f, Applicative f) => Monoid (Folding f a) where
hvr:       the second instance definition fails to typecheck because it cannot infer Apply f
hvr:       so you'd have to add Apply f to the class constraints of the Monoid instance
hvr:       the same problem w/ a missing Apply constraint then propagates to other places

phadej:    having (..., Apply f, Applicative f) => Monoid (Folding f a) is IMHO good enough

edwardk:   I expect we'll have to do some surgery on lens, and that we'll lose a bit of functionality in the process
edwardk:   The decision will need to be made on each point to worsen support for third party applicatives or to worsen support for relevant traversals
edwardk:   We knew it would come up eventually

phadej:    if there aren't too many such instances, would it make sense to have two variants? one with Apply in both, and other with Applicative in both?
```

/cc @phadej @RyanGlScott @ekmett

bgamari:
I'd like to +1 this; one of the things that came up during the Implementator's Workshop is our lack of smoke-testing against user packages between releases. Keeping packages low on the dependency tree building with `HEAD` will make testing much easier, both between and (perhaps more importantly) directly prior to releases.

hvr:
related: #765

RyanGlScott:
Fixed in #766.

PR #765, opened 2017-09-17
[RFC] Compatibility with GHC 8.4

bgamari:


glguy:
@bgamari needs a fix for the days before Applicative => Monad

src/Control/Lens/Internal/Fold.hs
```diff
@@ -72,8 +72,8 @@ instance (Contravariant f, Applicative f) => Monoid (Folding f a) where
 -- The argument 'a' of the result should not be used!
 newtype Traversed a f = Traversed { getTraversed :: f a }
 
-instance Apply f => Semigroup (Traversed a f) where
-  Traversed ma <> Traversed mb = Traversed (ma .> mb)
+instance Applicative f => Semigroup (Traversed a f) where
```
RyanGlScott: I can't help but feel like changing the `Apply` constraint to `Applicative` here (as well as in the `Semigroup` instance for `Sequenced`) is the wrong choice. After all, `Apply` is intended to be `Applicative` without `pure`, so there are more `Apply` instances than `Applicative` ones, meaning that this change would rule out some programs which would otherwise have been permitted.

My inclination would be to adopt the approach that you took for `Effect` here—leave the `Semigroup` instance alone, and add `Apply` constraints to the `Monoid` instances. This is still a breaking change, but one that's much easier to adjust to, since all `Applicative` instances can trivially be made into `Apply` instances.

bgamari: The problem is that `Apply` instances are much less common than `Applicative`s, meaning that the `Semigroup Traversed` instance will become significantly less useful than it currently is. As mentioned by @ekmett in #764, there is a nasty trade-off here.

RyanGlScott: I'm aware of this. But the tradeoff in the other direction is arguable worse, in my opinion. You're demanding that the `f` in this instance not just be an `Apply`, but an `Applicative`. This completely rules out making `f=IntMap`, for example, since `IntMap` is an `Apply` but can't be an `Applicative`.

It's true that there are some `Applicative` instances that aren't `Apply` instances at the moment, but this is easy to retrofit, as `Apply` is just `Applicative` without `pure`.

glguy: I'm in favor of having the more correct instance using `Apply`. I think that there will be very few actually occurrences where the first argument of one of the `traverseOf_` family of definitions will be constructed using the `Monoid` instance. The work-around would either be to add an `Apply` instance or to write a stand-alone replacement for `mappend` that simply doesn't require the `Apply` instance. It's going to be extremely rare to find some code that both uses `Monoid` to build the `traverseOf_` argument and also requires that that code be built using `Monoid` in particular instead of some local replacement that omits the `Apply` constraint.

bgamari: Fair enough; I can see the argument here. I'm happy to rework this the other way around.

bgamari:
I just pushed a patch implementing the other extreme: pushing `Apply` all the way down through to the consumers. Note however that this affects `itraverse_`, `ifor_`, `imapM_` and `iforM_`.

phadej:
I wonder if this should leave in @hvr's head.hackage for a while, I'd like to test how much this breaks my "real code".

bgamari:
Oleg Grenrus <notifications@github.com> writes:

> I wonder if this should leave in @hvr's head.hackage for a while, I'd like to test how much this breaks my "real code".
>
I suspect it probably should. I'll admit I am quite worried about the
effect of the added constraints.

phadej:
@bgamari, indeed it broke things

```
No instance for (Data.Functor.Bind.Class.Apply (HtmlT Identity)
```

where `HtmlT` is from `lucid`.

At work I can just add an instance to the collection of others `Orphans`, I'd like eventually upstream `Apply` instances.

It will be quite a battle to get people depend on `semigroupoids`. I feel strongly about this, I think `Apply` is a nice class. I'd volunteer to bootstrap `apply-bind` package (with only GHC bundled libs as deps, name open to bikeshedding), and reverse instance dependencies in all of @ekmett packages. Sooner or (but most likely quite) later we'd like to have `Apply` and `Bind` in `base`, and `apply-bind` could be a first step towards that goal.

RyanGlScott:
> It will be quite a battle to get people depend on `semigroupoids`.

I'm lost. If your package depends on `lens`, then it _already_ transitively depends on `semigroupoids`, so no one will have to incur an additional dependency just to add `Apply` instances to make their `lens` code work in the next version. Or am I missing something?

src/Control/Lens/Indexed.hs
```diff
@@ -422,7 +423,7 @@ imapM_ f = liftM skip . getSequenced #. ifoldMap (\i -> Sequenced #. f i)
 -- @
 -- 'Control.Lens.Fold.forMOf_' l a ≡ 'iforMOf' l a '.' 'const'
 -- @
-iforM_ :: (FoldableWithIndex i t, Monad m) => t a -> (i -> a -> m b) -> m ()
+iforM_ :: (FoldableWithIndex i t, Apply m, Monad m) => t a -> (i -> a -> m b) -> m ()
```
RyanGlScott: Looks like this needs an explicit `Applicative` constraint to make `lens` build on old GHCs, as shown in [this](https://travis-ci.org/ekmett/lens/jobs/277127377#L615) failing Travis CI run.

phadej:
@RyanGlScott I use `lens` and `lucid`, I cannot use `lens` combinators with `lucid`'s `HtmlT` monad, because latter doesn't depend on kmettverse at all.

```
semigroupoids
     |
    lens        lucid
       |         |
     olegs-mega-app
```

phadej:
See: https://github.com/ekmett/lens/commit/b1446f7b97a0de5c0c79f73bc4c62b093ed027ea

There I introduce new newtype, so we have now three types: `Traversed`, `TraversedA`, and `Sequence` with `Apply`, `Appplicative` and `Monad` constraining their `Semigroup` instance respectively. That's unfortunate, we'd like to have only one such `newtype`.

RyanGlScott:
To be clear, is the source of breakage due to the additional constraints on `ifor_`/`iforM_`/`itraverse_`/`imapM_`? The additional constraints on the `Monoid` instances for various datatypes in `lens`? Both?

If it's just the former, then as you've noted, we could very easily introduce an additional `newtype` to get the constraints for `ifor_`/`iforM_`/`itraverse_`/`imapM_` just right so as to avoid incurring additional constraints. I'm quite fond of this suggestion, actually!

> That's unfortunate, we'd like to have only one such `newtype`.

I mean, sure. In an ideal world where the dust has long settled after everyone has adjusted to recent superclass changes, we could do away with a lot of these "legacy" newtypes. But we're a long ways away from living in that world, so until then, we've got to accept that some hackery is needed to keep the ship afloat.

But more importantly, all of these newtypes live in an `Internal` module to begin with. I feel no remorse about adding extra crap there, because we advertise up front that we can change the contents of these modules to our hearts' content.

src/Control/Lens/Internal/Fold.hs
```diff
@@ -53,8 +53,8 @@ import qualified Data.List.NonEmpty as NonEmpty
 -- | A 'Monoid' for a 'Contravariant' 'Applicative'.
 newtype Folding f a = Folding { getFolding :: f a }
 
-instance (Contravariant f, Apply f) => Semigroup (Folding f a) where
-  Folding fr <> Folding fs = Folding (fr .> fs)
+instance (Contravariant f, Applicative f) => Semigroup (Folding f a) where
```
RyanGlScott: Can you invert the dependency here as well? I'd rather see the `Monoid` instance occur an `Apply` constraint than have this `Semigroup` instance incur an `Applicative` constraint.

phadej:
@RyanGlScott unfortunately both. One source of breakage was additional constraints in `ifor_`, and that's fixed by new newtypes. **But** there was **also** a breakage due additional constraints on `Monoid` instance, due

```
forOf_ :: Functor f => Getting (Traversed r f) s a -> s -> (a -> f r) -> f ()
```

which for `Fold` dif:

```diff
-forOf_ :: Applicative f            => Fold s a       -> s -> (a -> f r) -> f ()
+forOf_ :: (Apply f, Applicative f) => Fold s a       -> s -> (a -> f r) -> f ()
```

--

There's "easy" fix, we already have

```haskell
forMOf_ :: Monad m => Getting (Sequenced r m) s a -> s -> (a -> m r) -> m ()
```

so we can have

```haskell
forOf_  :: Functor f => Getting (TraversedA r f) s a -> s -> (a -> f r) -> f () -- only Applicative
forFOf_ :: Functor f => Getting (Traversed  r f) s a -> s -> (a -> f r) -> f () -- with Apply for Fold1 etc.
```

`F` comes from `liftF2` etc. But that will mean a few new names, luckily only in `Fold` module if I see correctly.

RyanGlScott:
If you find yourself needing `forFOf_` and the like, it might be worthwhile to just add variants for all the `Traversed` combinators that use `TraversedA` instead. (As well as a lengthy `changelog` note explaining what the deal is.)

src/Control/Lens/Internal/Fold.hs
phadej: It's used for `ifolded` default implementation (and nothing else), so this way is better.

RyanGlScott:
An alternative take on this PR is found at #766.

To the author, @phadej: I'm still woefully confused about the choices of instance context changes, as some data types have their `Monoid` instances changed, and some have their `Semigroup` instances changed, with no apparent explanation of what's going on.

I posit that we should try and keep the changes minimized to just the `Monoid` instances, and you have responded that for some of these datatypes, it doesn't matter, since they're only used internally. Fair enough, but we should make an honest effort to categorize which of these datatypes are public-facing and which aren't. Otherwise, it's difficult for me to gauge whether these changes are correct (and I suspect I'm not alone in this regard).

phadej:
I did best effort using `grep`:

- `Folding` isn't exported from any non-`Internal` module, also it isn't used in any public type-signature. AFAICS it used only to implement default `ifolded`. 
- `Traversed` and `Sequence` (and `TraversedF`) are exported from `Control.Lens.Fold` and are visible in `traverseOf_` family of functions they are used to implement.
- `Effect` is indirectly public thru `Magnified` type-family, but seems like used with `Monad m` only (so there one have still to adjust the constraints).

phadej:
@RyanGlScott I changed `Effect`'s Semigroup instance too in #766. So now all three types have stronger constraint in the `Semigroup` instance, and `Monoid` contexts are unchanged. To cope with genuine `Apply` effects: `traverse1Of_` & co are added, using new `newtype TraversedF`.

IMO the patch is uniform and not too controversial now.

RyanGlScott:
Very well. I'd like to hear what @glguy or @ekmett think about #766 before proceeding.

ekmett:
Thinking about it, I'm more willing to weaken our ability to use `Apply` and `semigroupoids` than I am to lose our compatibility with third party packages that don't supply instances for `Apply`. To that end I'd be willing to weaken all of the things that currently get away with a semigroup/apply constraint combo to make them only work with monoids by default, and then consider adding second class citizens for working with `Apply` for relevant traversals. In particular our combinators shouldn't be pushing `Apply` constraints on users by default.

Yes, incurring the Apply constraints are more correct. Unfortunately, they'd also drastically limit the usability of `lens`.

phadej:
@ekmett i.e. #766 is the way to go?

ekmett:
From what I can see, the combinators in Control.Lens.Indexed pick up unwanted Apply constraints in that patch, so I don't think so.

ekmett:
Or did that get cleaned up in the next patch? I'm on my phone -- hard to tell.

phadej:
@ekmett Are you sure you looked at #766, and not this PR. There e.g. `ifor` works for non-`Apply` effects. I changed all `Apply` to `Applicative` so `Semigroup` instances are less-general than they could be, but OTOH `Monoid` instances don't need `Apply`.

ekmett:
The confusion stemmed from f5f9017 adding `Apply` constraints before you removed them in the next patch. I think #766 is the way to go for now.

phadej:
@RyanGlScott should we merge and have a major-bump release?

RyanGlScott:
@phadej, see my comments in https://github.com/ekmett/lens/pull/766#issuecomment-332189840.

PR #766, opened 2017-09-19
Another take on GHC-8.4

phadej:
A top of https://github.com/ekmett/lens/pull/765

### DIFF
--- a/src/Control/Lens/Internal/Zoom.hs
+++ b/src/Control/Lens/Internal/Zoom.hs
@@ -291,7 +291,7 @@ instance (Apply m, Semigroup r) => Semigroup (Effect m r a) where
   Effect ma <> Effect mb = Effect (liftF2 (<>) ma mb)
   {-# INLINE (<>) #-}
 
-instance (Monad m, Monoid r) => Monoid (Effect m r a) where
+instance (Apply m, Monad m, Monoid r) => Monoid (Effect m r a) where
   mempty = Effect (return mempty)
   {-# INLINE mempty #-}
   Effect ma `mappend` Effect mb = Effect (liftM2 mappend ma mb)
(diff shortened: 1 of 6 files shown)


src/Control/Lens/Internal/Zoom.hs
```diff
@@ -291,7 +291,7 @@ instance (Apply m, Semigroup r) => Semigroup (Effect m r a) where
   Effect ma <> Effect mb = Effect (liftF2 (<>) ma mb)
   {-# INLINE (<>) #-}
 
-instance (Monad m, Monoid r) => Monoid (Effect m r a) where
+instance (Apply m, Monad m, Monoid r) => Monoid (Effect m r a) where
```
phadej: I never used `Zoom` functionality, so I cannot comment on this change. Is here better to add `Monad` to the `Semigroup`, not sure.

phadej: If someone knows any public package where `magnify` or such are used, I'm keen to look and try how this affects them.

RyanGlScott:
I will comment on this PR in #765 to keep the discussion in one place.

RyanGlScott:
Before merging, can you add some extra documentation? Specifically,

* Mention all that's changed in the changelog
* It would be worth putting @ekmett's wisdom from https://github.com/ekmett/lens/pull/765#issuecomment-331927743 into a Note somewhere in `lens`, and referring to it from the relevant instances. That way, we can address potential confusion in the future about these overly strong contexts. Moreover, if we ever make `Apply` a superclass of `Applicative`, we can immediately know which instances we can weaken back to their original forms.

phadej:
I bumped version to `4.16` as this change may potentially break someone.

ekmett:
Merging this. We can do any quibbling about the details with small patches in `master`.

PR #767, opened 2017-09-29
Add head1 and last1

bucklereed:
I'm not sure if this is the final form that this pull request should take, and I have some questions for the lens gurus about it.

1. These lenses are index-oblivious. Not all `Traversable1` instances will be indexable, so maybe that's unavoidable? There could still be `ihead1` and `ilast1`, but I don't think I understand indexed optics well enough to implement them.

2. Should `First1` be exposed in an internal module? I had a look with `git grep` and I couldn't find any other non-exposed implementation-detail types. Would it be better to stick it in a new `Control.Lens.Internal.Lens` and fill in instances, even though I can't see it being very useful to anyone?

3. Placement. They're in `Control.Lens.Lens` because they're lenses and that's where I would look for them. But they don't depend on anything from `lens` (well, unless they get rewritten to be indexable or something), so I think there's an argument for actually having them in `semigroupoids` (which is lower on the dependency graph). Discoverability could go either way, but having them in `semigroupoids` would mean that the doctests would have to go.


bucklereed:
I have another, slightly different, implementation of this that may be nicer, that introduces this type:

```
data AltWriter f a = AltWriter (f a) a

instance (Functor f) => Functor (AltWriter f) where
  fmap f (AltWriter fa a) = AltWriter (f <$> fa) (f a)

instance (Alt f) => Apply (AltWriter f) where
  AltWriter ff f <.> AltWriter fx x = AltWriter ((($ x) <$> ff) <!> (f <$> fx)) (f x)

instance (Plus f) => Applicative (AltWriter f) where
  pure = AltWriter zero
  (<*>) = (<.>)
```

With semigroupoids#64, `First1 f` in the current pull is then the same as `AltWriter (Compose First f)`.

ekmett:
I'll want to do a quick pass and remove the random prophylactic ()'s around the contexts to match the surrounding style, but otherwise I have no objection to these existing.

PR #770, opened 2017-10-18
Add Control.Lens.Unsound

phadej:


src/Control/Lens/Unsound.hs
```diff
@@ -0,0 +1,92 @@
...
+
+#if __GLASGOW_HASKELL__ < 708
+{-# LANGUAGE Trustworthy #-}
+#endif
+
+{-# LANGUAGE RankNTypes #-}
+
+-------------------------------------------------------------------------------
+-- |
+-- Module      :  Control.Lens.Unsafe
```
Taneb: I'm like 995 sure this should be Control.Lens.Unsound

phadej: fixed

phadej:
finally green.

phadej:
@ekmett any comments on this, should I merge?

ekmett:
Let's start here and keep dumping things in there. There are similar products for things like traversals, etc.

PR #773, opened 2017-10-29
Make dumped splices of PrismTH compilable

ddssff:
I'm still dumping splices from ghc so that ghcjs compiles don't take forever.  This patch prevents compile errors when doing that.  See also issue #696 and #602.

### DIFF
--- a/src/Control/Lens/Internal/PrismTH.hs
+++ b/src/Control/Lens/Internal/PrismTH.hs
@@ -322,7 +322,7 @@ makeReviewer conName fields =
 -- ) :: s -> Either s a
 makeSimpleRemitter :: Name -> Int -> ExpQ
 makeSimpleRemitter conName fields =
-  do x  <- newName "x"
+  do x  <- newName "xb"
      xs <- newNames "y" fields
      let matches =
            [ match (conP conName (map varP xs))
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I'm confused as to what the purpose of this patch is—shouldn't `newName` suffice to make names globally unique? What compiler errors does this fix? (Sorry if I'm being dense here, as I don't have a GHCJS installation handy to test this.)

glguy:
It's so that if you print out the splices, and presumably turn off the printing of unique names that the names are still unique. It's a hack, but I don't suppose it hurts anything to allow it.

glguy:
@ddssff What I'm confused about is: What was the name `x` clashing with before?

ddssff:
I will look - I made that patch a year ago but never got around to sending it.

ddssff:
@RyanGlScott - Running template haskell is very slow under ghcjs, so I run it under ghc and have it write the result to a file, and check that into git so that ghcjs can compile it later.  I can't keep the newName disimbiguation because then the output would change every time ghc compiled it.  So then a patch would have to go into git and ghcjs would have to rebuild every time ghc compiled the code.  GHCJS takes a long time even to compile this particular template haskell output.

ddssff:
@glguy - I'm sorry, this was only an issue for lens < 4.15.1

ddssff:
@glguy - would you like me to submit a pull request to revert it?

ddssff:
Thanks Eric.

PR #777, opened 2017-12-15
Add Monoid (f a) => Monoid (Indexing f a) instance

phadej:


phadej:
@ekmett Can you take a look, is this instance ok? It was motivated by a question on IRC.

PR #778, opened 2017-12-17
Add traversalUnion to Control.Lens.Unsound

Gurkenglas:
From my position, "Traversing the same entry twice is not lawful." is hearsay: Ctrl-F `no duplicates` on https://artyom.me/lens-over-tea-2

This implementation looks to me like an inefficient proof of concept.


ekmett:
Not keen on the name and may change it, but the functionality is definitely worth adding.

PR #785, opened 2018-02-08
Add operators for Consing and Snocing onto a setter

glittershark:
Add `|>~`, `|>=`, `<|~`, and `<|=` operators for `cons`ing and `snoc`ing onto the target of ASetter both inside and outside state.

Because more operators is just what the Lens library needs, right?

Jokes aside, I went looking for these and was surprised I wasn't able to find them, so I figured I'd submit them.

### DIFF
--- a/src/Control/Lens/Cons.hs
+++ b/src/Control/Lens/Cons.hs
@@ -26,7 +26,7 @@ module Control.Lens.Cons
   (
   -- * Cons
     Cons(..)
-  , (<|)
+  , (<|), (<|~), (<<|~), (<|=), (<<|=)
   , cons
   , uncons
   , _head, _tail
@@ -35,7 +35,7 @@ module Control.Lens.Cons
 #endif
   -- * Snoc
   , Snoc(..)
-  , (|>)
+  , (|>), (|>~), (<|>~), (|>=), (<|>=)
   , snoc
   , unsnoc
   , _init, _last
@@ -46,12 +46,15 @@ module Control.Lens.Cons
   ) where
 
 import Control.Lens.Equality (simply)
+import Control.Lens.Setter
 import Control.Lens.Fold
 import Control.Lens.Prism
 import Control.Lens.Review
 import Control.Lens.Tuple
 import Control.Lens.Type
 import Control.Lens.Internal.Coerce
+import qualified Control.Monad.State.Class as State
+import           Control.Monad.State.Class (MonadState)
 import qualified Data.ByteString      as StrictB
 import qualified Data.ByteString.Lazy as LazyB
 import           Data.Monoid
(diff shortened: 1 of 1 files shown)


Taneb:
Could you also add `<<|~` and `<<|=` to be consistent with the other similar operators? (eg. `+~` and friends)

glittershark:
Happy to! The `snoc` variant is interesting, though, as `<|>` is already a pretty familiar operator...

glittershark:
@Taneb force-pushed with the 4 extra variants 😄

glguy:
`<|>~` is unfortunately named considering the more common `<|>` operator. It seems like simply using operator slices and `%~` would lead to clearer code than adding yet more cute operator combinations.

glittershark:
Yeah, I'm honestly neutral on the `<` variants of the operators myself - I can't think of a time I've ever reached for them.

ekmett:
FWIW- other than the `%~` and `.~` variants, which relate to 'mod' and := respectively, we basically didn't make any "new" 'base' operators up, for assignment so much as mangle some existing ones in a consistent pattern. The reasons that an operator for `cons=` doesn't exist is that a.) it requires us to make up a starting symbol for users to memorize, outside of our usual practice, b.) it requires users to memorize the fact that unlike every other operator it'd have to apply flipped in order to typecheck and c.) if we obviate a by adopting the use of `<|` and `|>` from Control.Lens.Cons we run into the problem doesn't help that `<|>=` and `<<|>=` would be ambiguous under our naming convention. They could be referring to Alternative via <|> or to "snoc and return the new result" as a modification of `|>`.

Given the alternative is to `%~ (x:)`, which clearly states intent and does fall within the vocabulary we delimited, we deliberately backed off on adding cons in operator form. Not saying the current reasoning is perfect, just giving an insight into what thought that _was_ put into the status quo before we run off in a flurry of "ooh me too"s and change it.

glittershark:
Thanks for the background @ekmett - useful to know that it was a conscious omission rather than just having been forgotten about.
At this point I honestly could go either way - the operators are little more than a tiny convenience and as you point out `%~ (x <|)`is barely more typing than the operators introduced in this PR. That said:

- `<|` and `|>` *do* have at least a little bit of precedent, as I'm sure you know, from the `Seq` library, so they fit at least somewhat into the principle of not introducing new operators, only new *combinations* of operators

- the confusion around `<|>~` with `Alternative` is pretty easily worked around by just *not including* the "return the new result" variant of the `cons=` operators - as I mentioned above I personally don't ever really feel a need for those variants of the operators, though obviously anecdotal evidence isn't really worth much.

Again, totally fine with a decision not to merge this, just wanted to provide a little perspective from my side.

ekmett:
One could make up an arbitrary rule about which one wins between `<|> ~` and `< |> ~`, but it'd be just that, an arbitrary rule, and another thing for folks to memorize. 

`%~ cons x` is terse enough that the rule doesn't cross the Fairbairn threshold in my eyes.

ekmett:
Letting this drop for the stated reasons.

PR #789, opened 2018-02-22
Reimplement holesOf

### LINKED ISSUE #788: Consider a cleaner holesOf implementation
The implementation of `holesOf` is not the cleanest thing. In particular, it uses `unconsWithDefault` in a context where the default will never be chosen, because it loses track of the length of the traversal. I have no idea how to deal with the current `holesOf` type, but I figured I would demonstrate that it's possible to implement a reasonably efficient old-school `holesOf` without that particular ugliness.

I'll start with a way to reassociate `Applicative` operations.

```haskell
newtype Grump f a = Grump {runGrump :: forall b c. (a -> b -> c) -> f b -> f c}

instance Functor (Grump f) where
  fmap f (Grump g) = Grump $ \h xs -> g (h . f) xs

instance Functor f => Applicative (Grump f) where
  pure x = Grump $ \g xs -> fmap (g x) xs
  liftA2 f xs ys = Grump $
    \g ps -> runGrump xs (\a ~(m,n) -> g (f a m) n) (runGrump ys (,) ps)

liftGrump :: Applicative f => f a -> Grump f a
liftGrump xs = Grump $ \f ys -> liftA2 f xs ys

lowerGrump :: Applicative f => Grump f a -> f a
lowerGrump gr = runGrump gr const (pure ())
```

Now we can make a version of the classic `FunList` with a little `Coyoneda` embedded to make it lazier:

```haskell
data FL a t where
  D :: t -> FL a t
  M :: (a -> r -> t) -> a -> FL a r -> FL a t

runFL :: FL a t -> t
runFL (D t) = t
runFL (M g a xs) = g a (runFL xs)

instance Functor (FL a) where
  fmap f (D t) = D (f t)
  fmap f (M g a ys) = M (\a' r -> f (g a' r)) a ys

instance Applicative (FL a) where
  pure = D
  liftA2 f (D x) ys = f x <$> ys
  liftA2 f (M g a l) ys =
    M (\a' ~(r,b) -> f (g a' r) b) a $ liftA2 (,) l ys
```

Now we get to the meat of the matter:

```haskell
bp :: a -> FL a a
bp a = M const a (D ())

hl :: FL a t -> [Context a a t]
hl = go id
  where
    go :: (t -> u) -> FL a t -> [Context a a u]
    go _ (D _) = []
    go f (M g a as) = Context (\a' -> f $ g a' $ runFL as) a :
         go (\h -> f (g a h)) as

holesOf :: forall s t a. LensLike (Bazaar a a) s t a a -> s -> [Context a a t]
holesOf f s = hl . lowerGrump $ runBazaar (f sell s) (liftGrump . bp)
-- We use Grump to avoid badly-associated FL appends.
````

### PR DESCRIPTION
**Pull request by treeowl**
Use a more principled implementation.

Fixes #788

### DIFF
--- a/src/Control/Lens/Traversal.hs
+++ b/src/Control/Lens/Traversal.hs
@@ -151,6 +151,7 @@ import Data.Foldable (Foldable)
 #endif
 import Data.Functor.Compose
 import Data.Functor.Day.Curried
+import Data.Functor.Identity
 import Data.Functor.Yoneda
 import Data.Int
 import Data.IntMap as IntMap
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by treeowl** [comment]
I wish I understood why the lazy pair in `liftA2` for `Dayly` seems to be both necessary and sufficient, while a similarly placed one in `liftA2` for `Listed` is not sufficient and doesn't seem to be necessary. This twists my brain a bit.

**Comment by treeowl** [comment]
By the way, I haven't really done a proper job of figuring out what should be inlined and how. I'm also a bit annoyed with `Conjoined`. It offers a way to optimize for `(->)`, which I use because I can, but I'd be much more interested in optimizing for `Indexed`, which I believe I can only do using rewrite rules. The trouble is that I'm not allowed to force the `(index, value)` pair when I change the value; that would be too strict for the function case. Furthermore, a list of indices and values should really look like

```haskell
data L i v = N | C i v (L i v)
```

but we're pretty much forced into something like `[(i, v)]`.

treeowl:
@ElvishJerricco, I think you may find this idea interesting.

ekmett:
`Conjoined` was mostly to make it so we didn't pay through the nose when not using indices, as it was often several times more expensive to use `traversed` than to `traverse`, etc. I'm open to ideas for how to make it more efficient for the indexed case.

treeowl:
I guess you can probably avoid some space leaks in the indexed case by using something like

```haskell
data IList p a = INil | ICons (Corep p (a, IList p a))
```

But of course that's wretched for the non-indexed case, and has all sorts of other likely downsides. A bifunctor version of `Corep` would be marginally better for that, but this is all fairly horrible.

PR #792, opened 2018-02-27
Add class Rixed ... where rix :: Index m => Lens' m (IxValue m)

phadej:
This is being asked once in a while

- Also https://hackage.haskell.org/package/total-map can be an instance of `Rixed`
- `adjunctions` is in dependency closure already (`lens` depends on `kan-extensions`)
- between `riix` and `irix` I chose former.

Why `rix`? Because it's short and *most* instances are `Representable` functors.

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -184,6 +184,7 @@ flag j
 
 library
   build-depends:
+    adjunctions               >= 4.2      && < 5,
     array                     >= 0.3.0.2  && < 0.6,
     base                      >= 4.5      && < 5,
     base-orphans              >= 0.5.2    && < 1,
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -47,6 +47,10 @@ module Control.Lens.At
   , Ixed(ix)
   , ixAt
   , iix
+  -- * Rixed
+  , Rixed (..)
+  , rixRep
+  , riix
   -- * Contains
   , Contains(contains)
   , icontains
@@ -58,6 +62,7 @@ import Control.Lens.Lens
 import Control.Lens.Setter
 import Control.Lens.Type
 import Control.Lens.Indexed
+import Data.Functor.Rep as Rep (Representable (..))
 import Data.Array.IArray as Array
 import Data.Array.Unboxed
 import Data.ByteString as StrictB
@@ -123,6 +128,7 @@ type instance Index LazyB.ByteString = Int64
 -- $setup
 -- >>> :set -XNoOverloadedStrings
 -- >>> import Control.Lens
+-- >>> import Data.Char
 -- >>> import Debug.SimpleReflect.Expr
 -- >>> import Debug.SimpleReflect.Vars as Vars hiding (f,g)
 -- >>> let f  :: Expr -> Expr; f = Debug.SimpleReflect.Vars.f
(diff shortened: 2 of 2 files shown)


phadej:
It's my take on #695 ping @treeowl . @fumieval I use both `a -> b` and `TotalMap k v` (likewise type) instances at work. It's also asked on IRC

treeowl:
The name isn't terrifically clear, but I'll take it by any name.

phadej:
I agree that name isn't the best one, but i'd prefer 3-4 character long name

phadej:
About the name, another variant is to have `lix` for `Lens` `ix`. We *could* have `Prism` variant too, but I don't know it would have any use.

```haskell
type instance Index (Either a b) = Bool
type instance IxValue (Either a b) = a
instance a ~ b => Pixed (Either a b) where
  pix False = _Left
  pix True  = _Right
```

And similarly `Identity a` `Boring a => a -> b` could have `Iso` variant (where this hungarian notation will break :), but that's not so useful anyway.

phadej:
@ekmett Do you have opinions on a name?

ekmett:
Knee jerk reaction: I think the name shouldn't involve `Representable` or `Sieve"`as both carry connotations that imply that the thing is _only_ a function space -- the former via an isomorphism, and the latter via the requirement that it be a sub-functor of `(->) e` for some `e` -- so to me the "R" is out.

Name-wise, I'd prefer to find a clever whole word rather than randomly prefix in order to fit the surrounding code style. I don't do Hungarian notation when I can help it. Simonyi has no power over me! =)

phadej:
Maybe:

```haskell
toUpper ^. pick 'd'
```

English is not my strongest language :/

xplat:
Actually, couldn't this just be added to Ixed by using an associated type synonym or two with a default?

ekmett:
@xplat: That doesn't work very well without `QuantifiedConstraints`. You really need the associated type synonym to know that it is implied by `Applicative`, otherwise you drastically worsen the experience for any user trying to quantify over `Ixed` instances.

Assuming `QuantifiedConstraints` makes it into 8.6 you'd be able to use

```haskell
class (forall f. Applicative f => IxedC m f) => Ixed m where
  type IxedC m :: (* -> *) -> Constraint
  type IxedC m = Applicative
  ix :: IxedC m f => Index m -> LensLike' f m (IxValue m)
```

Then you can refine that point-wise for particular choices of `t` to Functor, without worsening the user experience for the normal case. Mind you its several years out from being viable to change `lens` in this fashion.

The moral equivalent of that without `QuantifiedConstraints` still involves using another name to get better inference, and takes a crapload of boilerplate to do right:

https://github.com/ekmett/lens-constraints/blob/master/Constraints.hs#L105

xplat:
Ouch, good points.  Another alternative is

```haskell
class IxedIn c m where
  ix :: c f => Index m -> LensLike' f m (IxValue m)

type Ixed = IxedIn Applicative
```

but that would break existing instance declarations ...

ekmett:
That latter version has the problem that every combinator you use `Ixed` with infects your type signatures. If you use it heavily your types get downright silly. That is the way the classes in `Control.Exception.Lens` used to work, before we decided to rip it all out.

xplat:
Well, breaking instances was enough to doom it anyway without some further insight ...

phadej:
Would *projection* be better?

```haskell
class Ixed m => Projection m where
    proj :: Index m -> Lens' m (IxValue m)
```

or does it have wrong connotations as well?

ekmett:
I think `Projection` has funny connotations as well.

Given that this sort of wallowed for a year or so, maybe a better path forward would be for you to build the class directly into your package and then we can revisit moving it upstream at some future date.

This removes (or at least postpones) bikeshedding concerns and lets you get something done without `lens`, which has become remarkably stable of late, expanding significantly in scope.

phadej:
@ekmett, I do have such class in the application code, and as I didn't need it elsewhere (yet?) I can wait till we come up with a better name

ekmett:
Closing this more definitively as wontfix to help clear the queue.

PR #794, opened 2018-02-28
Reimplement holesOf again

### LINKED ISSUE #791: Consider reimplementing holesOf again
I believe we can implement `holesOf` using a type very much like `Magma`. I strongly suspect this will improve performance somewhat in the difficult case of a structure that leans hard to the left. The technique also offers an intriguing possibility: a version of `holesOf` that tries to share subtrees of its results. The latter wouldn't replace `holesOf`, because it would only be a win sometimes, but I suspect it would be a very big win when it worked properly.

treeowl:
* Noah "Rampion" Easterly figured out a way to improve on an
  idea I had for implementing a `holes` function, and the relevant
  techniques are equally applicable to the simpler context of
  `holesOf`. Notably, the modified containers will typically share
  a lot of structure with each other.

* Move `holesOf` so it's not between the `parts` and `singular`
  functions and their internals.

Fixes #791

### DIFF
--- a/src/Control/Lens/Traversal.hs
+++ b/src/Control/Lens/Traversal.hs
@@ -151,7 +151,6 @@ import Data.Foldable (Foldable)
 #endif
 import Data.Functor.Compose
 import Data.Functor.Day.Curried
-import Data.Functor.Identity
 import Data.Functor.Yoneda
 import Data.Int
 import Data.IntMap as IntMap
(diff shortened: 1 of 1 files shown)


treeowl:
@rampion, does this look good to you?

treeowl:
Actually, I think I *might* have found a better way. Let me see where that goes.

treeowl:
Yup, I got a better one.

PR #795, opened 2018-03-01
The last holesOf you'll ever need

### LINKED ISSUE #791: Consider reimplementing holesOf again

treeowl:
Reimplement `holesOf` to check all the boxes:

1. No incomplete pattern match.

2. No sketchy `Applicative` instance.

3. Maximal sharing of results.

4. High speed.

Fixes #791

### DIFF
--- a/src/Control/Lens/Traversal.hs
+++ b/src/Control/Lens/Traversal.hs
@@ -151,7 +151,6 @@ import Data.Foldable (Foldable)
 #endif
 import Data.Functor.Compose
 import Data.Functor.Day.Curried
-import Data.Functor.Identity
 import Data.Functor.Yoneda
 import Data.Int
 import Data.IntMap as IntMap
@@ -167,7 +166,6 @@ import Data.Profunctor.Unsafe
 import Data.Reflection
 import Data.Semigroup.Traversable
 import Data.Semigroup.Bitraversable
-import Data.Tagged
 import Data.Traversable
 import Data.Tuple (swap)
 import GHC.Magic (inline)
@@ -506,7 +504,7 @@ iloci f w = getCompose (runBazaar w (Compose #. Indexed (\i -> fmap (indexed sel
 {-# INLINE iloci #-}
 
 -------------------------------------------------------------------------------
--- Parts and Holes
+-- Parts
 -------------------------------------------------------------------------------
 
 -- | 'partsOf' turns a 'Traversal' into a 'Lens' that resembles an early version of the 'Data.Data.Lens.uniplate' (or 'Data.Data.Lens.biplate') type.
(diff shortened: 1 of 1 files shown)


src/Control/Lens/Traversal.hs
```diff
@@ -820,6 +701,74 @@ unconsWithDefault d []     = (d,[])
...
+-- implementation of a related holes function: see
+-- https://stackoverflow.com/a/49001904/1477667.
+
+newtype Holes p a x = Holes
+  { runHoles :: forall t. (x -> t) -> (Endo [Pretext p a a t], x) }
+
+oneb :: forall p a. (Corepresentable p, Category p) => Corep p a -> Holes p a a
+oneb x = Holes $ \xt -> ( Endo (fmap xt (cosieve sell x) :)
+                        , cosieve (id :: p a a) x)
+
```
treeowl: One question: do we need quite so much laziness in the instances? Can we match strictly on the pair in `fmap`, and the first of the two in each of `(<*>)` and `liftA2`? It seems likely, since we're making a list, but this stuff is a bit confusing!

treeowl:
Closing temporarily.

treeowl:
My only remaining question is whether we can be very slightly stricter in the `Functor` and `Applicative` instances because we're making a list in the end. Can we pattern match eagerly on the pair in `fmap`? Can we pattern match eagerly on the first pairs in `(<*>)` and `liftA2`?

treeowl:
I think we *can't* be stricter.

```haskell
fs <*> xs = Holes $ \xt ->
  let
   (pf, pv) = runHoles fs (xt . ($ qv))
   (qf, qv) = runHoles xs (xt . pv)
  in (pf `mappend` qf, pv qv)
```

We can use `qv` without needing `qf`. So we'd better not be too strict in the recursive call.

bitemyapp:
@treeowl The way you blended Easterly's and Cheplyaka's solutions is really interesting, thank you for writing this up.

treeowl:
@bitemyapp, thanks! I only wish I knew of a way to blend them for the problem they were actually trying to solve! That one's a bit tougher.

ekmett:
I _really_ like this version.

treeowl:
Thanks!

ekmett:
My challenge to you now is to see if you can repeat this feat and kill the illegal `Molten`/`Mafic` types in the Internal.Magma module with the same trick. That would make me feel much better about the instances we expose.

treeowl:
It doesn't work for everything. I suspect we can get a lot of bang out of the `traverse2` I proposed for `bifunctors` today, but I don't think we can implement that without something molten.

treeowl:
Where's the bang come from? `Holes t` is `Biapplicative`.

PR #796, opened 2018-03-02
Use superclass

treeowl:
I missed the fact that `Comonad (Corep p)` was a superclass of
`Conjoined p`. That shortens `holesOf` slightly.


treeowl:
By the way: if we want, we can weaken the constraint on `holesOf` from `Conjoined p` to `(Corepresentable p, Comonad (Corep p)`. I don't know if that would be useful.

PR #801, opened 2018-04-01
Consume (Over (Indexed i)) instead of (IndexedLensLike i)

Lysxia:
Improves type inference. For example, this no longer throws an error:

    (%%@~) traversed
      -- inferred type:
      :: (Applicative f, Traversable t) => (Int -> a -> f b) -> t a -> f (t b)


ekmett:
We should never be consuming a rank-2 type like that. Good catch!

Issue #806, opened 2018-06-05
filteredHas combinator


yairchu:
I'm suggesting this combinator:

    filteredHas :: Lens.Fold s i -> Lens.IndexedTraversal' i s s
    filteredHas fold f val =
        case val ^? fold of
        Nothing -> pure val
        Just proof -> Lens.indexed f proof val

It can replace code like this:

    Lens.filtered (Lens.has part) %~ \x -> something (x ^?! part) x

With `filteredHas part %@~ something`. I think it's quite useful and am using it on an expression tree sugaring pass.

Disclaimer: It's quite possible it already exists and I just haven't found it, or has a very succinct way to write it that makes it unnecessary. I often have difficulty finding combinators in `lens`.

ekmett:
It doesn’t really pass the Fairbairn threshold for me, but it’s come up enough that I’m open to it.

Maybe filteredBy would read better?

yairchu:
`filteredBy` LGTM, shall I create a PR?

ekmett:
The problem with `filteredBy` taking `(a -> Maybe b)` is that then the common usecase still requires a call to `preview` inside it, exploding it back to the same number of "words" has before with `has`.

ekmett:
`filteredBy (preview x)` is longer than `filtered (has x)`

glguy:
That's true that it's longer, but the `filteredBy` version is supposedly more useful because it puts the extracted value into the index. Access to that value in the index is the only reason I think it's worth defining this in the first place.

ekmett:
@yairchu: Subject to the two bullet points raised by @glguy, sure.

ekmett:
I've been asked for some flavor of this about every 2-3 months for a couple of years now. Exposing the filtered part as an index finally gave me enough of an excuse to say yes. ;)

yairchu:
I'm stuck on the PR, due to having trouble to make the types work for this simple example which I came up with for the doc-test:

    >>> [(Just 2, 3), (Nothing, 4)] <&> filteredBy (_1 . _Just) . _2 %@~ (*)

This neither works with the type mentioned above nor with the more general

    filteredBy :: (Indexable i p, Applicative f) => Getting (First i) a i -> p a (f a) -> a -> f a

I'm afraid I haven't yet mastered the different lens types and need some advice here..

(Note that I have been successfully using this combinator in my own code and haven't noticed this issue)

yairchu:
@glguy oh oops :) thanks!

[...]

PR #807, opened 2018-06-07
add filteredBy

yairchu:
see discussion at https://github.com/ekmett/lens/issues/806

### DIFF
--- a/src/Control/Lens/Fold.hs
+++ b/src/Control/Lens/Fold.hs
@@ -68,6 +68,7 @@ module Control.Lens.Fold
   , unfolded
   , iterated
   , filtered
+  , filteredBy
   , backwards
   , repeated
   , replicated
@@ -163,7 +164,7 @@ import Control.Monad.Reader
 import Control.Monad.State
 import Data.CallStack
 import Data.Foldable
-import Data.Functor.Apply
+import Data.Functor.Apply hiding ((<.))
 import Data.Functor.Compose
 import Data.Functor.Contravariant
 import Data.Int (Int64)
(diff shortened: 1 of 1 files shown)


src/Control/Lens/Fold.hs
```diff
@@ -361,6 +362,24 @@ filtered :: (Choice p, Applicative f) => (a -> Bool) -> Optic' p f a a
 filtered p = dimap (\x -> if p x then Right x else Left x) (either pure id) . right'
 {-# INLINE filtered #-}
 
+-- | Obtain a potentially empty 'IndexedFold' by taking the first element from another,
+-- potentially empty `Fold` and using it as an index.
+--
+-- The resulting optic can be composed with to filter another 'Lens', 'Iso', 'Getter', 'Fold' (or 'Traversal').
+--
+-- >>> [(Just 2, 3 :: Int), (Nothing, 4)] <&> filteredBy (_1 . _Just) <. _2 %@~ (*)
```
phadej: Can we change to `& mapped . filteredBy ...`, that would be more lens-y. (at most one fancy operator per example: here it's definitely `%@~` :)

src/Control/Lens/Fold.hs
```diff
@@ -361,6 +362,24 @@ filtered :: (Choice p, Applicative f) => (a -> Bool) -> Optic' p f a a
 filtered p = dimap (\x -> if p x then Right x else Left x) (either pure id) . right'
 {-# INLINE filtered #-}
 
+-- | Obtain a potentially empty 'IndexedFold' by taking the first element from another,
```
phadej: should it mention `IndexedTraversal`, as the more concrete type signature uses it?

PR #812, opened 2018-07-04
Add comment about `each` to `both`

yoricksijsling:
Hopefully this improves discoverability of the `Each` module. It took me quite some time before i found out that this is how you can generalize `both` to arbitrary tuples.

### DIFF
--- a/src/Control/Lens/Traversal.hs
+++ b/src/Control/Lens/Traversal.hs
@@ -779,7 +779,8 @@ instance Monoid m => Applicative (Holes t m) where
 
 -- | Traverse both parts of a 'Bitraversable' container with matching types.
 --
--- Usually that type will be a pair.
+-- Usually that type will be a pair. Use 'Control.Lens.Each.each' to traverse
+-- the elements of arbitrary homogeneous tuples.
 --
 -- >>> (1,2) & both *~ 10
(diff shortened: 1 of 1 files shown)


src/Control/Lens/Traversal.hs
```diff
@@ -779,7 +779,8 @@ instance Monoid m => Applicative (Holes t m) where
 
 -- | Traverse both parts of a 'Bitraversable' container with matching types.
 --
--- Usually that type will be a pair.
+-- Usually that type will be a pair. Use 'Control.Lens.Each.each' to traverse
```
phadej: I'd add this after last example, repeating the example:

```haskell
-- Use 'Control.Lens.Each.each' to traverse the elements of arbitrary homogeneous tuples.
-- >>> ("hello", "world","!") ^. each
-- "helloworld!"
```

phadej: And as we on it, I'd also add an example using `Either`, so we don't get impressions this is useful for tuples only.

phadej: ping @yoricksijsling

ekmett:
Merging, we can fix up any remaining wibbles in-repo.

PR #814, opened 2018-08-09
Only incur semigroups and void dependencies on old GHCs.

quasicomputational:
With other recent changes in dependencies, now it should be possible
to depend on `lens` without incurring a dependency on either of those!

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,3 +1,7 @@
+Next
+----
+* Only incur `semigroups` and `void` dependencies on old GHCs.
+
 4.17 [2018.07.03]
 -----------------
 * Allow building with GHC 8.6.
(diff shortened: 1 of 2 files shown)


quasicomputational:
Yep, that's what I've been doing (though not with that exact workflow, which I might have to try next time). ekmett/comonad#44 got merged yesterday, amongst other things (thanks, Ryan!), and I've got PRs for the rest today. Once the dust settles I'll see if I missed anything.

phadej:
@quasicomputational I updated dep graphs on https://github.com/phadej/acme-kmett, feel free to ping me to update them again. (I track `master`s there, not released versions).

Currently it looks like `bifunctors` are the next in row to drop` semigroups` dependency.

PR #823, opened 2018-11-05
Add .png files to extra-doc-files

RyanGlScott:
This allows us to drastically simplify Setup.hs, as we no longer need to manually copy over these PNG files during the Haddock phase.

This feature requires `Cabal-1.18` or later, so I needed to sprinkle in uses of `default-language: Haskell2010` in order to please `cabal check`.

cc @phadej—can you think of anything bad that could happen with this approach?

### DIFF
--- a/Setup.lhs
+++ b/Setup.lhs
@@ -1,24 +1,17 @@
 \begin{code}
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wall #-}
 module Main (main) where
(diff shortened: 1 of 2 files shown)


phadej:
@RyanGlScott if you tried `cabal new-haddock` locally, and the images are where they should be, then that's good enough for me. (bonus points for `stack haddock`).

RyanGlScott:
I've tried `cabal new-haddock`, `cabal v1-haddock`, and `stack haddock`, and they all appear to do the right thing. Let's do this.

Issue #759, opened 2017-08-15
Add `getter = to` alias


phadej:
`getter` would be in line with `lens` and `prism`, (`traversal = id`). Something prevents me of re-exporting `to` from our internal prelude, but we have `getter`.

```haskell
foo :: Getter Bar Fooo
foo = getter $ \b -> b ^. fo <> b ^. oo
```

I can make a PR if this change is ok.

emilypi:
We could probably close this en lieu of the discussion here: https://github.com/ekmett/lens/pull/830

PR #830, opened 2019-01-21
Add 'getter'

emilypi:
See #759

### DIFF
--- a/src/Control/Lens/Getter.hs
+++ b/src/Control/Lens/Getter.hs
@@ -73,6 +73,7 @@ module Control.Lens.Getter
   , ito
   , like
   , ilike
+  , getter
   -- * Combinators for Getters and Folds
   , (^.)
   , view, views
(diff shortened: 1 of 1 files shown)


phadej:
This is a synonym of `to`, isn't it?

I sympahetise the addition, I have used `getter` name in own developments

emilypi:
@phadej yes - i could probably just point to `to` instead of expanding the statement. Would you prefer it just be `getter = to`?

ekmett:
I'm somewhat inclined to just keep `to` as the only name for this. It is shorter, we already took it, and adding getter means more random spurious conflicts and local name shadowing complaints as it is a reasonable local name people might use locally in their own code.

phadej:
True. I imagine there are a lot of code like

```haskell
myFancyLens = lens getter setter where
  getter = ...
  setter = ...
```

emilypi:
Thanks @ekmett I hadn't thought of that. I guess @phadej you'll have to do wtih a local alias 😄

emilypi:
So should we close this?

PR #835, opened 2019-02-15
Use <|> rather than mplus in rewrite docs

Taneb:


src/Control/Lens/Plated.hs
```diff
@@ -317,7 +317,7 @@ children = toListOf plate
 --
 -- Usually 'transform' is more appropriate, but 'rewrite' can give better
 -- compositionality. Given two single transformations @f@ and @g@, you can
--- construct @\a -> f a `mplus` g a@ which performs both rewrites until a fixed point.
+-- construct @\a -> f a <|> g a@ which performs both rewrites until a fixed point.
```
ocharles: Could you also fix the `\` here? It needs to be escaped.

src/Control/Lens/Plated.hs
```diff
@@ -331,7 +331,7 @@ rewrite = rewriteOf plate
 --
 -- Usually 'transformOf' is more appropriate, but 'rewriteOf' can give better
 -- compositionality. Given two single transformations @f@ and @g@, you can
--- construct @\a -> f a `mplus` g a@ which performs both rewrites until a fixed point.
+-- construct @\a -> f a <|> g a@ which performs both rewrites until a fixed point.
```
ocharles: Another `\` to escape.

phadej:
Thanks

Issue #833, opened 2019-01-25
Control.Lens.Fold.Internal has incorrect documentation for `LeftMost`


emilypi:
The comment in question is [here](https://github.com/ekmett/lens/blob/60b89993903af95f138411a0721e952e84047a6c/src/Control/Lens/Internal/Fold.hs#L195). 

The most up-to-date definition for `preview` does not use `LeftMost`, it uses `First`. This may be stale documentation which is an easy fix. I haven't done a sweep for any other documentation that may need some love, but I'll update here if I find anything else before anyone responds.

emilypi:
Fixed in #836

PR #836, opened 2019-03-04
Add 'ilocally' + doc fixes

emilypi:
This adds an indexed variant for `locally`, and fixes some docs I noticed needed some love. Also, addresed #833

### DIFF
--- a/src/Control/Lens/Setter.hs
+++ b/src/Control/Lens/Setter.hs
@@ -63,7 +63,7 @@ module Control.Lens.Setter
   , passing, ipassing
   , censoring, icensoring
   -- * Reader Combinators
-  , locally
+  , locally, ilocally
   -- * Simplified State Setting
   , set'
   -- * Indexed Setters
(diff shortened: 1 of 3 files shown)


src/Control/Lens/Setter.hs
```diff
@@ -1149,6 +1149,24 @@ locally :: MonadReader s m => ASetter s s a b -> (a -> b) -> m r -> m r
...
+-- 'locally' l f ≡ 'ilocally' l f . const
+-- 'ilocally' l f ≡ 'locally' l f . 'Indexed'
+-- @
+--
+-- @
+-- ilocally :: MonadReader s m => 'IndexedLens' s s a b      -> (a -> b) -> m r -> m r
+-- ilocally :: MonadReader s m => 'IndexedTraversal' s s a b -> (a -> b) -> m r -> m r
+-- ilocally :: MonadReader s m => 'IndexedSetter' s s a b    -> (a -> b) -> m r -> m r
+-- @
+ilocally :: MonadReader s m => IndexedSetter i s s a b -> (i -> a -> b) -> m r -> m r
```
phadej: It should take `AnIndexedSetter`, see https://hackage.haskell.org/package/lens-4.17/docs/Control-Lens-Combinators.html#v:iover

emilypi: Thanks - late night over sight 😄

PR #842, opened 2019-04-18
Add gplate1

glguy:
Using the `Generic1` instance can be useful for types like:

```haskell
data T a = C a [T a]
```

where the normal Generic instance would not be able to find
the occurrence of `T a` under a list. `GPlated1` will use
`Traversable` instances to allow `plate` to find nested
children.

```haskell
>>> data T a = T [T a] deriving (Show, Generic, Generic1)
>>> toListOf gplate (T [T []])
[]
>>> toListOf gplate1 (T [T []])
[T []]
```

### DIFF
--- a/src/Control/Lens/Plated.hs
+++ b/src/Control/Lens/Plated.hs
@@ -6,6 +6,10 @@
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 
+#if __GLASGOW_HASKELL__ >= 706
+{-# LANGUAGE PolyKinds #-} -- gplate1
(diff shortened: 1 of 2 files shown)


src/Control/Lens/Plated.hs
```diff
@@ -96,6 +104,7 @@ module Control.Lens.Plated
 
   -- * Generics
   , gplate
+  , gplate1
   , GPlated
```
RyanGlScott: Since we export `GPlated`, we should probably export `GPlated1` as well, right?

glguy: You're right; fixed!

Issue #753, opened 2017-07-26
Add more instances for Swapped


treeowl:
For example, the `bifunctors` package has several types that should be instances:

```haskell
instance (Swapped f, Swapped g) => Swapped (Product f g) where
  swapped = iso (\(Pair m n) -> Pair (m ^. swapped) (n ^. swapped))
                (\(Pair m n) -> Pair (m ^. swapped) (n ^. swapped))

instance (Swapped p, Swapped q) => Swapped (Sum p q) where
  swapped = iso (\(L2 x) -> L2 (x ^. swapped)) (\(L2 x) -> L2 (x ^. swapped))

instance (Functor f, Swapped p) => (Swapped (Tannen f p)) where
  swapped = iso (\(Tannen x) -> Tannen (fmap (^. swapped) x))
                (\(Tannen x) -> Tannen (fmap (^. swapped) x))
```

There can also be instances for `(,,) a`, `(,,,) a b`, etc.

RyanGlScott:
Sounds reasonable to me. Pull requests welcome.

emilypi:
Looks like we covered what Bifunctors there were. Can anyone think about any else `Swapped`-able?

treeowl:
The class constraint means only `Bifunctor`s are possible. Otherwise there would be some (not-defined-in-lens-yet) reified `Iso` and `Equality`-related types to add.

jozanek:
@RyanGlScott I think this can be closed.

treeowl:
Yeah, I think this has been done for a while.

PR #843, opened 2019-04-25
Add additional Swapped instances

emilypi:
See: https://github.com/ekmett/lens/issues/753

@treeowl if you know of any other bifunctors with admissible `Swapped` instances, lets throw them on the pile!

### DIFF
--- a/src/Control/Lens/Iso.hs
+++ b/src/Control/Lens/Iso.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
 
 #if __GLASGOW_HASKELL__ >= 710
 {-# LANGUAGE PatternSynonyms #-}
@@ -20,7 +21,7 @@
 -- License     :  BSD-style (see the file LICENSE)
 -- Maintainer  :  Edward Kmett <ekmett@gmail.com>
 -- Stability   :  provisional
--- Portability :  Rank2Types
+-- Portability :  Rank2Types, TypeFamilies, FunctionalDependencies
 --
 ----------------------------------------------------------------------------
 module Control.Lens.Iso
@@ -99,16 +100,21 @@ import Control.Lens.Review
 import Control.Lens.Type
 import Control.Monad.State.Lazy as Lazy
 import Control.Monad.State.Strict as Strict
-import Control.Monad.Writer.Lazy as Lazy
-import Control.Monad.Writer.Strict as Strict
-import Control.Monad.RWS.Lazy as Lazy
-import Control.Monad.RWS.Strict as Strict
+import Control.Monad.Writer.Lazy as Lazy hiding (Product, Sum)
+import Control.Monad.Writer.Strict as Strict hiding (Product, Sum)
+import Control.Monad.RWS.Lazy as Lazy hiding (Product, Sum)
+import Control.Monad.RWS.Strict as Strict hiding (Product, Sum)
 import Control.Monad.ST.Lazy as Lazy
 import Control.Monad.ST as Strict
+
 import Data.Bifunctor
+import Data.Bifunctor.Biff
+import Data.Bifunctor.Flip
+import Data.Bifunctor.Product
+import Data.Bifunctor.Sum
+import Data.Bifunctor.Tannen
 import Data.ByteString as StrictB hiding (reverse)
 import Data.ByteString.Lazy as LazyB hiding (reverse)
-
 import Data.Functor.Identity
 import Data.Text as StrictT hiding (reverse)
 import Data.Text.Lazy as LazyT hiding (reverse)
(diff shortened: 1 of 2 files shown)


treeowl:
Perhaps add the sums and products from `GHC.Generics`?

emilypi:
Ah righto i did miss those

src/Control/Lens/Iso.hs
```diff
@@ -409,6 +412,47 @@ instance Swapped (,) where
...
+instance (Swapped p, Swapped q) => Swapped (Sum p q) where
+  swapped = iso f f
+    where
+      f (L2 x) = L2 (x ^. swapped)
+      f (R2 x) = R2 (x ^. swapped)
+
+instance (Functor f, Swapped p) => Swapped (Tannen f p) where
+  swapped = iso f f
+    where
+      f (Tannen x) = Tannen $ fmap (^. swapped) x
```
phadej: - How about `instance Swapped p => Swapped (Flip p)`?
- `(f ~ g, Functor f, Swapped p) => Swapped (Biff p f g)` is a bit on the edge

emilypi: Thanks @phadej! I was hoping to find a more satisfying `Biff` instance last night, but i suppose `f ~ g` is the best we get huh? 😄

phadej: Well, maybe `Coercible` bla blah (but that would be GHC >= 7.8), or some `Generic1` `Rep1` are isomorphic blah blah. I'm fine with just leaving it out too.

emilypi: I think we're good on instances now with the latest push. It turns out `:+:` and `:*:` are functors in `GHC.Generics`. I hadn't noticed before! very odd.

phadej: they are essentially `Product f g a` not `Product p q a b`

treeowl:
Oh, sorry, I misread which `Product` and `Sum` those were. You now have a redundant `GHC.Generics` import, I believe.

emilypi:
@treeowl no worries. I'm surprised hlint didn't catch that on mine. All fixed now, thanks.

phadej:
@emilypi looks good. If you mind squash when you are done polishing, we'll merge.

emilypi:
@phadej beat you to the squash 😄

treeowl:
I'm not very convinced by that `Biff` instance. Is it more likely to be used on purpose or by mistake,

emilypi:
I imagine you'd have to go out of your way to provide `f ~ g => Biff p f g` on purpose. It'd be terribly difficult to it by mistake. Do you have some place we can discuss off line?

treeowl:
I don't immediately see how to prove it's lawful. Can you demonstrate that?

phadej:
With `f ~ g`, `Biff p f g a b` is

```haskell
newtype Biff p f g a b = Biff (p (f a) (g b))
  =< f ~ g >

newype Biff p f f a b = Biff (p (f a) (f b))
```

looks quite symmetric to me. I'm sure you can prove the laws,
using `Swapped p` e.g. `swapped . swapped = id` is quite trivial already.

src/Control/Lens/Iso.hs
```diff
@@ -409,6 +415,57 @@ instance Swapped (,) where
...
+
+instance Swapped ((,,) x) where
+  swapped = iso f f
+    where
+      f ~(x,a,b) = (x,b,a)
```
glguy: Why are these tuple instances using an irrefutable match? The instance for 2-tuples doesn't nor do any of the other new instances for product-types. If there's a good reason can we document it on the instances so people know to rely on it. If there isn't can we switch it to a normal pattern?

emilypi: I'm a proponent of lazy pattern matches for composability reasons, but I'm not tied to them. I figured it's a Kmett library who enjoys going full boar on these things. Considering `swapped` is a good candidate for lazy behavior, it seemed like a good idea.

glguy: OK, then let's make them consistent and document them in the haddock comments

phadej:
Good point about `_1'`. Would it make sense to add strict `swapped'` (I'm not sure whether it can be external/not-class-member combinator).

emilypi:
@phadej I would assume that would be its own combinator in the same way that `_1'` is. I wouldn't expect my `swapped` behavior to be strict by default, which is exacly how this is also how it's done in `Tuple.hs`: https://github.com/ekmett/lens/blob/master/src/Control/Lens/Tuple.hs#L104

phadej:
I meant that i’m not sure that swapped' can be implemented using swapped, or if it have to be class member

emilypi:
Ah, no. I'm not sure if it can.

phadej:
Hmm, I think

```haskell
swapped' = iso f g where
  f x = x `seq` view swapped x
  g y = y `seq` review swapped y
{-# INLINE swapped' #-}
```

would be good enough.

treeowl:
Personally, I tend to be wary of hidden laziness, since it can cause space leaks and slow down code. The lazy instances also aren't strictly law-abiding, because swapping twice may give a more-defined value than doing nothing. But I recognize that Ed has a somewhat different perspective WRT class instances than I do. And the double-swap law *could* be changed to an inequality.

phadej:
@treeowl can lazy `swapped` be implemented in terms of strict `swapped'`?

treeowl:
@phadej, it can with a `Generic` constraint and something confining it to be a record. Arguably, strict swap should only be expressible from lazy swap with an additional `Seqable` (or whatever) constraint; the fact that `seq` is fully polymorphic in Haskell is a bit ugly. In the case of the `Arrow` instance for functions and (therefore?) the `Bifunctor` instance for pairs, `first` and `second` are lazy to support `ArrowLoop`. Is there a major use-case for laziness here?

ekmett:
Merging the PR, any remaining wibbles can be fixed up in-repo.

phadej:
@ekmett your opinion, whether `swapped` should use strict or irrefutable pattern matches?

PR #847, opened 2019-04-28
Document the inverse of isn't

alexanderkjeldaas:


ekmett:
Technically, the negation of a statement isn't its inverse.

The logician in me wants any referene to the "inverse of P -> Q" to be "~P->~Q"

ekmett:
Another way to write it that doesn't get hung up on language is to just state a law:

```haskell
isn't = not . is
```

Issue #844, opened 2019-04-26
Split Language.Haskell.TH.Lens into own package


phadej:
`template-haskell` may (and will with GHC-8.8.1) introduce breaking changes, which isn't possible to shim across `template-haskell`s bundled with GHC that `lens` support.

I suggest splitting that package out.
I think that `lens` itself does use something from there for `makeLenses` etc. machinery, but those bits doesn't need to be in the public interface.

RyanGlScott:
I tentatively support this idea, if only for the very practical reason that any breaking API change to `template-haskell` essentially requires `lens` to propagate those same changes to `Language.Haskell.TH.Lens`, which itself requires a major version bump to `lens`. A side effect of this is that we have to crank out major versions of `lens` for every new version of GHC that is released, which feels rather silly.

RyanGlScott:
Assuming that we do wish to split off `Language.Haskell.TH.Lens`, what name should this hypothetical package have? `template-haskell-lens`? `lens-th`? Something else?

phadej:
I like both lens-th and template-haskell-th; lens-th sounds like package to derive lens with th though. lens-template-haskell feels wrong.

So: template-haskell-lens

glguy:
Not that it makes it easier to pick but we also have `lens-aeson` out there

phadej:
- https://hackage.haskell.org/package/zlib-lens
- https://hackage.haskell.org/package/writer-cps-lens
- https://hackage.haskell.org/package/xml-html-conduit-lens
- https://hackage.haskell.org/package/hreader-lens
- https://hackage.haskell.org/package/time-lens

but

- https://hackage.haskell.org/package/lens-toml-parser
- https://hackage.haskell.org/package/lens-datetime
- https://hackage.haskell.org/package/lens-xml

---

I personally favour `$packagename-lens` naming; yet for `servant` we use `servant-$pkgname`. In anyway, consistency in `lens`-land is far sailed ship :)

fumieval:
FWIW, lens-aeson is named so because there was already aeson-lens which is unlawful

Taneb:
I'm strongly in favour of `template-haskell-lens`

glguy:
Should we orphan the `Plated` instances into this new package?

glguy:
WIP https://github.com/glguy/template-haskell-lens

Taneb:
[...]

PR #851, opened 2019-05-07
Extract template haskell

glguy:
See #844 for current discussion. Don't merge this until we're ready with a release of `template-haskell-lens`


RyanGlScott:
A couple of remarks:

* This branch appears to have an extra `Add gplate1` commit, except that it's different from the one that's currently in `master`. Can this branch be rebased on top of `master` to avoid this? Otherwise, it makes looking at the diff noisier than it should be.
* This goes a bit beyond what I originally imagined and not only removes `Language.Haskell.TH.Lens`, but also `Control.Lens.TH`. Personally, I only desire to package up `Language.Haskell.TH.Lens` separately simply because its API changes so frequently to keep in sync with `template-haskell` changes. On the other hand, `Control.Lens.TH` is quite stable, so I don't think the same argument applies there. What do you think?

phadej:
I also had only splitting out the `Language.Haskell.TH.Lens`.

`Control.Lens.TH` does use `Language.Haskell.TH.Lens`, but only a small portion (at least not many top level identifiers)

```
    • Variable not in scope:
        substType :: Map k0 Type -> [Type] -> [Type]
    • Perhaps you want to add ‘substType’ to the import list
      in the import of ‘Language.Haskell.TH.Lens’
      (src/Control/Lens/TH.hs:119:1-34).
    |
524 |         substType (Map.fromList $ zip vars tys) args
    |         ^^^^^^^^^

src/Control/Lens/TH.hs:535:29: error:
    • Variable not in scope: typeVars :: Getting (Set a) [Type] a
    • Perhaps you want to add ‘typeVars’ to the import list
      in the import of ‘Language.Haskell.TH.Lens’
      (src/Control/Lens/TH.hs:119:1-34).
    |
535 |       vars = toList $ setOf typeVars args
    |                             ^^^^^^^^

src/Control/Lens/TH.hs:566:40: error:
    • Variable not in scope:
        conFields
          :: (s0 -> Const (Endo [Type]) t0) -> Con -> Const (Endo [Type]) Con
    • Perhaps you want to add ‘conFields’ to the import list
      in the import of ‘Language.Haskell.TH.Lens’
      (src/Control/Lens/TH.hs:119:1-34).
    |
566 |                 , [field] <- toListOf (conFields._2) con
    |                                        ^^^^^^^^^

src/Control/Lens/TH.hs:577:29: error:
    • Variable not in scope: name :: Getting b TyVarBndr b
    • Perhaps you meant data constructor ‘Name’ (imported from Language.Haskell.TH.Syntax)
      Perhaps you want to add ‘name’ to the import list in the import of
      ‘Language.Haskell.TH.Lens’ (src/Control/Lens/TH.hs:119:1-34).
    |
577 |    let typeArgs = map (view name) (dataParameters dataDecl)
    |                             ^^^^

src/Control/Lens/TH.hs:581:14: error:
    • Variable not in scope:
        substTypeVars :: Map Name Name -> [b0] -> [Name]
    • Perhaps you want to add ‘substTypeVars’ to the import list
      in the import of ‘Language.Haskell.TH.Lens’
      (src/Control/Lens/TH.hs:119:1-34).
    |
581 |      return (substTypeVars m typeArgs)
    |              ^^^^^^^^^^^^^

src/Control/Lens/TH.hs:605:22: error:
    • Variable not in scope: name :: Getting t Con t
    • Perhaps you meant data constructor ‘Name’ (imported from Language.Haskell.TH.Syntax)
      Perhaps you want to add ‘name’ to the import list in the import of
      ‘Language.Haskell.TH.Lens’ (src/Control/Lens/TH.hs:119:1-34).
    |
605 |   let conName = view name con
    |                      ^^^^

src/Control/Lens/TH.hs:606:27: error:
    • Variable not in scope:
        typeVars :: Getting (Endo [a]) [TyVarBndr] a
    • Perhaps you want to add ‘typeVars’ to the import list
      in the import of ‘Language.Haskell.TH.Lens’
      (src/Control/Lens/TH.hs:119:1-34).
    |
606 |   let typeArgs = toListOf typeVars (dataParameters dataDecl)
```

glguy:
> A couple of remarks:
> 
> * This branch appears to have an extra `Add gplate1` commit, except that it's different from the one that's currently in `master`. Can this branch be rebased on top of `master` to avoid this? Otherwise, it makes looking at the diff noisier than it should be.

That was a mistake.

> * This goes a bit beyond what I originally imagined and not only removes `Language.Haskell.TH.Lens`, but also `Control.Lens.TH`. Personally, I only desire to package up `Language.Haskell.TH.Lens` separately simply because its API changes so frequently to keep in sync with `template-haskell` changes. On the other hand, `Control.Lens.TH` is quite stable, so I don't think the same argument applies there. What do you think?

But if everyone wants to keep half the TH stuff around and only drop the `Language.Haskell.TH.Lens` module into its own project I'm fine with that. I just misunderstood the goal. I think the fact that `Control.Lens.TH` is so easily separable might be a good argument for its removal. It would allow us to drop at least the `th-abstraction` dependency for people who don't need it. It would also potentially allow the `Control.Lens.TH` module to be cleaned up to use actual Template Haskell in its implementation. The current `Name` construction is an unfortunate hack and some of the other implementation would be nicer, I think, if it was able to actually use splices.

I'm in favor of making `lens` as small as it can be. For a lot of uses the TH modules aren't needed. We had luck shaving off dependencies like `aeson` and this just seemed like a chance to go further.

phadej:
`th-abstraction` is a dependency of e.g. `bifunctors`, so dropping it is pointless.

phadej:
@glguy if you'd like to split the `Control.Lens.TH` out too, then it would go into `lens-th` package :) 

I'm not against that either.

However, `template-haskell-lens` is currently a priority for GHC-8.8

PR #852, opened 2019-05-11
Split off Language.Haskell.TH.Lens

RyanGlScott:
This rips out the `Language.Haskell.TH.Lens` module in preparation for it being placed in a package of its own, `template-haskell-lens`. See #844 for more discussion on this.

Since various other parts of the `lens` library implemented functionality on top of `Language.Haskell.TH.Lens`, the most challenging part of this patch was figuring out how to replace those bits. Many parts of `Language.Haskell.TH.Lens` were easily interchangeable with equivalent functions from `th-abstraction`; I make heavy use of the `freeVariables`, `freeVariablesWellScoped`, and `quantifyType` functions, for instance (my use of the last function warranted bumping the lower version bounds on `th-abstraction` to 0.2.10). Other parts I couldn't figure out how to implement elegantly without simply copy-pasting the relevant parts of `Language.Haskell.TH.Lens` (e.g., the prisms for type family declarations). Perhaps there's a cleaner way to do this, but it's not obvious to me.

Fixes the `lens` side of #844.


RyanGlScott:
See lens/template-haskell-lens#1 for the corresponding `template-haskell-lens` PR.

PR #856, opened 2019-05-23
Add some API disclaimers to Language.Haskell.TH.Lens

RyanGlScott:
In lieu of #844, we should attempt to advertise the fact that the API offered by `Language.Haskell.TH.Lens` is not stable among different versions of `template-haskell`.

### DIFF
--- a/src/Language/Haskell/TH/Lens.hs
+++ b/src/Language/Haskell/TH/Lens.hs
@@ -20,7 +20,13 @@
 -- Stability   :  experimental
 -- Portability :  TemplateHaskell
 --
--- Lenses, Prisms, and Traversals for working with Template Haskell
+-- Lenses, Prisms, and Traversals for working with Template Haskell.
+--
+-- Beware that the API offered in this module is subject to change, as it
+-- mirrors the API exposed by the @template-haskell@ package, which
+-- frequently changes between different releases of GHC. An effort is made
+-- to identify the functions in this module which have different type
+-- signatures when compiled with different versions of @template-haskell@.
 ----------------------------------------------------------------------------
 module Language.Haskell.TH.Lens
   (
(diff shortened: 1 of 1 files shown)


src/Language/Haskell/TH/Lens.hs
```diff
@@ -1152,23 +1205,40 @@ _DataInstD
...
+      reviewer (x, y, z, w, u) = DataInstD x y z w u
+      remitter (DataInstD x y z w u) = Just (x, y, z, w, u)
+      remitter _ = Nothing
+#endif
 
-_NewtypeInstD :: DataPrism' [Type] Con
+-- |
+-- @
+-- _NewtypeInstD :: 'Prism'' 'Dec' ('Cxt', 'Maybe' ['TyVarBndr'], 'Type', 'Maybe' 'Kind', 'Con', ['DerivClause']) -- template-haskell-2.15+
+-- _NewtypeInstD :: 'Prism'' 'Dec' ('Cxt', 'Name', ['Type'],              'Maybe' 'Kind', 'Con', ['DerivClause']) -- template-haskell-2.12 through 2.14
```
phadej: I think this formatting goes wrong, as quotes are zero-space when rendered.

RyanGlScott: Ugh. What is the idiomatic way to align things in Haddock, then?

RyanGlScott: Or, more to the point, do you think it's even worth trying to align things at all?

phadej:
I just counted the quotes, there are “examples” in lens too (e.g. Folds needing Monoid and getters not)

src/Language/Haskell/TH/Lens.hs
RyanGlScott: After lots of lots of counting, I believe everything should be lined up properly now.

PR #862, opened 2019-05-29
Spread Iso ideas around

### LINKED ISSUE #860: substEq and simply should be levity polymorphic
`substEq` and `simply` should be levity polymorphic. This goes for anything else that may be lying around in continuation passing style.

### LINKED ISSUE #861: Add lensy basics for Equality
I think we should probably pull some things from `Control.Lens.Iso` into `Control.Lens.Equality`.

```haskell
cloneEquality :: AnEquality s t a b -> Equality s t a b
cloneEquality an = substEq an simple

-- | A version of 'simple' that takes explicit equality evidence
equality :: (s :~: a) -> (b :~: t) -> Equality s t a b
equality Refl Refl = simple

equality' :: (a :~: b) -> Equality' a b
equality' Refl = simple

-- | Recover a "profunctor lens" form of equality
underEquality :: AnEquality s t a b -> p t s -> p b a
underEquality an = substEq an id

-- | A version of 'substEq' that provides explicit equality evidence
withEquality :: forall s t a b (r :: TYPE rep).
   AnIso s t a b -> ((s :~: a) -> (b :~: t) -> r) -> r
withEquality an = substEq an (\f -> f Refl Refl)
```

treeowl:
* Add several functions to `Control.Lens.Equality` mimicking
  similar ones in `Control.Lens.Iso`.

* Add `withLens` mimicking `withIso`.

* Make several CPS-style functions levity-polymorphic, because we
  can.

Closes #861
Closes #860

### DIFF
--- a/src/Control/Lens/Equality.hs
+++ b/src/Control/Lens/Equality.hs
@@ -2,9 +2,14 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 #if __GLASGOW_HASKELL__ >= 706
 {-# LANGUAGE PolyKinds #-}
 #endif
+#if __GLASGOW_HASKELL__ >= 800
+{-# LANGUAGE TypeInType #-}
+#endif
+
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Control.Lens.Equality
@@ -20,6 +25,7 @@ module Control.Lens.Equality
   -- * Type Equality
     Equality, Equality'
   , AnEquality, AnEquality'
+  , (:~:)(..)
   , runEq
   , substEq
   , mapEq
@@ -27,12 +33,26 @@ module Control.Lens.Equality
   , simply
   -- * The Trivial Equality
   , simple
+  -- * 'Iso'-like functions
+  , equality
+  , equality'
+  , withEquality
+  , underEquality
+  , overEquality
+  , fromLeibniz
+  , fromLeibniz'
+  , cloneEquality
   -- * Implementation Details
   , Identical(..)
   ) where
 
 import Control.Lens.Type
 import Data.Proxy (Proxy)
+import Control.Lens.Internal.Equality ((:~:)(..))
+#if __GLASGOW_HASKELL__ >= 800
+import GHC.Exts (TYPE)
+import Data.Kind (Type)
+#endif
 
 #ifdef HLINT
 {-# ANN module "HLint: ignore Use id" #-}
@@ -68,14 +88,21 @@ runEq l = case l Identical of Identical -> Identical
 {-# INLINE runEq #-}
 
 -- | Substituting types with 'Equality'.
+#if __GLASGOW_HASKELL__ >= 800
+substEq :: forall s t a b rep (r :: TYPE rep).
+           AnEquality s t a b -> ((s ~ a, t ~ b) => r) -> r
+#else
 substEq :: AnEquality s t a b -> ((s ~ a, t ~ b) => r) -> r
+#endif
 substEq l = case runEq l of
   Identical -> \r -> r
 {-# INLINE substEq #-}
 
 -- | We can use 'Equality' to do substitution into anything.
-#if __GLASGOW_HASKELL__ >= 706
-mapEq :: forall KVS(k1 k2) (s :: k1) (t :: k2) (a :: k1) (b :: k2) (f :: k1 -> *) . AnEquality s t a b -> f s -> f a
+#if __GLASGOW_HASKELL__ >= 800
+mapEq :: forall k1 k2 (s :: k1) (t :: k2) (a :: k1) (b :: k2) (f :: k1 -> Type) . AnEquality s t a b -> f s -> f a
+#elif __GLASGOW_HASKELL__ >= 706
+mapEq :: forall (s :: k1) (t :: k2) (a :: k1) (b :: k2) (f :: k1 -> *) . AnEquality s t a b -> f s -> f a
 #else
 mapEq :: AnEquality s t a b -> f s -> f a
 #endif
@@ -89,7 +116,12 @@ fromEq l = substEq l id
 
 -- | This is an adverb that can be used to modify many other 'Lens' combinators to make them require
 -- simple lenses, simple traversals, simple prisms or simple isos as input.
+#if __GLASGOW_HASKELL__ >= 800
+simply :: forall p f s a rep (r :: TYPE rep).
+  (Optic' p f s a -> r) -> Optic' p f s a -> r
+#else
 simply :: (Optic' p f s a -> r) -> Optic' p f s a -> r
+#endif
 simply = id
 {-# INLINE simply #-}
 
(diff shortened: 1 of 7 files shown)


src/Control/Lens/Equality.hs
```diff
@@ -100,3 +120,42 @@ simply = id
...
+equality' :: (a :~: b) -> Equality' a b
+equality' Refl = id
+
+-- | Recover a "profunctor lens" form of equality
+underEquality :: AnEquality s t a b -> p t s -> p b a
+underEquality an = substEq an id
+{-# INLINE underEquality #-}
+
+-- | Convert a "profunctor lens" form of equality to an equality
+unLeibniz :: (forall p. p a b -> p s t) -> Equality s t a b
```
treeowl: Should probably reorder type arguments to match `underEquality`....

treeowl:
What should I do about the fact that ancient `base` versions don't have `Data.Type.Equality`?

RyanGlScott:
> What should I do about the fact that ancient `base` versions don't have `Data.Type.Equality`?

How about using the [`Identical`](https://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Combinators.html#t:Identical) data type instead?

treeowl:
> > What should I do about the fact that ancient `base` versions don't have `Data.Type.Equality`?
> 
> How about using the [`Identical`](https://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Combinators.html#t:Identical) data type instead?

I don't think that's likely to be what someone happens to have on hand or happens to need.

RyanGlScott:
> I don't think that's likely to be what someone happens to have on hand or happens to need.

This doesn't seem like that big of a deal in practice to me. It's not terribly difficult to convert from an `Identical` to `(:~:)` (or some other equality-carrying data type)—just pattern match on `Identical` and GHC's constraint solver will take care of the rest.

The only other alternative that I can envision is to use CPP to define `withEquality` only on those versions of GHC that provide `Data.Type.Equality`, but I'd like to avoid this if possible.

treeowl:
I don't think the CPP would be too terribly evil. Code that actually uses those functions will *need* to have the relevant parts of `D.T.E` in scope, and we don't reexport them, so I don't think this introduces any real version bounds issues.

RyanGlScott:
Fair enough. In the end, both the `(:~:)` and `Identical` versions are small deltas on `substEq`, so the choice of data type isn't of particular importance here. I'm suggesting `Identical` only because I think it will make maintaining the associated code easier in the future, but at the end of the day, this is your contribution, not mine. Choose whichever you feel is right.

treeowl:
I think reasonable people can disagree on this. Do you have any better names for `unLeibniz` and `unLeibniz'`? I really don't like those, but I couldn't think of anything better.

phadej:
`fromLeibniz` ?

treeowl:
> `fromLeibniz` ?

Yes, that's definitely less bad. Is there anything even better?

treeowl:
Hmmm..... Should `fromLeibniz` reverse `underEquality`, or would it actually be better to reverse `overEquality`?

treeowl:
I think this is ready to go. Should I squash the commits?

treeowl:
Note: the build failure is not my fault. It timed out for some reason and should probably just be restarted.

ekmett:
What is the oldest `base` that has `Data.Type.Equality`?

ekmett:
Looks like that is base 4.7? So the cost of using `Data.Type.Equality` is that we either don't export some combinators on older GHCs or we provide our own polyfilled equality type for those versions or get it from some shim library that maybe Ryan already maintains that provides it?

@RyanGlScott Is there a reason `base-compat` doesn't have a `Data.Type.Equality.Compat`?

ekmett:
Ah, @treeowl just pointed out to me that `base-compat` doesn't actually supply any of the missing types.

I'm cool with either version of it, where we either don't export instances on old compilers or where we polyfill the functionality from either an external or internal type. It wouldn't be the first time we rolled our own replacement for missing types, and just re-exported them when they weren't available on the target platform.

I'm somewhat more in favor of just in some `Control.Lens.Internal.Equality` module defining `(:~:)(Refl)` if its missing and export and have it re-export the one from `Data.Type.Equality` if we're on a new enough GHC. That fits with our surrounding behavior for things like `Control.Lens.Internal.Coerce`.

treeowl:
Oh, for the `with` stuff. I'm not a big fan of re-exports. I think it's okay to just not offer that for old versions. Who cares? It can't be used anyway.

treeowl:
I think this should be good to go.

phadej:
@ekmett @RyanGlScott, I have written a shim `Data.Type.Equality` not so long ago, in https://github.com/phadej/some/blob/master/src-old/Data/Type/Equality.hs

I can put it into some `type-equality-compat` package. I kind of needed it elsewhere, but there I needed a polykinded version, so simply requiring GHC-7.8, and not GHC-7.6 was easier.

Yet, here, a monokinded `:~:`, would work, won't it?

RyanGlScott:
I don't know if we need a full-blown compatibility package for `(:~:)` quite yet, but I would like to see the shimmed version in `Control.Lens.Equality` moved to its own module (perhaps `Control.Lens.Internal.Equality`). The reason is that `lens` _already_ defines another shim elsewhere in `Data.Data.Lens` ([here](https://github.com/ekmett/lens/blob/60b89993903af95f138411a0721e952e84047a6c/src/Data/Data/Lens.hs#L269-L270)), so now we have two of these things floating around. It would nice to consolidate them in a central location.

RyanGlScott:
If we wanted to later, we (for some value of "we") could create a compatibility package for `Data.Type.Equality` and replace `Control.Lens.Internal.Equality` with that. `lens` isn't even the only package that would benefit from this:

* `bound` has [a test suite that can't be run on older versions of GHC](https://github.com/ekmett/bound/blob/727e7b5ddf9ef1afc20e89dee5a351ed496a0320/bound.cabal#L141-L142) due to requiring `Data.Type.Equality`.
* `constraints` defines its [own `(:~~:)`](https://github.com/ekmett/constraints/blob/75e3ac1145ce5d455a6b992545e1ee409dc0d429/src/Data/Constraint/Deferrable.hs#L89-L97).
* `machines` defined its [own `(:~:)`](https://github.com/ekmett/machines/blob/ec1c64187031cb1b770b453bd447e2d0812d4df0/src/Data/Machine/Is.hs#L21-L23) (although swapping it out for `Data.Type.Equality` isn't entirely straightforward due to https://github.com/ekmett/machines/issues/88#issuecomment-318774661).

Obviously, that is more work than just defining another module in `lens`, so I don't think we should block this PR on that. But it does provide an interesting direction to consider for the future.

treeowl:
@RyanGlScott Do you want me to expose `equality`, `equality'`, and `withEquality` when compiling with `base` versions that will require users to import the internal shim to use them? Or should I leave those conditional?

RyanGlScott:
If we're going to bother creating a shim in the first place, I think it would be reasonable to expose these functions on all versions of `base`. We could make things slightly easier on the programmers' end by re-exporting `(:~:)(..)` so they won't have to import the `.Internal` module themselves.

treeowl:
I generally dislike re-exports, but I guess it's okay this time...

src/Control/Lens/Internal/Equality.hs
```diff
@@ -0,0 +1,83 @@
...
+inner Refl = Refl
+
+-- | Extract equality of type constructors from an equality of applied types
+outer :: (f a :~: g b) -> (f :~: g)
+outer Refl = Refl
+
+deriving instance Eq   (a :~: b)
+deriving instance Ord  (a :~: b)
+deriving instance Show (a :~: b)
+deriving instance a ~ b => Bounded (a :~: b)
```
RyanGlScott: Due to an old GHC bug, you won't be able to derive this instance on GHC 7.4 or 7.6. As a workaround, I'd just define manually:

```hs
instance a ~ b => Bounded (a :~: b) where
  minBound = Refl
  maxBound = Refl
```

src/Control/Lens/Internal/Equality.hs
```diff
@@ -0,0 +1,83 @@
...
+-- | Type-safe cast, using propositional equality
+castWith :: (a :~: b) -> a -> b
+castWith Refl x = x
+
+-- | Generalized form of type-safe cast using propositional equality
+gcastWith :: (a :~: b) -> ((a ~ b) => r) -> r
+gcastWith Refl x = x
+
+-- | Apply one equality to another, respectively
+apply :: (f :~: g) -> (a :~: b) -> (f a :~: g b)
```
RyanGlScott: This definition won't kind-check unless `(:~:)` is poly-kinded, and `PolyKinds` is known to be buggy on GHC 7.4. For the sake of simplicity, I'd vote to just leave out `apply`, since we're not using it in `lens`.

src/Control/Lens/Internal/Equality.hs
```diff
@@ -0,0 +1,83 @@
...
+apply Refl Refl = Refl
+
+-- | Extract equality of the arguments from an equality of applied types
+inner :: (f a :~: g b) -> (a :~: b)
```
RyanGlScott: Interestingly, GHC 7.4 and 7.6 appear to be unable to typecheck `inner` (again, likely due to old GHC bugs). Let's just leave out `inner`.

src/Control/Lens/Internal/Equality.hs
```diff
@@ -0,0 +1,83 @@
...
+inner Refl = Refl
+
+-- | Extract equality of type constructors from an equality of applied types
+outer :: (f a :~: g b) -> (f :~: g)
```
RyanGlScott: GHC 7.4 and 7.6 can't typecheck `outer`, much like they can't typecheck `inner`. Let's drop it too.

src/Control/Lens/Internal/Equality.hs
ekmett: In theory, you should be able to convert the equality to a Liebnizian equality and use the approach I use in the old `eq` package in order to get `inner` to work, as that seems to work from 7.0 on, if we want it.

src/Control/Lens/Internal/Equality.hs
ekmett: Virtually any attempt to use `PolyKinds` in 7.4 is broken, yielding object files that you can't safely link against without segfaults, so this one could at best be salvaged to work with 7.6 using something like the approach in `eq`.

RyanGlScott:
What is the current status of this, @treeowl?

treeowl:
@RyanGlScott I just pushed an update I *hope* will resolve the remaining issues.

RyanGlScott:
Thanks!

PR #869, opened 2019-06-12
Disambiguate TH-generated classy prisms for punned data type names

### LINKED ISSUE #865: `makeClassyPrisms` double definitions
Something is rotten in Denmark.

```
newtype SDLException = SDLException String deriving Show
makeClassyPrisms ''SDLException
```

```
src/Engine/SDL.hs:20:1: error:
    Multiple declarations of ‘_SDLException’
    Declared at: src/Engine/SDL.hs:20:1
                 src/Engine/SDL.hs:20:1
   |
```

I don't use classy prisms very often, but this seems broken, as this is almost a textbook usecase.

RyanGlScott:
Consider a scenario in which one uses Template Haskell to generate classy prisms for a data type that shares a name with one of its constructors, as in the following example from #865:

```hs
newtype SDLException = SDLException String deriving Show
makeClassyPrisms ''SDLException
```

We clearly can't use the name `_SDLException` to refer to both the prism for the data type and the prism for the constructor, so we need to disambiguate them somehow. This patch adopts the convention that in such a scenario, the name of the prism for the data type will be prefixed with an extra `_` (i.e., `__SDLException`) to distinguish it from the constructor prism, `_SDLException`.

Fixes #865.

### DIFF
--- a/src/Control/Lens/Internal/PrismTH.hs
+++ b/src/Control/Lens/Internal/PrismTH.hs
@@ -105,6 +105,31 @@ makePrisms = makePrisms' True
 -- Generate an "As" class of prisms. Names are selected by prefixing the constructor
 -- name with an underscore.  Constructors with multiple fields will
 -- construct Prisms to tuples of those fields.
+--
+-- In the event that the name of a data type is also the name of one of its
+-- constructors, the name of the 'Prism' generated for the data type will be
+-- prefixed with an extra @_@ (if the data type name is prefix) or @.@ (if the
+-- name is infix) to disambiguate it from the 'Prism' for the corresponding
+-- constructor. For example, this code:
+--
+-- @
+-- data Quux = Quux Int | Fred Bool
+-- makeClassyPrisms ''Quux
+-- @
+--
+-- will create:
+--
+-- @
+-- class AsQuux s where
+--   __Quux :: Prism' s Quux -- Data type prism
+--   _Quux :: Prism' s Int   -- Constructor prism
+--   _Fred :: Prism' s Bool
+--
+--   _Quux = __Quux . _Quux
+--   _Fred = __Quux . _Fred
+--
+-- instance AsQuux Quux
+-- @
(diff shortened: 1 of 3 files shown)


src/Control/Lens/Internal/PrismTH.hs
```diff
@@ -105,6 +105,31 @@ makePrisms = makePrisms' True
...
 -- construct Prisms to tuples of those fields.
+--
+-- In the event that the name of a data type is also the name of one of its
+-- constructors, the name of the 'Prism' generated for the data type will be
+-- prefixed with an extra @_@ (if the data type name is prefix) or @.@ (if the
+-- name is infix) to disambiguate it from the 'Prism' for the corresponding
+-- constructor. For example, this code:
+--
+-- @
+-- data Quux = Quux Int | Quuz Bool
```
phadej: These are dyslexia unfriendly, I read that as `Quux Int | Quux Bool`; maybe `Quux Int | Fuzzy Bool` ?

RyanGlScott: I somewhat deliberately chose the metasyntactic variable name `Quuz` so that: it would have the same length as `Quux`, ensuring that the spacing of their type signatures would align nicely. I'd certainly be willing to switch it out with a different variable name of the same length. Some potential candidates are `Fred` and `Thud` (taken from [here](https://tools.ietf.org/html/rfc3092)).

RyanGlScott: I went ahead and swapped out `Quuz` for `Fred`.

phadej: Thanks!

RyanGlScott:
```
08:53:01 <RyanGlScott> edwardk: Is this the design you had in mind for `makeClassPrisms` + punned data type names? https://github.com/ekmett/lens/pull/869
08:53:24 <edwardk> yep
```

PR #871, opened 2019-06-16
Add more data type prisms to Control.Exception.Lens

RyanGlScott:
This is a follow-up to #869 that implements the proposed refactoring to the classy prisms in `Control.Exception.Lens` stated in https://github.com/ekmett/lens/issues/865#issuecomment-501368037. For each class dedicated to a data type that shares a name with one of its constructors (e.g., `AsNonTermination`), we add a data type prism prefixed with `__`, following the convention established in 867d41f76043b6909d13959d415d24bc2af64952. We also define default implementations for the existing prisms in terms of the new data type prism.

In the strictest sense, this is a backwards-compatible change, as existing code that defines instances of `AsNonTermination` _et al._ will continue to compile. However, this code will now produce warnings stating that `__AsNonTermination` is not defined, so this will likely inspire people to update their code anyway. Luckily, this change will debut in a new major version of `lens`, so it's easier to justify.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -12,6 +12,16 @@
   classy prism for the data type will be prefixed with an extra `_` (for
   prefix names) or `.` (for infix names) to disambiguate it from the prism
   for the constructor.
+* Several type classes in `Control.Exception.Lens` now have additional
+  prisms corresponding to the data type that they focus on, in accordance
+  with the new convention established in the bullet point above. For example,
+  `AsNonTermination` now has an additional
+  `__NonTermination :: Prism' t NonTermination` method, and the existing
+  `_NonTermination :: Prism' t ()` method now has a default implementation in
+  terms of `__NonTermination`.
+
+  As a consequence of this change, you may need to update instances of these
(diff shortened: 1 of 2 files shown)


RyanGlScott:
I pre-emptively received an approval for this commit from @ekmett on the `#haskell-lens` IRC channel:

```
08:53:49 <RyanGlScott> Cool. What are your thoughts on this potential follow-up to that PR? https://github.com/ekmett/lens/issues/865#issuecomment-501368037
08:53:49 <RyanGlScott> This would actually involve changing the API of `lens` slightly, so I wanted an explicit nod of approval before considering it
09:18:09 <RyanGlScott> edwardk: I've implemented this idea in this branch ( https://github.com/ekmett/lens/tree/more-exception-prisms ), so it's ready to go if you're alright with it
09:19:18 <edwardk> lgtm
```

But I've posted it here as well for wider publicity.

ekmett:
As a general statement that you can point to in the future if anybody questions it, I trust your judgment implicitly where my libraries are concerned. If you feel the need to patch something or to change an edge case, change it.

PR #876, opened 2019-07-15
Generalize magnify

treeowl:
The constraints get pretty horrible when you can't assume, a priori, that `to` works. Hopefully, this makes sense.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -28,6 +28,9 @@
 * Add `filteredBy`.
 * Add `adjoin` to `Control.Lens.Unsound`.
 * Add `Each (Either a a) (Either b b) a b` instance.
+* Make `magnify` offer its getter argument the `Contravariant` and `Functor`
+  instances it will require. This allows `magnify` to be used without
+  knowing the concrete monad being magnified.
 
 4.17.1 [2019.04.26]
 -------------------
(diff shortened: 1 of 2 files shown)


treeowl:
@RyanGlScott any idea why this fails for older GHC versions?

treeowl:
It smells like maybe the type checker was looking for equality where it needed subsumption, or failed to expand a type synonym, or whatever. How should I work around it?

RyanGlScott:
Manual eta expansion usually works around old GHC bugs of this sort. That is:

```diff
@@ -251,7 +251,7 @@ instance Monad m => Magnify (ReaderT b m) (ReaderT a m) b a where
 -- 'magnify' = 'views'
 -- @
 instance Magnify ((->) b) ((->) a) b a where
-  magnify = views
+  magnify l = views l
   {-# INLINE magnify #-}

 instance (Monad m, Monoid w) => Magnify (Strict.RWST b w s m) (Strict.RWST a w s m) b a where
```

treeowl:
Oh, so that's not even a bug, per se, but a different sort of change.
Thanks for the tip!

treeowl:
@RyanGlScott do you think this change is a good idea? Also, for bleeding-edge GHC, should we instead use `QuantifiedConstraints` to make those superclass constraints?

RyanGlScott:
To be honest, this part of `lens` goes well beyond my understanding, so I don't think I'm qualified to make a well-informed judgment on its design.

treeowl:
Thanks!

PR #877, opened 2019-07-17
Prisms.TH: Generate INLINE pragmas for ordinary/iso prisms

Peaker:


RyanGlScott:
This seems reasonable to me.

I noticed that the final `makeConsPrisms` case (i.e., the `makeConsPrisms t cons (Just typeName)` case) still won't generate any `INLINE` pragmas. Is this intended?

Peaker:
I don't use those. Not sure if those are desired.

RyanGlScott:
I know that @ekmett uses `makeClassyPrisms` (which exercises the code path I mention in https://github.com/ekmett/lens/pull/877#issuecomment-512222759), so I'm sure he'd appreciate having the same `INLINE` treatment there that other TH-based prism generators receive. After all, the code that `makeClassyPrisms` generates is _very_ similar:

```
λ> data Foo = MkFoo1 | MkFoo2
λ> ; $(makeClassyPrisms ''Foo)
<interactive>:7:5-26: Splicing declarations
    makeClassyPrisms ''Foo
  ======>
    class AsFoo r_a8LQ where
      _Foo :: Prism' r_a8LQ Foo
      _MkFoo1 :: Prism' r_a8LQ ()
      _MkFoo2 :: Prism' r_a8LQ ()
      _MkFoo1 = ((.) _Foo) _MkFoo1
      _MkFoo2 = ((.) _Foo) _MkFoo2
    instance AsFoo Foo where
      _Foo = id
      _MkFoo1
        = (prism (\ () -> MkFoo1))
            (\ x_a8LR
               -> case x_a8LR of
                    MkFoo1 -> Right ()
                    _ -> Left x_a8LR)
      _MkFoo2
        = (prism (\ () -> MkFoo2))
            (\ x_a8LS
               -> case x_a8LS of
                    MkFoo2 -> Right ()
                    _ -> Left x_a8LS)
```

Granted, the implementation of `_Foo` and the default definitions of `_MkFoo1` and `_MkFoo2` are quite simple, so perhaps GHC can inline those without needing explicit guidance. But the implementations of `_MkFoo1` and `_MkFoo2` in the `AsFoo Foo` instance are more-or-less exactly the same as what `makePrisms` would generate, so if `makePrisms` generates `INLINE` pragmas, then it feels like `makeClassyPrisms` ought to generate `INLINE` pragmas in those spots as well.

PR #878, opened 2019-07-22
Add a Field1 instance for GHC.Tuple.Unit

treeowl:


src/Control/Lens/Tuple.hs
```diff
@@ -110,6 +111,9 @@ class Field1 s t a b | s -> a, t -> b, s b -> t, t a -> s where
 instance Field1 (Identity a) (Identity b) a b where
   _1 f (Identity a) = Identity <$> f a
 
+instance Field1 (Unit a) (Unit b) a b where
+  _1 f (Unit a) = Unit <$> f a
```
treeowl: Or should this be lazy???? I think *all* the tuple instances should be strict, but I know others disagree...

RyanGlScott: I'd vote to keep it the way you've written it, as it's consistent with the other `Field1` instances that are currently defined. If others feel strongly about changing this convention, then I think it would be best to leave that to a separate issue.

src/Control/Lens/Tuple.hs
```diff
@@ -63,6 +63,7 @@ import           Data.Profunctor       (dimap)
 import           Data.Proxy            (Proxy (Proxy))
 import           GHC.Generics          ((:*:) (..), Generic (..), K1 (..),
                                         M1 (..), U1 (..))
+import           GHC.Tuple             (Unit)
```
RyanGlScott: Based on the CI failures, I think this should be `(Unit(..))`.

src/Control/Lens/Tuple.hs
phadej: Other `Field1` insances are inconsistent. See `Pair f g a` and `(a, b)`.

RyanGlScott: Oops, I'm not sure what I was thinking when I wrote https://github.com/ekmett/lens/pull/878/files#r306307359, as the tuple instances are indeed implemented using irrefutable pattern matches. Hrm. I don't personally have a strong opinion on this, but for the sake of avoiding this PR from being bogged down by bikeshedding, it might be wise to stick to the current convention for now and raise a separate issue about how to reconcile the strictness of the instances for tuples and `Product`.

glguy: We have `_1'` for when you want to reduce leaks

glguy:
Most `Field1` instances (everything but `Pair` and (`:*:`)) are lazy. I suspect these two were added later and their being strict is a mistake. We have `_1'` for recovering the strict behavior.

RyanGlScott:
`Unit` is now named `Solo`, so it would be worth using the modern nomenclature in this code.

phadej:
We could depend on `OneTuple` to add `Solo` instances. We can still have instances for `Unit` separately where it exists.

PR #879, opened 2019-08-03
RFC: Add GHCJS Travis-CI job

phadej:


lens.cabal
```diff
@@ -406,7 +408,8 @@ test-suite hunit
   hs-source-dirs: tests
   default-language: Haskell2010
 
-  if !flag(test-hunit)
+  -- test-framework-th doesn't work due regex-posix
```
RyanGlScott: Doesn't work on GHCJS in particular, you mean? Can you elaborate more on this?

lens.cabal
```diff
@@ -428,6 +431,9 @@ test-suite doctests
   default-language:  Haskell2010
   x-doctest-options: --fast
 
+  if impl(ghcjs)
+    buildable: False
```
RyanGlScott: Why is this not buildable on GHCJS?

lens.cabal
```diff
@@ -387,7 +388,8 @@ test-suite properties
     lens-properties/src
   include-dirs: include
   default-language: Haskell2010
-  if !flag(test-properties)
+
+  if !flag(test-properties) || impl(ghcjs)
```
RyanGlScott: Why is this not buildable on GHCJS?

examples/lens-examples.cabal
```diff
@@ -49,7 +50,7 @@ library
   ghc-options: -Wall
 
 executable lens-pong
-  if !flag(pong)
+  if !flag(pong) || impl(ghcjs)
```
RyanGlScott: Why is this not buildable on GHCJS?

cabal.haskell-ci
```diff
@@ -1,6 +1,8 @@
+distribution:           bionic
```
RyanGlScott: Ooh, I didn't realize this was configurable now. Nice!

Out of curiosity, what is the default setting? Is there a reason to prefer something over the default setting?

.travis.yml
```diff
@@ -111,75 +134,77 @@ install:
...
   # cabal check...
-  - (cd lens-* && ${CABAL} -vnormal check)
-  - (cd lens-examples-* && ${CABAL} -vnormal check)
-  - (cd lens-properties-* && ${CABAL} -vnormal check)
+  - (cd ${PKGDIR_lens} && ${CABAL} -vnormal check)
+  - (cd ${PKGDIR_lens_examples} && ${CABAL} -vnormal check)
+  - (cd ${PKGDIR_lens_properties} && ${CABAL} -vnormal check)
   # haddock...
-  - ${CABAL} v2-haddock -w ${HC} ${TEST} ${BENCH} all | color_cabal_output
+  - if ! $GHCJS && [ $HCNUMVER -ge 70400 ] ; then ${CABAL} v2-haddock $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output ; fi
```
RyanGlScott: Isn't this `$HCNUMVER -ge 70400` check redundant, given that the minimum version of GHC that `lens` supports is 7.4?

.travis.yml
```diff
@@ -111,75 +134,77 @@ install:
...
+    echo "              26021a13b401500c8eb2761ca95c61f2d625bfef951b939a8124ed12ecf07329" >> $CABALHOME/config
+    echo "              f76d08be13e9a61a377a85e2fb63f4c5435d40f8feb3e12eb05905edb8cdea89" >> $CABALHOME/config
     echo "   key-threshold: 3"                                                            >> $CABALHOME/config
     fi
   - cat $CABALHOME/config
   - rm -fv cabal.project cabal.project.local cabal.project.freeze
   - travis_retry ${CABAL} v2-update -v
-  - if [ $HCNUMVER -eq 80801 ] ; then ${CABAL} v2-install -w ${HC} -j2 hlint --constraint='hlint ==2.1.*' | color_cabal_output ; fi
+  - if ! $GHCJS && [ $HCNUMVER -ge 80800 ] && [ $HCNUMVER -lt 80802 ] ; then ${CABAL} v2-install $WITHCOMPILER -j2 hlint --constraint='hlint ==2.1.*' | color_cabal_output ; fi
+  - if $GHCJS ; then ${CABAL} v2-install -w ghc-8.6.5 happy ; fi
```
RyanGlScott: I'm curious to know why this line is necessary. (Disclaimer: I'm not at all familiar with GHCJS's best practices, so I apologize in advance if this is a dumb question.)

.travis.yml
```diff
@@ -54,19 +56,36 @@ matrix:
...
   allow_failures:
     - compiler: ghc-head
     - compiler: ghc-8.8.1
 before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" = "linux" ]; then
+    sudo add-apt-repository -y ppa:hvr/ghc;
+    sudo add-apt-repository -y ppa:hvr/ghcjs;
+    sudo apt-get update;
+    sudo apt-get install $CC cabal-install-3.0 freeglut3-dev;
```
RyanGlScott: Why is this part necessary, given that `cabal-install-3.0` and `freeglut3-dev` are both listed in the `addons` part of the configuration matrix?

.travis.yml
```diff
@@ -4,10 +4,11 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.3.20190425
+# version: 0.3.20190804.2
 #
 language: c
-dist: xenial
+dist: bionic
+sudo: required
```
RyanGlScott: I'm very surprised that `sudo: required` is necessary now. Doesn't that prevent Travis from using [container-based builds](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/), which are purported to start faster than other builds?

RyanGlScott:
Some other questions:

* Do you know why the GHC 8.8.1 build [is failing](https://travis-ci.org/ekmett/lens/jobs/567527069) on this PR?
* Do you know what causes [the following oddity](https://travis-ci.org/ekmett/lens/jobs/567527060#L1598) in the GHCJS build?

   ```
   [1 of 2] Compiling T799             ( tests/T799.hs, /tmp/dist-test.tVHw/dist-newstyle/build/x86_64-linux/ghcjs-8.4.0.1/lens-4.18/build/templates/templates-tmp/T799.js_o )
   Linking Template Haskell ()
   Linking Template Haskell (ThRunner1)
   [2 of 2] Compiling Main             ( tests/templates.hs, /tmp/dist-test.tVHw/dist-newstyle/build/x86_64-linux/ghcjs-8.4.0.1/lens-4.18/build/templates/templates-tmp/Main.js_o )
   Linking Template Haskell (T799,ThRunner2)
   Linking Template Haskell (T799,ThRunner3)
   Linking Template Haskell (T799,ThRunner4)
   Linking Template Haskell (T799,ThRunner5)
   Linking Template Haskell (T799,ThRunner6)
   Linking Template Haskell (T799,ThRunner7)
   Linking Template Haskell (T799,ThRunner8)
   Linking Template Haskell (T799,ThRunner9)
   Linking Template Haskell (T799,ThRunner10)
   Linking Template Haskell (T799,ThRunner11)
   Linking Template Haskell (T799,ThRunner12)
   Linking Template Haskell (T799,ThRunner13)
   Linking Template Haskell (T799,ThRunner14)
   Linking Template Haskell (T799,ThRunner15)
   Linking Template Haskell (T799,ThRunner16)
   Linking Template Haskell (T799,ThRunner17)
   Linking Template Haskell (T799,ThRunner18)
   Linking Template Haskell (T799,ThRunner19)
   Linking Template Haskell (T799,ThRunner20)
   Linking Template Haskell (T799,ThRunner21)
   Linking Template Haskell (T799,ThRunner22)
   Linking Template Haskell (T799,ThRunner23)
   Linking Template Haskell (T799,ThRunner24)
   Linking Template Haskell (T799,ThRunner25)
   Linking Template Haskell (T799,ThRunner26)
   Linking Template Haskell (T799,ThRunner27)
   Linking Template Haskell (T799,ThRunner28)
   Linking Template Haskell (T799,ThRunner29)
   Linking Template Haskell (T799,ThRunner30)
   Linking Template Haskell (T799,ThRunner31)
   Linking Template Haskell (T799,ThRunner32)
   Linking Template Haskell (T799,ThRunner33)
   Linking /tmp/dist-test.tVHw/dist-newstyle/build/x86_64-linux/ghcjs-8.4.0.1/lens-4.18/build/templates/templates.jsexe (Main,T799)
   ```

lens.cabal
phadej: regex-posix uses C code, which doesn’t seem to be shimmed by GHCJS. In general regex-tdfa is better as pure-Haskell lib

lens.cabal
phadej: doctest depends on ghc, tricky with ghcjs

examples/lens-examples.cabal
phadej: Gloss: OpenGL

cabal.haskell-ci
phadej: Default is xenial, bionic isn’t officially supported. Installing hvr’s ppa GHCJS on xenial however goes bad as nodejs is old there

.travis.yml
phadej: I tried to make generation of these checks nicer, i’ll Check why this isn’t as pretty as it could be

.travis.yml
phadej: We need to have happy (and actually all tools) to be available. cabal crosscompiling story is subpar. Currently it does build happy with GHCJS but cannot it use (and imo would be silly to).

.travis.yml
phadej: With bionic apt pragma isn’t executed. No idea why.

.travis.yml
phadej: It’s required as I cannot add ghcjs repository otherwise. VM images are slower to start, but OTOH are beefier machines, maybe it evens out.

phadej:
Oddity: it’s how GHCJS does TH. I suspect there are 20 or so TH splices in that file.

phadej:
Good questions. I’ll recheck them when I’m back to proper internet

lens.cabal
phadej: test-framework-th -> regex-posix

.travis.yml
RyanGlScott: I'm still unclear on what exactly goes wrong if `happy` _isn't_ installed beforehand. Does a particular package fail to build?

lens.cabal
RyanGlScott: Interesting. Is there an issue about this somewhere where I can read more about this?

lens.cabal
RyanGlScott: Ah, I see. How does GHCJS normally handle packages with C code? Is there a repository of GHCJS-specific patches that people use? I'm aware of [`hackage-ghcjs-overlay`](https://github.com/nomeata/hackage-ghcjs-overlay), although that doesn't appear to have any patches at the moment, so surely that can't be the answer I'm seeking.

phadej: Something is built-in GHCJS itself: https://github.com/ghcjs/ghcjs/tree/ghc-8.6/lib/patches

cabal.haskell-ci
phadej: reverted back to `xenial` by manually installing nodejs.

.travis.yml
phadej: `haskell-src-exts` needs `happy`. And when we use GHCJS `cabal` fails to install&use proper `happy`. But things "just work" if there's `happy` in `.$PATH`.

I.e. there are bugs, but they luckily cancel each others out.

.travis.yml
phadej: Reverted to `xenial`. I'm not sure even where to report anything to travis nowadays. And especially about something which "accidentally" works, not publicly announced to be even an experimental feature.

lens.cabal
phadej: Not really. I don't think it's worth an issue. `doctest` doesn't work with any other compiler than GHC, neither with GHCi-less version (which GHCJS is).

phadej:
@RyanGlScott I think I addressed all of your last round comments.

lens.cabal
RyanGlScott: I see. If I'm looking at that right, all of those are boot libraries, correct? What happens if someone wishes to patch something that's not a boot library?

cabal.haskell-ci
RyanGlScott: Cool! Now that we're back on Xenial, doesn't this mean we can remove the `sudo: required` bits?

.travis.yml
RyanGlScott: I see! Has this bug been reported somewhere? I'm unclear if this is a bug with GHCJS, `cabal`, or both.

.travis.yml
```diff
@@ -111,75 +135,77 @@ install:
...
-  - ${CABAL} v2-build -w ${HC} ${TEST} ${BENCH} all | color_cabal_output
+  - ${CABAL} v2-build $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output
   # Testing...
-  - ${CABAL} v2-test -w ${HC} ${TEST} ${BENCH} all | color_cabal_output
+  - if ! $GHCJS ; then ${CABAL} v2-test $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output ; fi
   # HLint..
-  - if [ $HCNUMVER -eq 80801 ] ; then (cd lens-* && hlint src) ; fi
-  - if [ $HCNUMVER -eq 80801 ] ; then (cd lens-examples-* && hlint .) ; fi
-  - if [ $HCNUMVER -eq 80801 ] ; then (cd lens-properties-* && hlint src) ; fi
+  - if ! $GHCJS && [ $HCNUMVER -ge 80800 ] && [ $HCNUMVER -lt 80802 ] ; then (cd ${PKGDIR_lens} && hlint src) ; fi
```
RyanGlScott: Now I see why only the GHC 8.8 build is failing: it's because we're only invoking `hlint` on that particular configuration. I thought the convention was to only run `hlint` on the most recent, non-`HEADHACKAGE` configuration?

phadej: This is a bug...

.travis.yml
phadej: Well... It's a mess. `Cabal` and `cabal-install` should treat GHCJS as cross-compiler, and even some ground work is done we need a man-year to get there.

`ghcjs-8.4` reports

```
 ,("Build platform","x86_64-unknown-linux")
 ,("Host platform","x86_64-unknown-linux")
 ,("Target platform","x86_64-unknown-linux")
```

which confuses `Cabal`. I think it's fixed in GHCJS-8.6 (or so I was told, cannot check, I don't have bindists), so maybe things works better with that.

That said, I don't dare to ask HVR to try to package 8.6 (only for this reason).

lens.cabal
phadej: That's very good question. I think `hackage-ghcjs-overlay` existed for that kind of reason.

But now I'd just submit patches to the libraries. In this particular case, change `test-framework-th` to use `regex-tdfa`. Yet if one goes as far, maybe switching to `tasty` is an option too.

cabal.haskell-ci
phadej: No, we cannot. `sudo apt-get install ghcjs` still needs `sudo`.

RyanGlScott: I'm confused. https://github.com/ekmett/lens/pull/879#discussion_r310676432 gave me the impression that the only reason we needed this `sudo` business in the first place was due to the use of Bionic. Now that we're back to Xenial, shouldn't we be able to just put this stuff back into the `addons` stanza of the configuration matrix?

.travis.yml
RyanGlScott: OK. In that case, I'll keep this in mind when GHCJS-8.6 becomes widely available :)

lens.cabal
RyanGlScott: To be clear, I'm not super bothered by disabling the test suite for GHCJS in this particular situation. In the event that testing against GHCJS becomes more popular, however, it might be worthwhile to investigate this further. I'm content for now knowing that there's at least a glimmer of hope in `hackage-ghcjs-overlay`.

cabal.haskell-ci
phadej: No. `hvr-ghcjs` neither `nodejs` APT repositories aren't on the whitelist, and there is no way to put them there at the moment: https://github.com/travis-ci/apt-source-safelist

> At this moment, we are unable to continue processing the APT sources requests. If you require a new source, please use `sudo: required` in your `.travis.yml` file.

phadej: Also compare
- https://travis-ci.org/ekmett/lens/builds/567513619
- https://travis-ci.org/ekmett/lens/builds/567513619

The jobs seem to take about the same time.

One recent data point is https://github.com/haskell/cabal/pull/5538

> Our `sudo: true` infrastructure are full VMs and are generally faster build environments than the `false` counterpart.

.travis.yml
phadej: fixed

cabal.haskell-ci
RyanGlScott: > `hvr-ghcjs` neither `nodejs` APT repositories aren't on the whitelist, and there is no way to put them there at the moment

Sigh. I suppose there's nothing we can do about that, then.

In any case, it seems that my knowledge of how Travis works is very out of date. According to https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration, it seems that container-based builds (i.e., ones that use `sudo: false`) have actually been phased out. In fact, all Travis builds have been using VM-based builds since December 2018, and the `sudo` keyword is now deprecated! In that case, I take back my concerns about `sudo: true`—they're entirely unfounded in reality.

This does make me wonder though: do we even need to enable `sudo: true` at all? If I'm reading that blog post correctly, then the behavior of `sudo: true` is always enabled, regardless of configuration.

phadej: being explicit is better. but you are right, seems sudo is always there now according to https://docs.travis-ci.com/user/reference/overview/

phadej: I'll remove `sudo: true`.

.travis.yml
```diff
@@ -54,19 +56,37 @@ matrix:
...
     - compiler: ghc-head
     - compiler: ghc-8.8.1
 before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" = "linux" ]; then
+      sudo add-apt-repository -y ppa:hvr/ghcjs;
+      curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -
+      sudo apt-add-repository 'https://deb.nodesource.com/node_8.x xenial main'
+      sudo apt-get update;
+      sudo apt-get install $CC cabal-install-3.0 nodejs freeglut3-dev;
```
RyanGlScott: While installing `nodejs` with `apt-get install` appears to (unfortunately) be necessary, do we need to install these other things (`cabal-install-3.0`, `freeglut3-dev`, etc.) with `apt-get install` now that we're back on Xenial? Furthermore, do we need to install `nodejs` on non-GHCJS configurations?

phadej: Good point, let me split that line.
Trying to `apt-get install` what's already installed is no-op, so installing `$CC freeglug3-dev` is cheap.

phadej: done.

phadej:
Now, `HLint` finds more issues :(

phadej:
Now this PR is a lot tidier :)

lens.cabal
phadej: Found: https://github.com/ghcjs/ghcjs/tree/ghc-8.6/lib/boot/shims there are shims

.travis.yml
```diff
@@ -57,8 +59,27 @@ matrix:
...
 before_install:
+  - |
+    if [ "$TRAVIS_OS_NAME" = "linux" ]; then
+      if $GHCJS ; then sudo add-apt-repository -y ppa:hvr/ghcjs ; fi;
+      if $GHCJS ; then curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add - ; fi;
+      if $GHCJS ; then sudo apt-add-repository 'https://deb.nodesource.com/node_8.x xenial main' ; fi;
+      sudo apt-get update;
+      sudo apt-get install $CC freeglut3-dev;
+      if $GHCJS ; then sudo apt-get install -y nodejs cabal-install-3.0 ; fi;
+    fi
```
RyanGlScott: Apologies for focusing so much on this part, but this still seems wrong to me. I would have expected this part of the script to (1) only run if `$GHCJS` is set, and (2) only download `ghcjs` and `nodejs`, since those are the only packages that we can't list in `addons`. AFAICT, (1) isn't happening because this happens on any Linux builder (which means that we needlessly call `apt-get update` on non-GHCJS configurations), and (2) isn't happening because we install things besides `ghcjs`/`nodejs`.

This extra work can add up in the total build time. For instance, it takes 54.77s to run all of this on the GHC 8.6 configuration, as shown [here](https://travis-ci.org/ekmett/lens/jobs/572064517#L377).

Why can't we just do this?

```
  - |
    if [ "$TRAVIS_OS_NAME" = "linux" && $GHCJS ]; then
      sudo add-apt-repository -y ppa:hvr/ghcjs;
      curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | sudo apt-key add -;
      sudo apt-add-repository 'https://deb.nodesource.com/node_8.x xenial main';
      sudo apt-get update;
      sudo apt-get install -y $CC nodejs;
    fi
```

.travis.yml
```diff
@@ -57,8 +59,27 @@ matrix:
...
+  - |
+    if echo $CC | grep -q ghcjs; then
+        GHCJS=true
+        HC=${HC}js
+        WITHCOMPILER="--ghcjs ${WITHCOMPILER}js"
+    else
+        GHCJS=false;
+    fi
+  - if $GHCJS ; then sudo apt-get install -y ghc-8.4.4 ; fi
+  - if $GHCJS ; then PATH="/opt/ghc/8.4.4/bin:$PATH" ; fi
```
RyanGlScott: Similarly, should we be hardcoding `ghc/8.4.4` here?

.travis.yml
```diff
@@ -57,8 +59,27 @@ matrix:
...
   - WITHCOMPILER="-w $HC"
+  - |
+    if echo $CC | grep -q ghcjs; then
+        GHCJS=true
+        HC=${HC}js
+        WITHCOMPILER="--ghcjs ${WITHCOMPILER}js"
+    else
+        GHCJS=false;
+    fi
+  - if $GHCJS ; then sudo apt-get install -y ghc-8.4.4 ; fi
```
RyanGlScott: Should we be hardcoding `ghc-8.4.4` like this? If we ever support multiple GHCJS versions, it seems like we would need to use something like `ghc-${GHCJSNUMVER}`, right?

.travis.yml
```diff
@@ -171,17 +195,17 @@ script:
   # build & run tests, build benchmarks
   - ${CABAL} v2-build $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output
   # Testing...
-  - ${CABAL} v2-test $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output
+  - if ! $GHCJS ; then ${CABAL} v2-test $WITHCOMPILER ${TEST} ${BENCH} all | color_cabal_output ; fi
```
RyanGlScott: Why is `v2-test` not enabled at all for GHCJS? I do realize that three of the test suites have to be disabled due to uninstallable dependencies, but the fourth test suite (`templates`) should still be runnable, right?

phadej: https://github.com/haskell/cabal/issues/6175

Even one can do something like `find dist-newstyle -name all.js` to find all executable-likes built, I haven't found (yet?) a way to pickup exactly the test-suites from there.

.travis.yml
phadej: It's an issue to solve when there's some other GHCJS than 8.4 in hvr's (or actually any) apt repository.

.travis.yml
phadej: This step takes <1s on a non GHCJS build now.

RyanGlScott:
This looks pretty solid to me.

One last question: is this intended as a tech demo, or do you wish to merge this into `master`? I ask since, AFAICT, there are no GHCJS-specific code paths in `lens`. On the other hand, libraries like `semigroups` [do have GHCJS-specific code paths](https://github.com/ekmett/semigroups/blob/42aabc50f1713bd032dd5a9caddf7caaa40429d7/src-ghc7/Data/Semigroup.hs#L1212), so testing them against GHCJS would be quite valuable.

phadej:
It was kind of tech-demo yes, as `lens` stresses stuff. Up to you, whether you want to have one more job or not.

RyanGlScott:
My view is that the increase in CI turnabout time isn't justified unless we have code paths in `lens` that are only reachable on GHCJS. As far as I can tell, that isn't the case here.

It might be nice to add the `if !impl(ghcjs)` parts to `lens.cabal` separately, however.

phadej:
if`impl(ghcjs)` pieces are in tests, it's not worth it if CI isn't running them

PR #880, opened 2019-08-14
Fix HLint (regenerate .travis.yml)

phadej:
Let's fix HLint before adding GHCJS job.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -8,6 +8,11 @@
 - ignore: {name: Use lambda-case}
 - ignore: {name: Use tuple-section}
 - ignore: {name: Use fewer imports}
+- ignore: {name: "Use :"}
+- ignore: {name: Use typeRep} # typeRep is since base-4.7
+- ignore: {name: Eta reduce, within: [Control.Lens.Zoom]} # Breaks code
+- ignore: {name: Use camelCase, within: [Control.Lens.Internal.TH]}
+- ignore: {name: Use list comprehension, within: [Control.Lens.Internal.FieldTH]}
 
 - fixity: "infixl 9 :>"
 - fixity: "infixr 9 <.>, <., .>, ..."
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.3.20190425
+# version: 0.3.20190814.1
 #
 language: c
 dist: xenial
@@ -35,7 +35,6 @@ matrix:
   include:
     - compiler: ghc-8.8.1
       addons: {"apt":{"sources":["hvr-ghc"],"packages":["ghc-8.8.1","cabal-install-3.0","freeglut3-dev"]}}
-      env: GHCHEAD=true
     - compiler: ghc-8.6.5
       addons: {"apt":{"sources":["hvr-ghc"],"packages":["ghc-8.6.5","cabal-install-2.4","freeglut3-dev"]}}
     - compiler: ghc-8.4.4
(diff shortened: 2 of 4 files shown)


RyanGlScott:
I definitely agree about ignoring the `Use :` suggestions (that comes down to personal preference, IMO) and the `Eta reduce` suggestion (since eta-expansion is necessary to work around old GHC bugs in `Control.Lens.Zoom`).

On the other hand, the `Use typeRep` suggestion is actually somewhat reasonble, IMO, since I find `typeRep (Proxy :: Proxy a)` somewhat less scary than `typeOf (undefined :: a)`. Perhaps we should just apply those suggestions?

Taneb:
`typeRep` seems to only have been available since GHC 7.8, as a point against using it.

RyanGlScott:
True, although we could define our own shimmed version of `typeRep` pretty easily. We already have a [shim for `eqT`](https://github.com/ekmett/lens/blob/8fc321358d341bdc1fb8267846909151ef426b3a/src/Control/Lens/Internal/Equality.hs#L84-L86), in fact, so a shimmed `typeRep` would fit in nicely alongside it.

phadej:
@RyanGlScott if `eqT` and `typeRep` would leave in some `Control.Lens.Internal.Compat` module, than maybe. I could do that when `type-equality` story starts to unfold.

RyanGlScott:
Right—if we were to offload the work of backporting `(:~:)(..)` and friends to a separate package, then the only things that would be left in `Control.Lens.Internal.Equality` would be `eqT` and `typeRep`, which makes the name `Control.Lens.Internal.Equality` somewhat less apt. Perhaps at that point we should rename it to something like `Control.Lens.Internal.Typeable` to better reflect its purpose.

In any case, I agree that this is something that doesn't need to be done in this very PR. That being said, there are still some [remaining `hlint` failures](https://travis-ci.org/ekmett/lens/jobs/571887372#L1610):

```
$ if [ $HCNUMVER -ge 80600 ] && [ $HCNUMVER -lt 80800 ] ; then (cd ${PKGDIR_lens} && hlint src) ; fi
Warning: Can't find file "HsBaseConfig.h" in directories
	src/Control/Lens
	.
	include
  Asked for by: src/Control/Lens/Wrapped.hs  at line 83 col 1
src/Control/Lens/Internal/FieldTH.hs:270:30: Suggestion: Use list comprehension
Found:
  if classExists || Set.member c locals then [] else
    [makeFieldClass defType c n]
Perhaps:
  [makeFieldClass defType c n |
   not (classExists || Set.member c locals)]

src/Control/Lens/Internal/TH.hs:35:1: Error: Parse error
Found:
    --
    ----------------------------------------------------------------------------
  > module Control.Lens.Internal.TH where

    import Data.Functor.Contravariant


2 hints
The command "if [ $HCNUMVER -ge 80600 ] && [ $HCNUMVER -lt 80800 ] ; then (cd ${PKGDIR_lens} && hlint src) ; fi" exited with 1.
```

I'm ambivalent about the first one; I'd be fine with applying the suggestion or just ignoring it. On the other hand, I have no idea what is going on with the second error.

RyanGlScott:
I know why the second error happens now. It's because `Control.Lens.Internal.TH` uses an `ANN` pragma _before_ the `module` section, like so:

```hs
{-# LANGUAGE CPP #-}
#ifdef HLINT
{-# ANN module "HLint: ignore Use camelCase" #-}
#endif
module Control.Lens.Internal.TH where
```

Per `hlint-2.1.23`'s [changelog](http://hackage.haskell.org/package/hlint-2.1.23/changelog), `hlint` will now throw an error if your code does not parse with GHC itself. Indeed, that code will not compile if you pass `-DHLINT`, as GHC's parser expects `ANN` pragmas to come after the `module` section, like so:

```hs
{-# LANGUAGE CPP #-}
module Control.Lens.Internal.TH where

-- Imports

#ifdef HLINT
{-# ANN module "HLint: ignore Use camelCase" #-}
#endif
```

phadej:
I removed the `ANN` pragma foir that module. I'd remove it from other modules as well, as I don't want to encourage its use. IMHO it's silly that GHC uses TH-like mechanisms for these.

RyanGlScott:
I agree that we should migrate all these `ANN`s over to `.hlint.yaml` at some point. In any case, the build is now fixed—thanks!

PR #884, opened 2019-09-06
Type equality

phadej:
EDIT: `type-equality` released


src/Data/Data/Lens.hs
```diff
@@ -64,11 +64,13 @@ import qualified Data.HashSet as S
 import           Data.HashSet (HashSet)
 import           Data.IORef
 import           Data.Monoid
+import           Data.Proxy (Proxy (..))
 import           GHC.Exts (realWorld#)
 import           Prelude
 
 #if !(MIN_VERSION_base(4,7,0))
```
RyanGlScott: It seems silly to have to have to use CPP here now that we've gone through the effort of creating compatibility shims for these things. Could we avoid the use of CPP here by importing `Control.Lens.Internal.Typeable` and `Data.Type.Equality` qualified?

phadej: Fair point, I (hopefully) improved the situation

PR #885, opened 2019-09-06
Tasty

phadej:
Based on top #884 

Change to tasty is motivated by
- `test-framework` is not so well maintained
- and depends on `regex-base` which needs release for ghc-8.8.1
- and we could run tests with GHCJS (#879)


RyanGlScott:
While I don't have a particular preference for `test-framework` versus `tasty`, I can't say I fully understand the reasons you've listed.

> * `test-framework` is not so well maintained

In what sense?

> * and depends on `regex-base` which needs release for ghc-8.8.1

This is true, but I'm quite reluctant to rip out `test-framework` entirely for the sake of one dependency that hasn't updated to GHC 8.8.1 yet—8.8.1 is still quite recent, after all. Plus, there are other things that `lens`'s benchmarks and examples depend on that haven't updated to 8.8.1 yet, so we'll need to continue using `head.hackage` for 8.8.1 support for a little while longer.

> * and we could run tests with GHCJS (#879)

Again, I'm not sure that ripping out `test-framework` is the best way to achieve GHCJS test support. How difficult would it be to patch things on `test-framework`'s end? I'm not sure if it's possible to patch `regex-posix` directly, given its reliance on a C regex library, but perhaps `test-framework` could be patched to use a different regex library on GHCJS? In fact, I'm somewhat surprised that `test-framework` uses `regex-posix` at all, given that it's known to be [severely buggy on Windows](https://github.com/haskell/cabal/issues/4336).

phadej:
I'm not motivated to pursuit "let's fix `test-framework`" path. See who makes maintenance PRs to it https://github.com/haskell/test-framework/pulls?q=is%3Apr+is%3Aclosed, with `tasty` it's someone else.

PR #886, opened 2019-09-12
Remove -traditional flag from cpp

Taneb:
This seems to have been breaking profiling builds on GHC 8.8.


RyanGlScott:
I've uploaded [`lens-4.18.1`](http://hackage.haskell.org/package/lens-4.18.1) to Hackage with this fix.

PR #889, opened 2019-09-20
Depend on type-equality conditionally

phadej:


phadej:
@RyanGlScott is this worth changelog entry?

RyanGlScott:
This is a very minor change, but if you think it's worth mentioning, then go for it.

phadej:
Nah, I'm lazy.

PR #897, opened 2019-12-07
At: implement multi for chaining maps and sets

infinity0:
Pretty self-explanatory, see the doctest examples.

Happy to rename it if people think "multi" isn't a great name. "chain" is another possibility. Both are pretty generic...

### DIFF
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -40,6 +40,7 @@ module Control.Lens.At
   -- * At
     At(at)
     , sans
+    , multi
     , iat
   -- * Ixed
   , Index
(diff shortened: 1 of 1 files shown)


infinity0:
I suppose one potential issue is that `multi` is not technically a `Lens` since if you put in `Just mempty` then you'll get back out `Nothing`, but then again it's not advertised as a valid lens and is only meant to be combined with `at` and `contains`. If it's a big deal perhaps I could move it into `Control.Lens.Unsound` instead.

phadej:
How is this different from [`non`](https://hackage.haskell.org/package/lens-4.18.1/docs/Control-Lens-Iso.html#v:non) ?

infinity0:
Oh great, I thought I might have missed something. I'll just use that then, thanks.

PR #899, opened 2019-12-10
Re-enable -Wunused-matches

RyanGlScott:
Commit b2e0efe6842f25c545291579baecf8790ee24636 added `-Wno-unused-matches` to `lens.cabal` in an effort to work around [GHC#11451](https://gitlab.haskell.org/ghc/ghc/issues/11451), which affected GHC 8.0.1-rc1 (the newest compiler at the time). However, #11451 has since been fixed, and as far as I can tell, it does not affect the latest point release of GHC 8.0 (8.0.2). In light of this, it seems a bit silly to me to specifically disable this warning when it no longer provides any tangible benefit. Let's reenable it.

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -345,9 +345,8 @@ library
   if impl(ghc<7.4)
     ghc-options: -fno-spec-constr-count
 
-  -- hack around the buggy unused matches check for class associated types in ghc 8 rc1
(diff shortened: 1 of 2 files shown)


Taneb:
I think this is a good idea.

PR #909, opened 2020-01-18
Require semigroup in <>~

phadej:
I'm keen on doing this, as the next release could be the `4.19` if we also include e.g. GHC-8.10 required template-haskell changes

This change is "requested" on IRC

```
<idnar> is there a Semigroup version of (<>~)?
```

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,7 +1,9 @@
-next [????.??.??]
+4.19 [????.??.??]
 -----------------
 * Add `{Functor,Foldable,Traversable}WithIndex` instances for
   `Control.Applicative.Const` and `Data.Functor.Constant.Constant`.
+* Make `<>~` and `<>=` and their `<op` and `<<op` state variants require only
+  `Semigroup`, not `Monoid`
 
 4.18.1 [2019.09.13]
 -------------------
--- a/lens.cabal
+++ b/lens.cabal
@@ -1,6 +1,6 @@
 name:          lens
 category:      Data, Lenses, Generics
-version:       4.18.1
+version:       4.19
 license:       BSD2
 cabal-version: 1.18
 license-file:  LICENSE
--- a/lens.cabal
+++ b/lens.cabal
@@ -327,6 +327,7 @@ library
     Numeric.Natural.Lens
 
   other-modules:
+    Control.Lens.Internal.Prelude
     Paths_lens
 
   if flag(safe)
@@ -464,6 +465,7 @@ benchmark plated
   default-language: Haskell2010
   build-depends:
     base,
+    base-compat >=0.11.0 && <0.12,
     comonad,
     criterion,
     deepseq,
--- a/src/Control/Lens/Lens.hs
+++ b/src/Control/Lens/Lens.hs
@@ -132,10 +132,12 @@ module Control.Lens.Lens
   , fusing
   ) where
 
-import Control.Applicative
+import Prelude ()
+
 import Control.Arrow
 import Control.Comonad
 import Control.Lens.Internal.Context
+import Control.Lens.Internal.Prelude
 import Control.Lens.Internal.Getter
 import Control.Lens.Internal.Indexed
 import Control.Lens.Type
@@ -147,16 +149,7 @@ import Data.Profunctor
 import Data.Profunctor.Rep
 import Data.Profunctor.Sieve
 import Data.Profunctor.Unsafe
-import Data.Semigroup (Semigroup (..))
 import Data.Semigroup.Traversable
-import Data.Void
-import Prelude
-#if MIN_VERSION_base(4,8,0)
-import Data.Function ((&))
-#endif
-#if MIN_VERSION_base(4,11,0)
-import Data.Functor ((<&>))
-#endif
 #if __GLASGOW_HASKELL__ >= 800
 import GHC.Exts (TYPE)
 #endif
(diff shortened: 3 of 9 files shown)


src/Control/Lens/Setter.hs
```diff
@@ -1043,7 +1045,7 @@ l <?= b = do
   return b
 {-# INLINE (<?=) #-}
 
--- | Modify the target of a monoidally valued by 'mappend'ing another value.
+-- | Modify the target of a semigroup value by '<>'ing another value.
```
RyanGlScott: "`'<>'ing`" reads strangely to me. I would instead opt for this sort of phrasing:

```suggestion
-- | Modify the target of a 'Semigroup' value by using ('<>').
```

src/Control/Lens/Setter.hs
```diff
@@ -1055,16 +1057,16 @@ l <?= b = do
...
+-- ('<>~') :: 'Semigroup' a => 'Traversal' s t a a -> a -> s -> t
 -- @
-(<>~) :: Monoid a => ASetter s t a a -> a -> s -> t
-l <>~ n = over l (`mappend` n)
+(<>~) :: Semigroup a => ASetter s t a a -> a -> s -> t
+l <>~ n = over l (<> n)
 {-# INLINE (<>~) #-}
 
--- | Modify the target(s) of a 'Lens'', 'Iso', 'Setter' or 'Traversal' by 'mappend'ing a value.
+-- | Modify the target(s) of a 'Lens'', 'Iso', 'Setter' or 'Traversal' by '<>'ing a value.
```
RyanGlScott:
```suggestion
-- | Modify the target(s) of a 'Lens'', 'Iso', 'Setter' or 'Traversal' by using ('<>').
```

phadej:
Good catch on other `<>` operators. Will do the changes on them too.

CHANGELOG.markdown
```diff
@@ -1,7 +1,8 @@
-next [????.??.??]
+4.19 [????.??.??]
```
RyanGlScott: Go ahead and bump the version in `lens.cabal`, while you're at it.

CHANGELOG.markdown
```diff
@@ -1,7 +1,8 @@
...
 -----------------
 * Add `{Functor,Foldable,Traversable}WithIndex` instances for
   `Control.Applicative.Const` and `Data.Functor.Constant.Constant`.
+* Make `<>~` and `<>=` require only `Semigroup`, not `Monoid`
```
RyanGlScott: Don't forget to mention the other functions that were changed!

src/Control/Lens/Lens.hs
```diff
@@ -143,11 +143,11 @@ import Control.Monad.State as State
 import Data.Functor.Apply
 import Data.Functor.Reverse
 import Data.Functor.Yoneda
-import Data.Monoid
 import Data.Profunctor
 import Data.Profunctor.Rep
 import Data.Profunctor.Sieve
 import Data.Profunctor.Unsafe
+import Data.Semigroup (Semigroup (..))
```
RyanGlScott: Guard this behind `#if !(MIN_VERSION_base(4,11,0))` to avoid redundant import warnings on recent GHCs.

src/Control/Lens/Setter.hs
```diff
@@ -91,11 +91,12 @@ import Control.Monad.State.Class  as State
 import Control.Monad.Writer.Class as Writer
 import Data.Functor.Contravariant
 import Data.Functor.Identity
-import Data.Monoid
+import Data.Monoid (Monoid (..))
 import Data.Profunctor
 import Data.Profunctor.Rep
 import Data.Profunctor.Sieve
 import Data.Profunctor.Unsafe
+import Data.Semigroup (Semigroup (..))
```
RyanGlScott: Guard this behind `#if !(MIN_VERSION_base(4,11,0))` to avoid redundant import warnings on recent GHCs.

src/Control/Lens/Setter.hs
```diff
@@ -91,11 +91,12 @@ import Control.Monad.State.Class  as State
 import Control.Monad.Writer.Class as Writer
 import Data.Functor.Contravariant
 import Data.Functor.Identity
-import Data.Monoid
+import Data.Monoid (Monoid (..))
```
RyanGlScott: Guard this behind `#if !(MIN_VERSION_base(4,8,0))` to avoid redundant import warnings on recent GHCs.

CHANGELOG.markdown
phadej: Done

CHANGELOG.markdown
phadej: Done

src/Control/Lens/Setter.hs
phadej: I made `Control.Lens.Internal.Prelude`, got tired of copy & pasting CPP around.

lens.cabal
```diff
@@ -327,6 +327,7 @@ library
     Numeric.Natural.Lens
 
   other-modules:
+    Control.Lens.Internal.Prelude
```
RyanGlScott: By convention, `lens` puts all its `Control.Lens.Internal` modules under `exposed-modules`. There's also another benefit in doing this: there is [some CPP in `benchmarks/plated.hs`](https://github.com/ekmett/lens/blob/eb255ec6c1e5c4f0012ab56268b43a3059d60a16/benchmarks/plated.hs#L10-L12) that could be done away with by importing `Control.Lens.Internal.Prelude` instead.

src/Control/Lens/Unsound.hs
```diff
@@ -5,10 +5,6 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# OPTIONS_GHC -fno-warn-warnings-deprecations #-}
 
-#ifndef MIN_VERSION_mtl
```
RyanGlScott: I wonder if all these `#ifndef` statements should go into `include/lens-common.h` at some point... but let's not do so in this PR, since there's already enough code churn in this commit as it is.

lens.cabal
phadej: I'd avoid that for a time being, until it's truly not changing that much. The `Control.Lens.Internal` module are way too stable

RyanGlScott: The whole point of the `Internal` label is that we exempt it from PVP guarantees, right? Or an I missing something?

phadej: I don't want to increase API surface of `lens`. I'd then rather remove the commit and span the CPP all over the place again.

RyanGlScott: OK, but please make a note of the fact that `C.L.Internal.Prelude`

1. Is not exported, and
2. This convention differs from other `C.L.Internal` modules

In the Haddocks for `C.L.Internal.Prelude`.

phadej: Done, also

```
    Depend on base-compat in plated benchmarks. criterion depends
    on base-compat-batteries anyway.
```

PR #910, opened 2020-01-20
Consolidate CPP in lens-common.h, lens.cabal, and C.L.Internal.Prelude

phadej:
... and using `Control.Lens.Internal.Prelude` remove `{-# LANGUAGE CPP #-}` from few modules entirely.

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -346,8 +346,12 @@ library
   if impl(ghc<7.4)
     ghc-options: -fno-spec-constr-count
 
+  if impl(ghc >= 7.10)
+    ghc-options: -fno-warn-trustworthy-safe
+
   if impl(ghc >= 8)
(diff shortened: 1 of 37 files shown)


lens.cabal
```diff
@@ -348,6 +348,8 @@ library
 
   if impl(ghc >= 8)
     ghc-options: -Wno-missing-pattern-synonym-signatures
+    ghc-options: -Wno-trustworthy-safe
```
RyanGlScott: If we're going to enable this flag, we ought to guard it behind `if impl(ghc >= 7.10)`, since that's when this flag was introduced. Currently, we get warnings on GHC 7.10 as a result of not doing this ([example](https://travis-ci.org/ekmett/lens/jobs/639404566#L1055)).

phadej: Done. Build `-Werror` clean when tested locally, on all GHC since 7.4

PR #911, opened 2020-01-20
More use of Internal.Prelude

phadej:
I think we lie about `safe` flag, but I think that with `Internal.Prelude` it's actually easier to make that promise true, if we really want to.

The diff TL;DR is, more re-exports in `Internal.Prelude`, less imports elsewhere. Especially less unqualified wild imports

### DIFF
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -43,7 +43,10 @@ module Control.Lens.At
   , icontains
   ) where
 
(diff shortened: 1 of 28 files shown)


phadej:
Let's see, AFAIK there aren't big base changes in GHC-8.10 version, "sadly" we won't see if `Internal.Prelude` would helped with migration. I think current state is good, and we can improve as we go from now.

PR #916, opened 2020-03-24
Add inon, a version of non to be used for caching

ddssff:
I had a long bewildering experience using regular non for maintaining a cache, so I thought this modification might be a useful addition to the library.

### DIFF
--- a/src/Control/Lens/Iso.hs
+++ b/src/Control/Lens/Iso.hs
@@ -48,6 +48,7 @@ module Control.Lens.Iso
   , simple
   , non, non'
   , anon
+  , inon
   , enum
   , curried, uncurried
   , flipped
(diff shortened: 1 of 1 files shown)


src/Control/Lens/Iso.hs
```diff
@@ -397,6 +398,32 @@ anon a p = iso (fromMaybe a) go where
...
+-- from the map.  You have now lost the information that that user was
+-- ever present in the map:
+--
+-- >>> fromList [("user1", "!!!")] & (at "user1" . non def) .~ ""
+-- fromlist []
+--
+-- Using 'inon' instead of 'non' means that key will not be deleted
+-- from the map:
+--
+-- >>> fromList [("user1", "!!!")] & (at "user1" . inon def) .~ ""
```
phadej: Don't

```haskell
fromList [("user1", "!!!")] & at "user1" ?~ ""
```

has same effect?

And for caching, why the removal of defautl value would be bad. I don't understand that either.

ddssff: I'll answer the second question.  The scenario is that your cache is a Map, and the set of keys is an important piece of information, you don't want to forget that a key was once an element of the map.  Picture the user list of a web app, you don't want the user to disappear if they reset their user profile to the default.

ddssff: As for the first question, yes that has the same effect, but in my use case I am constructing the lens with non and then composing further lenses to look inside the value.  This lens gets passed to a function, so we no longer have access to the lens that looks at the map itself.

ddssff: like this:

```
>>> (fromList [("user1", "axc")]) & (at "user1" . non "abc" . ix 1) .~ 'b'`
fromList []
```

src/Control/Lens/Iso.hs
```diff
@@ -397,6 +398,32 @@ anon a p = iso (fromMaybe a) go where
...
+-- >>> fromList [("user1", "axc")] & (at "user1" . non "abc" . ix 1) .~ 'b'
+-- fromlist []
+--
+-- Using 'inon' instead of 'non' means that key will not be deleted
+-- from the map:
+--
+-- >>> fromList [("user1", "axc")] & (at "user1" . inon "abc" . ix 1) .~ 'b'
+-- fromlist [("user1", "abc")]
+inon :: a -> Iso' (Maybe a) a
+inon a = anon a (const False)
```
phadej: I have a gut feeling that you be ok with `inon :: a -> Setter (Maybe a) a`, which isn't as unlawful. (It still changes `Nothing` to `Just a` when you `%~ id`).

I'd ask you to make it a setter and move to `Unsound` module.

ddssff: Then I would have to pass both a getter and a setter to the function, which seems the opposite direction from where I am trying to go.  I will give it some thought.

phadej: I'll think more about this too.

ddssff: Ultimately I just started writing `anon v (const False)` to avoid this problem.

PR #920, opened 2020-04-15
Remove the test suite's dependency on test-framework-th

### LINKED ISSUE #918: Figure out what to do about the language-haskell-extract dependency
I am not yet able to run `lens`' test suite + GHC 8.10.1 on Travis CI without the use of `head.hackage`. Why, you ask? It is because various test suites depend on `test-framework-th`, which in turn depends on `language-haskell-extract`. Sadly, `language-haskell-extract` [does not build with `template-haskell-2.16.*`](https://github.com/finnsson/template-helper/issues/12), and the maintainer of the library has not been active on GitHub for about two years. Bottom line: `language-haskell-extract` is likely abandonware.

This poses a problem for `lens`. I can see two ways forward:

1. Beg someone to do a package takeover of `language-haskell-extract` and upload a new version.
2. Remove `lens`' dependency on `test-framework-th`. As far as I can tell, that package only provides a mild syntactic convenience in the form of automatic test discovery, so removing it shouldn't be too difficult.

RyanGlScott:
This is primarily motivated by `test-framework-th`'s dependency on `language-haskell-extract`, which does not build on GHC 8.10+ and appears to be abandoned by its maintainer. See the discussion at #918 for more details.

Removing this dependency finally allows us to build `lens` with GHC 8.10.1 on Travis without the need for `head.hackage`, so I took the liberty of regenerating `.travis.yml` in this patch as well.

Fixes #918.

### DIFF
--- a/tests/hunit.hs
+++ b/tests/hunit.hs
@@ -34,7 +34,6 @@ import Data.Map as Map
 import Data.Monoid
 #endif
 import Test.Framework.Providers.HUnit
-import Test.Framework.TH
 import Test.Framework
 import Test.HUnit hiding (test)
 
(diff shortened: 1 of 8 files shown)


tests/hunit.hs
```diff
@@ -285,4 +284,42 @@ case_write_through_list_entry =
...
+    , testCase "read maybe map entry" case_read_maybe_map_entry
+    , testCase "read maybe state map entry" case_read_maybe_state_map_entry
+    , testCase "read map entry" case_read_map_entry
+    , testCase "read state map entry" case_read_state_map_entry
+    , testCase "modify map entry" case_modify_map_entry
+    , testCase "insert maybe map entry" case_insert_maybe_map_entry
+    , testCase "delete maybe map entry" case_delete_maybe_map_entry
+    , testCase "read list entry" case_read_list_entry
+    , testCase "write list entry" case_write_list_entry
+    , testCase "write through list entry" case_write_through_list_entry
```
phadej, line 323: Is this essentially what `-ddump-slices` dumped?

If so, this is :+1:

RyanGlScott: Indeed, this is painstakingly copied over from the `-ddump-splices` output. If it weren't for that, I wouldn't know what names to give these test cases! :)

PR #922, opened 2020-04-28
Add compatibility functions for Profunctor optics

guaraqe:
This PR adds two functions:

- `overP`, which transforms a `Lens` into a profunctor one, based on `overA`;
- `applyPrism`, which transforms a `Prism` into a profunctor one.

These are very simple functions, which shouldn't introduce any extra burden to the maintenance of the library. They also open the way for some interoperation with profunctor optics libraries.

I think the name of `overP` is good, since it follows the pattern of `overA`. I choose `applyPrism` so that it can be used as:

```haskell
applyPrism _Left :: Choice p => p a b -> p (Either a c) (Either b c)
```

I don't know if any `INLINE`s would be appropriate.

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -285,6 +285,7 @@ library
     Control.Lens.Operators
     Control.Lens.Plated
     Control.Lens.Prism
+    Control.Lens.Profunctor
     Control.Lens.Reified
     Control.Lens.Review
     Control.Lens.Setter
--- a/src/Control/Lens/Profunctor.hs
+++ b/src/Control/Lens/Profunctor.hs
@@ -1,5 +1,13 @@
 {-# LANGUAGE RankNTypes #-}
 
+-------------------------------------------------------------------------------
+-- | This module provides conversion functions between the optics defined in
+-- this library and 'Profunctor'-based optics.
+--
+-- The goal of these functions is to provide an interoperability layer between
+-- the two styles of optics, and not to reimplement all the library in terms of
+-- 'Profunctor' optics.
+
 module Control.Lens.Profunctor
   ( -- * Conversion from Van Laarhoven optics
     fromLens
@@ -13,7 +21,7 @@ module Control.Lens.Profunctor
   , toTraversal
   ) where
 
-import Control.Applicative (pure)
+import Control.Applicative (Applicative (..))
 import Control.Lens.Type (Lens, Prism, Setter, Traversal)
 import Control.Lens.Internal.Context (Context (..), sell)
 import Data.Functor.Identity (Identity (..))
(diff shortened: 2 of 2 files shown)


phadej:
Let's assume we want this. I think, then we should rather put these conversion functions in separate module as we would like to:
- convert back and forth
- also convert
    - Traversals
    - Setters

guaraqe:
Conversion of `Traversal`s is done by `wander`, that is defined in the `Traversing` typeclass of `profunctors`.

What would be the equivalent profunctor optics for `Setter`? I cannot see a typeclass in `profunctors` that correspond to it.

I choose to add them in the respective modules because they are rather simple or already defined in practice.

If you think the best is separating such functions in a module, what would be its name?
For the functions, should we follow some kind of standard naming (such as `from/toLens`), or just use names that feel more intuitive?

phadej:
> Conversion of Traversals is done by wander, that is defined in the Traversing typeclass of profunctors.

Yes. But it is good to give it a easy to find name.

> What would be the equivalent profunctor optics for Setter? I cannot see a typeclass in profunctors that correspond to it.

`Mapping` with `roam` for to-conversion

> I choose to add them in the respective modules because they are rather simple (applyPrism) or already defined in practice (overP).

It is valid. My assumption that if people will look for conversions to/from profunctor optics, the separate module will stand out better (also naming becomes easier).

> If you think the best is separating such functions in a module, what would be its name?

Good question. I think that `Control.Lens.Profunctor` is a ok compromise, given the module header explains that the module provides conversion functions -- doesn't re-implement all of `Control.Lens` as profunctor optics :)

> For the functions, should we follow some kind of standard naming (such as from/toLens), or just use names that feel more intuitive?

`fromLens` / `toLens` are good.

guaraqe:
I added the functions, including those for `Setter` and `Traversal` in a separated module.

I see that all modules in `lens` have a specific header, should I add one, together with some documentation?

phadej:
>  I think that Control.Lens.Profunctor is a ok compromise, given the module header explains that the module provides conversion functions

guaraqe:
Sorry, I missed that part of the phrase. I added documentation to the module.

Travis is complaining about the lack of `roam`, which should not happen with the boundaries given to `profunctor` in the `cabal` file, is that normal?

guaraqe:
Only to make clear, the dependency error only happens for GHC 7.6 and 7.4.

phadej:
I can fix the GHC-7.6 and 7.4 failures. And also remove the need for `RankNTypes`.

@RyanGlScott what you think about this?

RyanGlScott:
> @RyanGlScott what you think about this?

I'm admittedly not well versed on profunctor optics, so if you feel like this is a worthwhile addition, I'll trust your judgment.

guaraqe:
If there is anything else I can do to improve the PR or fix errors, I would be happy to do it.

phadej:
Closed in favor of https://github.com/ekmett/lens/pull/923

PR #924, opened 2020-05-18
Import liftM2 from Control.Monad instead of Control.Monad.Reader

Bodigrim:
There is an intention to stop re-exporting functions from `base` in future versions of `mtl`: https://github.com/haskell/mtl/pull/74
This PR brings `lens` in line with the proposed change, because at the moment `Control.Lens.Internal.Zoom` imports `liftM2` (originally from `Control.Monad`) from `Control.Monad.Reader`.


phadej:
Thanks

PR #927, opened 2020-05-25
Safe Haskell support: Mark Control.Lens.Equality as Trustworthy

int-e:
The module imports TYPE (which is harmess) from GHC.Exts (which is unsafe).

### DIFF
--- a/src/Control/Lens/Equality.hs
+++ b/src/Control/Lens/Equality.hs
@@ -8,6 +8,7 @@
 #endif
 #if __GLASGOW_HASKELL__ >= 800
 {-# LANGUAGE TypeInType #-}
+{-# LANGUAGE Trustworthy #-}
 #endif
 
 -----------------------------------------------------------------------------


phadej:
Thanks!

phadej:
FWIW, @RyanGlScott the GHC-8.10.1 has `-Wwarn=inferred-safe-imports`, so we could groom Safe Haskell support in KmettVerse if we are ready to push lower-bounds to the versions with explicit Safe Haskell annotations. I'm willing to do it for `lens` dependency closure.

RyanGlScott:
I'll echo my comments in https://github.com/ekmett/semigroupoids/issues/69#issuecomment-361304259: if you're willing to put in the work to make everything `Safe`, then go for it. But be prepared to fix things if they break in the future :)

phadej:
That's over two year old comment now, good to know that your opinion haven't changed :)

PR #931, opened 2020-07-16
Drop pre ghc 7.8

### LINKED ISSUE #930: Drop pre-GHC-7.8 support
I think it's time. `profunctors` have dropped it. Bolder move would be to drop pre-GHC-8.0 support, that would clean a lot more of CPP. Yet my motivation is to cleanup things around `Coercible`, which is also why `profunctors-5.3` dropped pre-GHC-7.8 support.

GHC-7.8.4 was released on December 23rd 2014, i.e. over fiver years ago.

Opinions, @ekmett @RyanGlScott

phadej:
Resolve #930


src/Control/Lens/Internal/ByteString.hs
```diff
@@ -234,41 +215,14 @@ c2w = fromIntegral . ord
 
 -- | Unpack a strict 'B.Bytestring'
 unpackStrict :: B.ByteString -> [Word8]
```
RyanGlScott, line 217: Is there a good reason not to just remove this function and replace its use sites with `B.unpack`? I know that this function is exported by this module, but then again, this is an internal module, so we can make breaking changes here without fear.

src/Control/Lens/Internal/ByteString.hs
```diff
@@ -234,41 +215,14 @@ c2w = fromIntegral . ord
...
-               | otherwise = let !x = BI.inlinePerformIO $ do
-                                        x' <- peek p
-                                        touchForeignPtr fp
-                                        return x'
-                             in x : go (p `plusPtr` 1) q
-#endif
 {-# INLINE unpackStrict #-}
 
 -- | Unpack a strict 'B.Bytestring', pretending the bytes are chars.
 unpackStrict8 :: B.ByteString -> String
```
RyanGlScott, line 222: Similarly, is there a good reason not to remove this function and replace its use sites with `B8.unpack`?

src/Control/Lens/Internal/Setter.hs
```diff
@@ -1,8 +1,6 @@
 {-# LANGUAGE CPP #-}
 
-#if __GLASGOW_HASKELL__ < 708
 {-# LANGUAGE Trustworthy #-}
```
RyanGlScott, line 3: Previously, this line was only reachable on pre-7.8 versions of GHC, but now it's used unconditionally. Did you intend to make this change?

src/Control/Lens/Internal/TH.hs
```diff
@@ -59,11 +59,7 @@ conAppsT :: Name -> [Type] -> Type
 conAppsT conName = foldl AppT (ConT conName)
 
 fromSet :: (k -> v) -> Set.Set k -> Map.Map k v
```
RyanGlScott, line 61: Similarly, is there a good reason not to remove this function and replace its use sites with `Map.fromSet`?

src/Control/Lens/Internal/TH.hs
```diff
@@ -216,7 +212,6 @@ inlinePragma :: Name -> [DecQ]
 
 #ifdef INLINING
 
-#if MIN_VERSION_template_haskell(2,8,0)
 
 # ifdef OLD_INLINE_PRAGMAS
 -- 7.6rc1?
```
RyanGlScott, line 217: This comment (and similar comments below) suggest that this is only used on circa-7.6 versions of GHC, which makes me wonder if this code can be trimmed down even further.

src/Language/Haskell/TH/Lens.hs
```diff
@@ -1738,52 +1668,29 @@ _Interruptible
...
-  = prism' reviewer remitter
-  where
-      reviewer (x, y) = InlineP x y
-      remitter (InlineP x y) = Just (x, y)
-      remitter _ = Nothing
-#endif
 
 -- |
 -- @
 -- _SpecialiseP :: 'Prism'' 'Pragma' ('Name', 'Type', 'Maybe' 'Inline', 'Phases') -- template-haskell-2.8+
```
RyanGlScott, line 1682: Now that the API for `_SpecialiseP` is uniform across all supported versions of `template-haskell`, this Haddock comment is completely redundant. I'd favor removing it entirely.

src/Language/Haskell/TH/Lens.hs
```diff
@@ -1738,52 +1668,29 @@ _Interruptible
 -- |
 -- @
 -- _InlineP :: 'Prism'' 'Pragma' ('Name', 'Inline', 'RuleMatch', 'Phases') -- template-haskell-2.8+
```
RyanGlScott, line 1670: Now that the API for `_InlineP` is uniform across all supported versions of `template-haskell`, this Haddock comment is completely redundant. I'd favor removing it entirely.

src/Control/Lens/Equality.hs
```diff
@@ -74,11 +72,7 @@ data Identical a b s t where
   Identical :: Identical a b a b
 
 -- | When you see this as an argument to a function, it expects an 'Equality'.
-#if __GLASGOW_HASKELL__ >= 706
-type AnEquality (s :: k1) (t :: k2) (a :: k1) (b :: k2) = Identical a (Proxy b) a (Proxy b) -> Identical a (Proxy b) s (Proxy t)
-#else
 type AnEquality s t a b = Identical a (Proxy b) a (Proxy b) -> Identical a (Proxy b) s (Proxy t)
```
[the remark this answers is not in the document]
phadej: Mistake :)

src/Control/Lens/Equality.hs
```diff
@@ -102,10 +96,8 @@ substEq l = case runEq l of
 -- | We can use 'Equality' to do substitution into anything.
 #if __GLASGOW_HASKELL__ >= 800
 mapEq :: forall k1 k2 (s :: k1) (t :: k2) (a :: k1) (b :: k2) (f :: k1 -> Type) . AnEquality s t a b -> f s -> f a
```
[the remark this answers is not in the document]
phadej: I try to keep this massive patch "line based", so won't do this simplification.

src/Control/Lens/Internal/ByteString.hs
phadej: I thought that, but there's also `unpackLazy` (which is also an alias definition `= L.unpack`), so decided to not remove them in this patch.

src/Control/Lens/Internal/Setter.hs
phadej: Yes. I think this was something on the lines "inferred safe / trustworthy". Unconditional `Trustworthy` is clearer.

src/Control/Lens/Internal/TH.hs
phadej: Done.

src/Control/Lens/Internal/TH.hs
phadej: `OLD_INLINE_PRAGMAS` is toggable by flags. I haven't looked closely what they are about.

src/Control/Lens/Getter.hs
```diff
@@ -3,11 +3,8 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE Trustworthy #-}
-
-#if __GLASGOW_HASKELL__ && __GLASGOW_HASKELL__ >= 704
 {-# LANGUAGE NoPolyKinds #-}
```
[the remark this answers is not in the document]
phadej: Removed

src/Control/Lens/Internal/ByteString.hs
phadej: Ditto.

src/Control/Exception/Lens.hs
```diff
@@ -632,7 +630,6 @@ pattern Denormal_ <- (has _Denormal -> True) where
   Denormal_ = review _Denormal ()
 #endif
 
-#if MIN_VERSION_base(4,6,0)
 -- | Added in @base@ 4.6 in response to this libraries discussion:
```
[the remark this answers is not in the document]
phadej: Done

phadej:
Cleaned up missed 706 CPPs too.

src/Control/Lens/Equality.hs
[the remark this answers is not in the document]
RyanGlScott: Very well. Can you open an issue about this so that it does not get forgotten?

src/Control/Lens/Internal/ByteString.hs
RyanGlScott: Again, please open an issue about this so that it does not get forgotten.

src/Control/Lens/Internal/TH.hs
RyanGlScott: This arises from here:

https://github.com/ekmett/lens/blob/c073e468b67683c2a92adc5c80251b840612b8e3/lens.cabal#L343-L344

Which, in turn, has this description:

https://github.com/ekmett/lens/blob/c073e468b67683c2a92adc5c80251b840612b8e3/lens.cabal#L155-L158

Based on this, this appears to be a hack to support a release candidate version of GHC. (And an ancient one at that!) Yuck. I feel pretty confident that we can just rip out this hack now.

Speaking of which, there's several more unnecessary `impl`s in the `.cabal` file. I found the following:

* https://github.com/ekmett/lens/blob/c073e468b67683c2a92adc5c80251b840612b8e3/lens.cabal#L241-L243
* https://github.com/ekmett/lens/blob/c073e468b67683c2a92adc5c80251b840612b8e3/lens.cabal#L339-L341
* https://github.com/ekmett/lens/blob/c073e468b67683c2a92adc5c80251b840612b8e3/lens.cabal#L349-L350
* https://github.com/ekmett/lens/blob/c073e468b67683c2a92adc5c80251b840612b8e3/lens.cabal#L359-L360

Let's drop these `impl`s as well.

src/Control/Lens/Internal/Indexed.hs
```diff
@@ -157,10 +154,20 @@ instance Profunctor (Indexed i) where
 #ifndef SAFE
   ( .# ) ibc _ = coerce ibc
```
RyanGlScott, line 155: I see you did some cleanup relying on `lens`' CI using a more recent version of `hlint`, which is fine. There's another `hlint`-related simplification that I've been meaning to do for a while now involving lines like this one, where we write things like `( #. )`. There extra spaces surrounding the parentheses are quite strange, but they were necessary to work around an old `hlint` parsing bug. That bug has long been fixed, so we should be able to remove the extraneous spaces now.

Keep in mind that the line above is far from the only occurrence of this phenomenon. You don't necessarily have to perform this simplification in this PR, but if you decide not to, we should open an issue about this so that it is not forgotten.

src/Control/Lens/Internal/Indexed.hs
```diff
@@ -157,10 +154,20 @@ instance Profunctor (Indexed i) where
...
   {-# INLINE ( .# ) #-}
+#if __GLASGOW_HASKELL__ < 710
```
RyanGlScott, line 157: Hah, I didn't expect this patch to _introduce_ CPP. How unfortunate. Surely there is a way to avoid this somehow? Does the trick that the `profunctors` library use [here](https://github.com/ekmett/profunctors/blob/2ef115beef4c513b8b3ca3f1bc4abdf648182cc0/src/Data/Profunctor/Unsafe.hs#L175) also work here?

src/Control/Lens/Internal/Indexed.hs
phadej: I'll leave this to you.

src/Control/Lens/Internal/Indexed.hs
phadej: Well, I'd leave it here and rather drop GHC-7.10 in the year or two. CPP is explicit

RyanGlScott: Very well. But at the very least I'd like to minimize the use of CPP to one location, if possible. `coerce'` is defined four separate times in its own CPP block, and each use site of `coerce'` also uses its own CPP! Better yet to have a single definition for `coerce`', like this:

```hs
-- | Used only to work around a GHC 7.8-specific `Coercible` solver bug when implementing ('#.').
coerce' :: forall a b. Coercible a b => b -> a
#if __GLASGOW_HASKELL__ >= 710
coerce' = coerce
#else
coerce' = coerce (id :: a -> a)
#endif
{-# INLINE coerce' #-}
```

phadej: Yes, I removed `Internal.Coercible` with a hope that it's not necessary. But seems it is.

phadej:
I lost steam with this for today. Let's try again next week.

phadej:
Looks like it's new week in some alternative universe.

src/Control/Lens/Internal/Coerce.hs
```diff
@@ -16,9 +10,7 @@
 -- Stability   :  experimental
 -- Portability :  non-portable
 --
--- This module provides a shim around 'coerce' that defaults to 'unsafeCoerce'
--- on GHC < 7.8. It also exposes a type-restricted version of '(#.)' that
--- works around a bizarre GHC 7.10–specific bug.
+-- This module provided a shim around 'coerce'.
```
RyanGlScott, line 13: s/provided/provides/

Also, it would be worthwhile to say _why_ this shim exists. Previously, it was due to the need to use `unsafeCoerce`, but now it's due to the need to work around a GHC 7.8-specific `Coercible` solver bug. Say that.

benchmarks/traversals.hs
```diff
@@ -20,27 +20,6 @@ import Control.Lens
 import Criterion.Main
 import Criterion.Types
 
-#if !(MIN_VERSION_bytestring(0,10,0))
```
RyanGlScott, line 23: Now that this code is gone, we should be able to remove the `{-# LANGUAGE CPP #-}` and `{-# OPTIONS_GHC -fno-warn-orphans #-}` lines above.

src/Control/Lens/Internal.hs
```diff
@@ -41,7 +41,3 @@ import Control.Lens.Internal.Prism
 import Control.Lens.Internal.Review
 import Control.Lens.Internal.Setter
 import Control.Lens.Internal.Zoom
-
-#ifdef HLINT
```
RyanGlScott, line 45: Now that this is gone, we should be able to get rid of the `{-# LANGUAGE CPP #-}` line above.

src/Control/Lens/Internal/Context.hs
```diff
@@ -63,9 +61,8 @@ class IndexedFunctor w where
 -- "Parameterized Notions of Computation" <http://bentnib.org/paramnotions-jfp.pdf>
 -- and that construction is dualized here.
 class IndexedFunctor w => IndexedComonad w where
-#if __GLASGOW_HASKELL__ >= 708
```
RyanGlScott, line 66: Now that this code is gone, we should be able to get rid of the `{-# LANGUAGE CPP #-}` line above.

src/Control/Lens/Internal/Indexed.hs
RyanGlScott: I've done this in #933.

src/Control/Lens/Internal/ByteString.hs
RyanGlScott: I've done this in #933.

PR #937, opened 2020-08-07
Add an operator for constructing singleton maps

endgame:
I've often wanted a compact way to make singleton maps, and lens gives me generic tools to do that.

I'm not wedded to the use of `(==>)`; it matches a similar operator from Data.Dependent.Sum which is nice but clashes with an operator from QuickCheck. Maybe `(=:)`? Maybe something else?

I've picked `infixr 1` pretty arbitrarily but maybe a different fixity is appropriate.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -4,6 +4,7 @@ next [2020.xx.yy]
 * Add `Control.Lens.Profunctor` with conversion functions to and from
   profunctor optic representation
 * Mark `Control.Lens.Equality` as Trustworthy
+* Add `Control.Lens.At.==>` for constructing singleton maps.
 
 4.19.2 [2020.04.15]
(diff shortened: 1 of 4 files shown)


phadej:
I'm not the fan of ad-hoc combinators. See e.g. https://github.com/ekmett/lens/pull/708

>  To throw in the list of rules this needs to be efficient and the link it draws between Empty and Cons gives me a bad taste.

Here the link is drawn between Empty and At (which isn't that bad, but still).

---

`==>` name is terrible. This is ought to be non-infix. `singleton` is not super good name, as it could as well mean a `\x -> cons x Empty`.

emilypi:
Why not just `oneOf :: (At m, AsEmpty m) => Index m -> IxValue m -> m` or `singletonOf`?

PR #938, opened 2020-08-09
Remove the last unsafeCoerce

phadej:
`SAFE` CPP definition wasn't used anywhere anymore.

### DIFF
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -3,5 +3,3 @@ unconstrained:          False
 hlint:                  True
 apt:                    freeglut3-dev
 irc-channels:           irc.freenode.org#haskell-lens
--- Work around https://github.com/haskell/haddock/issues/242
-haddock:                <7.6 || >=7.8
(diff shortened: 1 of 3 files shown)


cabal.haskell-ci
```diff
@@ -3,5 +3,3 @@ unconstrained:          False
 hlint:                  True
 apt:                    freeglut3-dev
 irc-channels:           irc.freenode.org#haskell-lens
--- Work around https://github.com/haskell/haddock/issues/242
-haddock:                <7.6 || >=7.8
```
RyanGlScott, line 7: Out of curiosity, how did you figure out that this was no longer necessary? I have to use this hack on other projects as well, and I'd be interested to know if there is a general approach to removing the hack.

phadej: We dropped support for older GHCs :)

RyanGlScott: Oho. Right you are. (We really should put this fact in the `CHANGELOG`!)

PR #939, opened 2020-08-09
Use strict and assoc

### LINKED ISSUE #936: Documentation of `lazy` mentions `strict-base-types`.
https://hackage.haskell.org/package/lens-4.19.2/docs/Control-Lens-Iso.html#v:lazy

> See http://hackage.haskell.org/package/strict-base-types for an example use.

Yet, with release of 0.4 https://hackage.haskell.org/package/strict, strict-base-types is being deprecated. (I'll soon upload strict-lens package).

So, what to do with `Strict` class in `lens`? Options:

1. Remove it in favor of `Strict`-class from `strict`, define `strict = iso toStrict toLazy`. This requires adding dependency on `strict`.
2. Correct haddock, but leave everything in place
3. Something else?

I'm in favor of going all the way for 1.

phadej:
Fixes #936. We also depend on assoc for Swap(ped), as `strict`
depends on `assoc` to provide `Swap` instances.

I hope that some day `Swap` class will migrate to `bifunctors`
or even to `base`.

There is `strict-lens`, with `Data.Strict.Lens` module,
adding some optics for types in `Data.Strict`.
It can be merged into `lens` or kept separate.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -4,6 +4,26 @@
 * Add `Control.Lens.Profunctor` with conversion functions to and from
   profunctor optic representation
 * Mark `Control.Lens.Equality` as Trustworthy
+* The `Swapped` type class is removed in favor of `Swap` from the `assoc` package
+* The `Strict` type class is removed in favor of `Strict` from the `strict` package
+
+  The `swapped`, `strict` and `lazy` isomorphisms are now defined using "new" type classes.
+
+  Users which define own instances of old type classes are advised to
+  define instances of the new ones.
+
+  ```haskell
+  import qualified Data.Bifunctor.Swap as Swap
+  import qualified Control.Lens        as Lens
+
+  instance Swap.Swap MyType where
+    swap = ...
+
+  #if !MIN_VERSION_lens(4,20,0)
+  instance Lens.Swapped MyType where
+    swapped = iso Swap.swap Swap.swap
+  #endif
+  ```
 
 4.19.2 [2020.04.15]
 -------------------
--- a/src/Control/Lens/Each.hs
+++ b/src/Control/Lens/Each.hs
@@ -55,6 +55,8 @@ import Data.Vector.Storable (Storable)
 import qualified Data.Vector.Unboxed as Unboxed
 import Data.Vector.Unboxed (Unbox)
 import Data.Word
+import qualified Data.Strict as S
+import Data.These (These (..))
 
 -- $setup
 -- >>> :set -XNoOverloadedStrings
@@ -221,3 +223,37 @@ instance (Ix i, i ~ j) => Each (Array i a) (Array j b) a b where
 instance (Ix i, IArray UArray a, IArray UArray b, i ~ j) => Each (UArray i a) (UArray j b) a b where
   each f arr = array (bounds arr) <$> traverse (\(i,a) -> (,) i <$> f a) (IArray.assocs arr)
   {-# INLINE each #-}
+
+-------------------------------------------------------------------------------
+-- strict
+-------------------------------------------------------------------------------
+
+-- | @since 4.20
+instance (a ~ a', b ~ b') => Each (S.Either a a') (S.Either b b') a b where
+    each f (S.Left x)  = S.Left <$> f x
+    each f (S.Right x) = S.Right <$> f x
+    {-# INLINE each #-}
+
+-- | @since 4.20
+instance (a~a', b~b') => Each (S.Pair a a') (S.Pair b b') a b where
+    each f (a S.:!: b) = (S.:!:) <$> f a <*> f b
+    {-# INLINE each #-}
+
+-- | @since 4.20
+instance Each (S.Maybe a) (S.Maybe b) a b
+
+-- | @since 4.20
+instance (a ~ a', b ~ b') => Each (S.These a a') (S.These b b') a b where
+    each f (S.This a)    = S.This <$> f a
+    each f (S.That b)    = S.That <$> f b
+    each f (S.These a b) = S.These <$> f a <*> f b
+
+-------------------------------------------------------------------------------
+-- these
+-------------------------------------------------------------------------------
+
+-- | @since 4.20
+instance (a ~ a', b ~ b') => Each (These a a') (These b b') a b where
+    each f (This a)    = This <$> f a
+    each f (That b)    = That <$> f b
+    each f (These a b) = These <$> f a <*> f b
--- a/src/Control/Lens/Iso.hs
+++ b/src/Control/Lens/Iso.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE Rank2Types #-}
-{-# LANGUAGE FunctionalDependencies #-}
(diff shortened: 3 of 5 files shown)


phadej:
If this is fine, I'll add changelog afterwards (also remembering to mention that pre GHC-7.8 support is dropped).

src/Control/Lens/Iso.hs
```diff
@@ -51,16 +51,15 @@ module Control.Lens.Iso
   , enum
   , curried, uncurried
   , flipped
-  , Swapped(..)
+  , swapped
```
RyanGlScott, line 53: This is a breaking change, so now would be a good time to bump `lens`'s version number to `4.20`.

src/Control/Lens/Iso.hs
```diff
@@ -427,81 +409,25 @@ flipped :: Iso (a -> b -> c) (a' -> b' -> c') (b -> a -> c) (b' -> a' -> c')
...
+-- 'bimap' f g '.' 'swapped' = 'swapped' '.' 'bimap' g f
+-- @
+--
+-- >>> (1,2)^.swapped
+-- (2,1)
+swapped :: Swap p => Iso (p a b) (p c d) (p b a) (p d c)
+swapped = iso swap swap
+{-# INLINE swapped #-}
 
 -- | Ad hoc conversion between \"strict\" and \"lazy\" versions of a structure,
```
RyanGlScott, line 426: These Haddocks previously described the `Strict` class, but they now feel a bit off when describing the `strict` function, which is a combinator built on top of `Strict`. This may come down to personal taste, but I would prefer making the Haddocks more closely resemble those of `lazy`, i.e.,

```hs
-- | An 'Iso' between the lazy variant of a structure and its strict
-- counterpart.
--
-- @
-- 'strict' = 'from' 'lazy'
-- @
```

src/Control/Lens/Iso.hs
```diff
@@ -427,81 +409,25 @@ flipped :: Iso (a -> b -> c) (a' -> b' -> c') (b -> a -> c) (b' -> a' -> c')
 flipped = iso flip flip
 {-# INLINE flipped #-}
 
--- | This class provides for symmetric bifunctors.
-class Bifunctor p => Swapped p where
```
RyanGlScott, line 431: I hadn't considered the need to migrate `Swapped`, but I guess if we're going to migrate `Strict`, then `Swapped` would also be included. On the other hand, unlike `Strict`, there are actually uses of `Swapped` in Hackage libraries "in the wild" (a quick search reveals `accuerr`, `pipes-key-value-csv`, `validation`, and `separated`). As a result, please document carefully how to migrate from `lens:Swapped` to `assoc:Swap` in the `CHANGELOG` (in addition to how to migrate from `lens:Strict` to `strict:Strict`).

src/Control/Lens/Iso.hs
phadej: Good point. I expect the compatibility changes for GHC-9.0 to introduce some other breaking changes too, so I'd rather hold up this PR until then. I'll improve other aspects though now.

src/Control/Lens/Iso.hs
phadej: It's not strictly necessary. If `lens` depends on `strict`, than it makes sense to move all instances from `strict-lens` to `lens`.

`Swapped` is one. Removing whole class is easier approach :)

That reminds me, that I have to add `Each` instances for `strict` types to `lens`.

src/Control/Lens/Iso.hs
RyanGlScott: > I expect the compatibility changes for GHC-9.0 to introduce some other breaking changes too, so I'd rather hold up this PR until then.

I've just landed #942, which adds support for GHC 9.0 and bumps the major version number.

phadej: Are the breaking change in `TH` modules, as the #942 changelog entries don't highlight what is broken.

RyanGlScott: > Are the breaking change in `TH` modules

That's right.

phadej: Ok. I'll come back to this ASAP, in a day or two. Lets have more breaking changes :)

CHANGELOG.markdown
```diff
@@ -4,6 +4,8 @@ next [2020.xx.yy]
 * Add `Control.Lens.Profunctor` with conversion functions to and from
   profunctor optic representation
 * Mark `Control.Lens.Equality` as Trustworthy
+* `Swapped` type-class is removed in favor of `Swap` from `assoc` package
```
RyanGlScott, line 7:
```suggestion
* The `Swapped` type class is removed in favor of `Swap` from the `assoc` package
```

CHANGELOG.markdown
```diff
@@ -4,6 +4,8 @@ next [2020.xx.yy]
...
+* `Strict` type-class is removed in favor of `Strict from `strict` package
```
RyanGlScott, line 8:
```suggestion
* The `Strict` type class is removed in favor of `Strict` from the `strict` package
```

src/Control/Lens/Each.hs
```diff
@@ -221,3 +223,37 @@ instance (Ix i, i ~ j) => Each (Array i a) (Array j b) a b where
...
+
+-- | @since 4.20
+instance (a ~ a', b ~ b') => Each (S.Either a a') (S.Either b b') a b where
+    each f (S.Left x)  = S.Left <$> f x
+    each f (S.Right x) = S.Right <$> f x
+    {-# INLINE each #-}
+
+-- | @since 4.20
+instance (a~a', b~b') => Each (S.Pair a a') (S.Pair b b') a b where
+    each f ~(a S.:!: b) = (S.:!:) <$> f a <*> f b
```
RyanGlScott, line 239: The use of irrefutable pattern matching on a strict pair is somewhat peculiar. Is there a reason for this?

CHANGELOG.markdown
```diff
@@ -4,6 +4,8 @@
 * Add `Control.Lens.Profunctor` with conversion functions to and from
   profunctor optic representation
 * Mark `Control.Lens.Equality` as Trustworthy
+* `Swapped` type-class is removed in favor of `Swap` from `assoc` package
```
RyanGlScott, line 7: These changelog entries should briefly describe how to migrate from `lens:Swapped` and `lens:Strict` to `assoc:Swap` and `strict:Strict`, respectively.

src/Control/Lens/Each.hs
phadej: With the irrefutable pattern:

```haskell
>>> set each 'x' (error "foo") :: Pair Char Char
'x' :!: 'x'
```

without

```haskell
set each 'x' (error "foo") :: Pair Char Char
*** Exception: foo
```

RyanGlScott: Ah, right. I only considered a situation where you force one of the elements of the pair (and necessarily forcing the other element of the pair to WHNF), but I completely overlooked the situation where you force _zero_ elements!

RyanGlScott (approved):

PR #942, opened 2020-10-02
Allow building with GHC 9.0

RyanGlScott:
~~There are still some downstream dependency issues involving `th-abstraction` (most likely bos/aeson#809) that prevent a build plan from being discovered.~~ **EDIT**: All fixed now!

### DIFF
--- a/src/Control/Lens/Prism.hs
+++ b/src/Control/Lens/Prism.hs
@@ -95,7 +95,7 @@ withPrism k f = case coerce (k (Market Identity Right)) of
 --
 -- See 'Control.Lens.Lens.cloneLens' and 'Control.Lens.Traversal.cloneTraversal' for examples of why you might want to do this.
 clonePrism :: APrism s t a b -> Prism s t a b
-clonePrism k = withPrism k prism
+clonePrism k = withPrism k $ \bt sta -> prism bt sta
 {-# INLINE clonePrism #-}
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


src/Control/Lens/Prism.hs
```diff
@@ -132,13 +132,13 @@ outside k = withPrism k $ \bt seta f ft ->
...
         -> Prism (Either s u) (Either t v) (Either a c) (Either b d)
-without k =
-  withPrism k         $ \bt seta k' ->
-  withPrism k'        $ \dv uevc    ->
-  prism (bimap bt dv) $ \su ->
-  case su of
-    Left s  -> bimap Left Left (seta s)
-    Right u -> bimap Right Right (uevc u)
+without k uvcd =
+  (withPrism k         $ \bt seta k' ->
```
RyanGlScott, line 136: Can anyone think of a cleaner way to accomplish this?

phadej: Manually inlining `withPrism` definition...

phadej: Or just

```
without :: APrism s t a b
        -> APrism u v c d
        -> Prism (Either s u) (Either t v) (Either a c) (Either b d)
without k k' =
   withPrism k         $ \bt seta ->
   withPrism k'        $ \dv uevc    ->
   prism (bimap bt dv) $ \su ->
   case su of
     Left s  -> bimap Left Left (seta s)
     Right u -> bimap Right Right (uevc u)
```

RyanGlScott: https://github.com/ekmett/lens/pull/942#discussion_r498707113 is a fantastic idea. Suggestion applied.

PR #943, opened 2020-10-04
Make Control.Lens the first exposed module

phadej:


RyanGlScott:
Silly question: what exactly is the difference in `cabal repl` after this change? I would have thought that `cabal repl` would load `Control.Lens` either way.

phadej:
Currently `cabal repl` loads `*Control.Exception.Lens` first.

RyanGlScott:
Wow! I had no idea that the order of modules in `exposed-modules` affected which modules are loaded by default in `cabal repl` like this. Is this documented anywhere?

In any case, I agree that `Control.Lens` is probably what you want loaded by default.

phadej:
> Is this documented anywhere?

I think it is expected behavior that the modules are passed to GHC in order they are defined in `exposed-modules` (and then `other-modules`, see note). What GHCi does with that module list is upto GHCi, and I'm not sure that is documented and whether there is e.g. flags / ways to affect which module is loaded first.

note: the way Cabal implementation works the file

```cabal
exposed-modules: Mod1
other-modules: Mod2
if flag(foo)
  exposed-modules: Mod5
exposed-modules: Mod3
other-modules: Mod4
```

is the same as 

```cabal
exposed-modules: Mod1
exposed-modules: Mod3
other-modules: Mod2
other-modules: Mod4
if flag(foo)
  exposed-modules: Mod5
```

And I think this is documented somewhere (order matters for `ghc-options` e.g.) but I'm not 100% sure.

PR #945, opened 2020-10-08
Make the TH machinery handle PolyKinds more robustly

### LINKED ISSUE #917: Kind and type vars sometimes out of dependency order with makePrisms/makeLenses
I ran into an obscure issue with the `makePrisms`/`makeLenses` TH, which occurs when the kinds of different type variables are related.

Compiling the following:

```haskell
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE PolyKinds #-}
module Test where
import Control.Lens

data Foo (a :: k -> *) (b :: k -> *) = Foo

data Thing a b = Thing (Foo a (Const b))
$(makePrisms ''Thing)
```

... with `-ddump-splices` yields:
```
[1 of 1] Compiling Test             ( Test.hs, Test.o )
Test.hs:9:3-20: Splicing declarations
    makePrisms ''Thing
  ======>
    _Thing ::
      forall a_a3qw b_a3qx k_a3vt a_a74S b_a74T.
      Iso (Thing a_a74S b_a74T) (Thing a_a3qw b_a3qx) (Foo a_a74S (Const b_a74T :: k_a3vt
                                                                                   -> GHC.Types.Type)) (Foo a_a3qw (Const b_a3qx :: k_a3vt
                                                                                                                                    -> GHC.Types.Type))
    _Thing
      = (iso (\ (Thing x1_a74U) -> x1_a74U)) (\ x1_a74V -> Thing x1_a74V)
    {-# INLINE _Thing #-}

Test.hs:9:3: error:
    • These kind and type variables: a_a3qw b_a3qx k_a3vt a_a74S b_a74T
      are out of dependency order. Perhaps try this ordering:
        k_a3vt (a_a3qw :: k0 -> *) b_a3qx (a_a74S :: k0 -> *) b_a74T
    • In the type signature:
        _Thing :: forall a_a3qw b_a3qx k_a3vt a_a74S b_a74T.
                  Iso (Thing a_a74S b_a74T) (Thing a_a3qw b_a3qx) (Foo a_a74S (Const b_a74T :: k_a3vt
                                                                                               -> GHC.Types.Type)) (Foo a_a3qw (Const b_a3qx :: k_a3vt
                                                                                                                                                -> GHC.Types.Type))
  |
9 | $(makePrisms ''Thing)
```

One notices that the type signature starts with `forall a b k a' b'. ...`, which gets it into trouble, as the kinds of `a` and `b` are `k -> *` (i.e. `k -> Type`).

Looking at `makeFieldOptic` in `Control.Lens.Internal.FieldTH`, I can't easily tell why this would be the case.

RyanGlScott:
This is a collection of various Template Haskell–related fixes that, when all put together, fixes #917. This does the following:

* Rather than use `th-abstraction`'s `datatypeType` function, which strips off important kind information from type arguments, I defined a similar `datatypeTypeKinded` function that preserves kinds.
* `Control.Lens.Internal.{FieldTH,PrismTH}` is now more careful to use `freeVariablesWellScoped` (from `th-abstraction`) instead of `typeVars` to ensure that the resulting types are well scoped. This is particularly important for poly-kinded types, as the kind variables must always appear before the type variables.
* I deleted the `close` function from `Control.Lens.Internal.PrismTH` in favor of `quantifyType` and `quantifyType'`, which I have moved to `Control.Lens.Internal.TH` so that they may be used by `FieldTH` and `PrismTH` alike. Moreover, I now use `quantifyType'` in the definition of `PrismTH.makeClassyPrismClass` so that any type variables bound by the class itself do not get requantified in any class methods. The previous code was not doing this at all, which was just plain wrong.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -21,6 +21,13 @@
     swapped = iso Swap.swap Swap.swap
   #endif
   ```
+* Make the functions in `Control.Lens.TH` work more robustly with poly-kinded
+  data types. This can cause a breaking change under certain situations:
+  * TH-generated optics for poly-kinded data types are now much more likely to
+    mention kind variables in their definitions, which will require enabling
+    the `PolyKinds` extension at use sites in order to typecheck.
+  * Because TH-generated optics now quantify more kind variables than they did
+    previously, this can affect the order of visible type applications.
 * Add `Control.Lens.Profunctor` with conversion functions to and from
   profunctor optic representation
 * Add `Control.Lens.Review.reviewing`, which is like `review` but with a more
--- a/lens.cabal
+++ b/lens.cabal
@@ -212,7 +212,7 @@ library
     tagged                    >= 0.4.4    && < 1,
     template-haskell          >= 2.9.0.0  && < 2.18,
     these                     >= 1.1.1.1  && < 1.2,
-    th-abstraction            >= 0.4      && < 0.5,
+    th-abstraction            >= 0.4.1    && < 0.5,
     text                      >= 1.2.3.0  && < 1.3,
     transformers              >= 0.3.0.0  && < 0.6,
     transformers-compat       >= 0.4      && < 1,
@@ -353,10 +353,15 @@ library
 test-suite templates
   type: exitcode-stdio-1.0
   main-is: templates.hs
-  other-modules: T799
+  other-modules:
+    T799
+    T917
   ghc-options: -Wall -threaded
   hs-source-dirs: tests
   default-language: Haskell2010
+
+  if impl(ghc >= 8.6)
+    ghc-options: -Wno-star-is-type
 
   if flag(dump-splices)
     ghc-options: -ddump-splices
--- a/src/Control/Lens/Internal/FieldTH.hs
+++ b/src/Control/Lens/Internal/FieldTH.hs
@@ -95,9 +95,9 @@ makeFieldOpticsForDatatype rules info =
                      return (concat decss)
 
   where
-  tyName = D.datatypeName info
-  s      = D.datatypeType info
-  cons   = D.datatypeCons info
+  tyName = D.datatypeName     info
+  s      = datatypeTypeKinded info
+  cons   = D.datatypeCons     info
 
   -- Traverse the field labels of a normalized constructor
   normFieldLabels :: Traversal [(Name,[(a,Type)])] [(Name,[(b,Type)])] a b
@@ -320,12 +320,13 @@ makeClassyClass className methodName s defs = do
   let ss   = map (stabToS . view (_2 . _2)) defs
   (sub,s') <- unifyTypes (s : ss)
   c <- newName "c"
-  let vars = toListOf typeVars s'
+  let vars     = D.freeVariablesWellScoped [s']
+      varNames = map D.tvName vars
       fd   | null vars = []
-           | otherwise = [FunDep [c] vars]
+           | otherwise = [FunDep [c] varNames]
 
 
-  classD (cxt[]) className (map D.plainTV (c:vars)) fd
+  classD (cxt[]) className (D.plainTV c:vars) fd
     $ sigD methodName (return (lens'TypeName `conAppsT` [VarT c, s']))
     : concat
       [ [sigD defName (return ty)
@@ -334,7 +335,7 @@ makeClassyClass className methodName s defs = do
         inlinePragma defName
       | (TopName defName, (_, stab, _)) <- defs
       , let body = appsE [varE composeValName, varE methodName, varE defName]
-      , let ty   = quantifyType' (Set.fromList (c:vars))
+      , let ty   = quantifyType' (Set.fromList (c:varNames))
                                  (stabToContext stab)
                  $ stabToOptic stab `conAppsT`
                        [VarT c, applyTypeSubst sub (stabToA stab)]
@@ -356,8 +357,8 @@ makeClassyInstance rules className methodName s defs = do
            : map return (concat methodss)
 
   where
-  instanceHead = className `conAppsT` (s : map VarT vars)
-  vars         = toListOf typeVars s
+  instanceHead = className `conAppsT` (s : map tvbToType vars)
+  vars         = D.freeVariablesWellScoped [s]
   rules'       = rules { _generateSigs    = False
                        , _generateClasses = False
                        }
@@ -620,23 +621,3 @@ type HasFieldClasses = StateT (Set Name) Q
 
 addFieldClassName :: Name -> HasFieldClasses ()
 addFieldClassName n = modify $ Set.insert n
-
-------------------------------------------------------------------------
--- Miscellaneous utility functions
-------------------------------------------------------------------------
-
-
--- | Template Haskell wants type variables declared in a forall, so
--- we find all free type variables in a given type and declare them.
-quantifyType :: Cxt -> Type -> Type
-quantifyType = quantifyType' Set.empty
-
--- | This function works like 'quantifyType' except that it takes
--- a list of variables to exclude from quantification.
-quantifyType' :: Set Name -> Cxt -> Type -> Type
-quantifyType' exclude c t = ForallT vs c t
-  where
-  vs = map D.plainTVSpecified
-     $ filter (`Set.notMember` exclude)
-     $ nub -- stable order
-     $ toListOf typeVars t
--- a/src/Control/Lens/Internal/PrismTH.hs
+++ b/src/Control/Lens/Internal/PrismTH.hs
@@ -23,7 +23,6 @@ module Control.Lens.Internal.PrismTH
   ) where
 
 import Control.Applicative
-import Control.Lens.Fold
 import Control.Lens.Getter
 import Control.Lens.Internal.TH
 import Control.Lens.Lens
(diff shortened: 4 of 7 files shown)


RyanGlScott:
I have marked this as an WIP for now since I had to vendor in an unreleased version of `th-abstraction`. I do this because the `datatypeTypeKinded` refactor above caused me to run into glguy/th-abstraction#84. To work around this, I have temporarily vendored in a version of `th-abstraction` with a fix (glguy/th-abstraction#85), so once that fix makes it to Hackage, I can simply adjust the lower version bounds on `th-abstraction` accordingly.

RyanGlScott:
For what it's worth, this is definitely a breaking change, since any code that uses `makeLenses`, `makePrisms`, etc. with a data type with one or more type parameters will now be required to enable `KindSignatures`. As one example, this will break the `wreq` library:

<details>

```
[ 9 of 15] Compiling Network.Wreq.Lens.TH ( Network/Wreq/Lens/TH.hs, interpreted )

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseVersion’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_a5Pi’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseVersion’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseStatus’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_a5Pi’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseStatus’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseHeaders’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_a5Pi’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseHeaders’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseCookieJar’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_a5Pi’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseCookieJar’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseClose'’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_a5Pi’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseClose'’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseBody’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘GHC.Types.Type’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseBody’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_a5Pi’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseBody’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^

Network/Wreq/Lens/TH.hs:84:1: error:
    Illegal kind signature: ‘body_akQ5’
      Perhaps you intended to use KindSignatures
    In the type signature for ‘responseBody’
   |
84 | makeLenses ''HTTP.Response
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
```
</details>

This isn't a huge breakage, since one can fix it by simply enabling a language extension. But I thought this worth mentioning nonetheless.

arybczak:
FWIW when I fixed this in https://github.com/well-typed/optics/pull/313 I only annotated types with kind variables, not monomorphic kinds, so that `KindSignatures` is not needed (unless you use `PolyKinds`, but then it's implied anyway).

RyanGlScott:
I fear that the situation is slightly more complicated than just detecting kind variables. Consider this program:

```hs
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
{-# OPTIONS_GHC -ddump-splices #-}
module Foo where

import Control.Lens
import Data.Kind
import Data.Proxy

data family DF (a :: k)
data instance DF (a :: Type) = MkDF { _unDF :: Proxy a }
$(makeLenses 'MkDF)
```

If you generate this code without any kind signatures, you'll get the following:

```
/home/rgscott/Documents/Hacking/Haskell/Foo.hs:13:3-18: Splicing declarations
    makeLenses 'MkDF
  ======>
    unDF ::
      forall a_a44D a_a45f.
      Iso (DF a_a44D) (DF a_a45f) (Proxy a_a44D) (Proxy a_a45f)
    unDF = (iso (\ (MkDF x_a45g) -> x_a45g)) MkDF
    {-# INLINE unDF #-}
```

This won't typecheck:

```
/home/rgscott/Documents/Hacking/Haskell/Foo.hs:13:3: error:
    • Couldn't match type ‘k’ with ‘*’
      ‘k’ is a rigid type variable bound by
        the type signature for:
          unDF :: forall k k1 (a :: k) (a2 :: k1).
                  Iso (DF a) (DF a2) (Proxy a) (Proxy a2)
        at /home/rgscott/Documents/Hacking/Haskell/Foo.hs:13:3-18
      Expected type: p (Proxy a) (f (Proxy a2)) -> p (DF a) (f (DF a2))
        Actual type: p (Proxy a0) (f (Proxy a1)) -> p (DF a0) (f (DF a1))
    • In the expression: (iso (\ (MkDF x_a3YS) -> x_a3YS)) MkDF
      In an equation for ‘unDF’:
          unDF = (iso (\ (MkDF x_a3YS) -> x_a3YS)) MkDF
    • Relevant bindings include
        unDF :: p (Proxy a) (f (Proxy a2)) -> p (DF a) (f (DF a2))
          (bound at /home/rgscott/Documents/Hacking/Haskell/Foo.hs:13:3)
   |
13 | $(makeLenses 'MkDF)
   |   ^^^^^^^^^^^^^^^^
```

This is because `DF a` (i.e., `DF (a :: k)`) is too general for the data instance, which requires `DF (a :: Type)`. You'd need to generate this code instead:

```hs
    unDF ::
      forall (a_a2sl :: Type) (a_a3YQ :: Type).
      Iso (DF (a_a2sl :: Type)) (DF (a_a3YQ :: Type)) (Proxy a_a2sl) (Proxy a_a3YQ)
    unDF = (iso (\ (MkDF x_a3YR) -> x_a3YR)) MkDF
    {-# INLINE unDF #-}
```

That being said, I do see your point about not needing `KindSignatures` everywhere, however. Perhaps it would suffice to strip off monomorphic kinds as long as we are dealing with non-data family instances. After all, if you are dealing with data families, then you almost assuredly have `TypeFamilies` enabled, and `TypeFamilies` implies `KindSignatures`. (This is the approach that [libraries like `deriving-compat` take](https://github.com/haskell-compat/deriving-compat/blob/8d4403db4cb93a6c971f4ebc01e5cc601de257f6/src/Data/Deriving/Internal.hs#L389-L395).)

RyanGlScott:
I've pushed some changes that alter `datatypeTypeKinded` to only preserve kinds that contain kind variables (as long as it is not a data family instance), much in the spirit of well-typed/optics#313. Although this PR still contains breaking changes, they are now far less likely to occur in practice. (`wreq` will now compile with this version of `lens`, at the very least.)

RyanGlScott:
Now that `th-abstraction-0.4.1.0` has been released with a fix for glguy/th-abstraction#84, the last remaining obstacle in the way of this PR has been unblocked. If there are no further review comments, I'll likely land this some time in the coming week or so.

PR #948, opened 2020-12-02
Allow bytestring-0.11

Bodigrim:


phadej (approved):

RyanGlScott:
Thanks!

For future reference, how were you able to test this? The last time I tried building libraries with `bytestring-0.11`, I ran into several issues with boot libraries failing to compile. I'm interested in figuring this out since there is another restrictive upper version bound on `bytestring` in this repo that I'd like to lift:

https://github.com/ekmett/lens/blob/c83fc0e7a731db322de450cf981814ca7f0228ca/examples/lens-examples.cabal#L43

But I'd hate to just lift the bounds without having tested it myself first.

phadej:
I expect `bytestring` maintainer to know what they are doing, when they say upgrading is fine.

RyanGlScott:
To be clear, I trust @Bodigrim's judgment too! I'm mainly interested in learning how to do this so that I don't have to bug other people to figure out if my code is `bytestring-0.11`-compliant or not :)

phadej:
I ended up building my own local repository (`url: file+noindex:///....`) with patched packages, with `text`, `hashable`, `attoparsec` patches applied. It's cumbersome (if you really want to test pre GHC-8.0 - no pattern synonym case).

phadej:
That said, We can have a policy that if dependency relaxation is not achievable without jumping through the hoops, we'll leave PRs / issues open until there are install plans.

RyanGlScott:
> I ended up building my own local repository (url: file+noindex:///....) with patched packages, with e.g. text, hashable, attoparsec patches applied.

Ah, I was afraid that that might be the case. This is why I procrastinated on doing `bytestring`-related maintenance until there was a GHC (pre-)release that bundled `bytestring-0.11`. Speaking of which, is GHC 9.0 slated to include `bytestring-0.11`? If so, I might hold off on the task in https://github.com/ekmett/lens/pull/948#issuecomment-737173423 until that is out.

phadej:
>  Speaking of which, is GHC 9.0.1 slated to include `bytestring-0.11`?

As far as I know, *no*, GHC-9.0.1 will ship `bytestring-0.11.x.y`. But I also assumed GHC-9.0 will be out in September, so don't trust me on that.

RyanGlScott:
> As far as I know, no, GHC-9.0.1 will ship `bytestring-0.11.x.y`.

I see. In that case, it's likely that I'll need to rebuild boot libraries anyway. The thing that tripped me up was actually a different library than what you mentioned: `Cabal`.

```
[ 17 of 242] Compiling Distribution.Utils.Structured ( Distribution/Utils/Structured.hs, dist/build/Distribution/Utils/Structured.o, dist/build/Distribution/Utils/Structured.dyn_o )

Distribution/Utils/Structured.hs:90:1: error:
    Could not load module ‘Data.ByteString.Lazy.Builder’
    It is a member of the hidden package ‘bytestring-0.10.10.0’.
    Perhaps you need to add ‘bytestring’ to the build-depends in your .cabal file.
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
90 | import qualified Data.ByteString.Lazy.Builder as Builder
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Do you know if `Cabal` has been patched upstream to support `bytestring-0.11`? I noticed that the `master` branch of `Cabal` [still has `bytestring < 0.11` bounds](https://github.com/haskell/cabal/blob/5905a4a004052a22f362c2fa1544bf71fe574c69/Cabal/Cabal.cabal#L42), but perhaps I'm looking in the wrong place.

phadej:
Yes, `Cabal` is patched. `Cabal-3.4` will build with `bytestring-0.11`.

But the bounds are not relaxed as e.g. there is no `text` release. So there is a chain.

Bodigrim:
Two month ago the idea was that GHC 9.0 goes ahead with `bytestring-0.10`, so that its release is not delayed by bumps to core libraries. Not sure how much this rationale still holds; seems that the ecosystem is slowly catching up.

PR #950, opened 2020-12-07
Add _Cons and _Snoc instances for (Dual s) (Dual t) a b

Taneb:
This allows treating a cons-list as a snoc-list in a reasonable way and vice versa

One use case for this is Dual [a] can be used as a snoc list in an efficient way, if a snoc list is more "semantically" correct.

### DIFF
--- a/src/Control/Lens/Cons.hs
+++ b/src/Control/Lens/Cons.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE UndecidableInstances #-}
 #ifdef TRUSTWORTHY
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Is it commonplace to interpret `Dual [a]` as a snoc list? I'm accustomed to thinking of `Dual` in the "dual monoid" sense, but these instances seem to be giving `Dual` an additional meaning, in the sense that it treats `Cons` and `Snoc` as dual operations. Perhaps this is a sensible thing to do, but I can't recall having seen something like this before.

Taneb:
I don't know if it's used often, but I reached for it today, thinking of its `Monoid` instance

phadej:
I think this is on edge of being too ad-hoc, and I'm not sure on which side. `_Empty`, `_Cons`, `Monoid` aren't "officially" related, but (I guess) are true for all instances (of `Cons`). If we had `Singleton` we could tie this all together *formally*, but otherwise it is a collection of names which *accidentally* make sense together.

PR #952, opened 2020-12-16
hlint.yaml: Remove fixities that hlint already knows about

yairchu:
hlint is aware of those fixities since https://github.com/ndmitchell/hlint/commit/d3d03781b89a5467c7e68560c7649924e5c212e7
which went into hlint's v2.1 release (now at v3.2.3, this was almost 3 years ago)

removing those makes it easier to keep track of which operators hlint doesn't know about.


RyanGlScott:
Huh, I had no idea that `hlint` was aware of this. Thanks!

PR #957, opened 2020-12-26
Complete doctest-setup blocks

phadej:


examples/Aeson.hs
```diff
@@ -13,12 +13,17 @@ import Control.Lens
...
+-- >>> import Control.Lens
+-- >>> import Data.ByteString.Lazy
+
 -- |
 -- >>> review aeson 5
 -- "5"
 -- >>> [1,2,3]^.re aeson
 -- "[1,2,3]"
--- >>> aeson.both +~ 2 $ (2,3)^.re aeson
+-- >>> let intPair = simple :: Iso' (Int,Int) (Int, Int)
```
RyanGlScott, line 25: I'm surprised that changes like this are required. Can you explain what motivated this?

RyanGlScott (commented):
Just to be clear, is this an effort to make `lens`' doctests work with `-fobject-code`, in the same spirit as ekmett/comonad#55?

phadej:
> Just to be clear, is this an effort to make `lens`' doctests work with `-fobject-code`, in the same spirit as [ekmett/comonad#55](https://github.com/ekmett/comonad/pull/55)?

Yes. My "-fobject-code" like runner is able to run `lens` doctests with these changes. This is an effort towards `build-type: Simple` lens.

examples/Aeson.hs
phadej: it doesn't work. Without we get

```
          <interactive>:41:1: error:
              • Ambiguous type variable ‘r0’ arising from a use of ‘aeson’
                prevents the constraint ‘(FromJSON
                                            (r0 Integer Integer))’ from being solved.
                Probable fix: use a type annotation to specify what ‘r0’ should be.
                These potential instances exist:
                  instance (FromJSON a, FromJSON b) => FromJSON (Either a b)
                    -- Defined in ‘aeson-1.5.4.1:Data.Aeson.Types.FromJSON’
                  instance forall k a (b :: k). FromJSON a => FromJSON (Const a b)
                    -- Defined in ‘aeson-1.5.4.1:Data.Aeson.Types.FromJSON’
                  instance (FromJSON a, FromJSON b) => FromJSON (a, b)
                    -- Defined in ‘aeson-1.5.4.1:Data.Aeson.Types.FromJSON’
                  ...plus 13 others
                  ...plus 7 instances involving out-of-scope types
                  (use -fprint-potential-instances to see them all)
              • In the first argument of ‘(.)’, namely ‘aeson’
                In the first argument of ‘(+~)’, namely ‘aeson . both’
                In the expression: aeson . both +~ 2
          
          <interactive>:41:7: error:
              • Ambiguous type variable ‘r0’ arising from a use of ‘both’
                prevents the constraint ‘(Data.Bitraversable.Bitraversable
                                            r0)’ from being solved.
                Probable fix: use a type annotation to specify what ‘r0’ should be.
                These potential instances exist:
                  instance Data.Bitraversable.Bitraversable Either
                    -- Defined in ‘Data.Bitraversable’
                  instance Data.Bitraversable.Bitraversable (,)
                    -- Defined in ‘Data.Bitraversable’
                  instance Data.Bitraversable.Bitraversable ((,,) x)
                    -- Defined in ‘Data.Bitraversable’
                  ...plus four others
                  ...plus 17 instances involving out-of-scope types
                  (use -fprint-potential-instances to see them all)
              • In the second argument of ‘(.)’, namely ‘both’
                In the first argument of ‘(+~)’, namely ‘aeson . both’
                In the expression: aeson . both +~ 2
```

I.e. `both +~ 2` part doesn't pin which type to parse JSON to modify it then.

Note: these doctests aren't run atm, that is why we didn't notice.

phadej: or simply put, the code does `show . (+1) . read $ "2"`, which works (in GHCi) due defaulting. Pairs and `both` are too much.

RyanGlScott: > Note: these doctests aren't run atm, that is why we didn't notice.

Ah, that's a pretty good reason for changing it then :)

PR #958, opened 2020-12-27
Use cabal-docspec

phadej:


RyanGlScott:
Ping me when this is ready for a review.

phadej:
The missing part is cleaning up the `doctests`. CI works (and is as simple to use locally).

If you could look through https://github.com/phadej/cabal-extras/blob/master/cabal-docspec/MANUAL.md, if you have some questions about `cabal-docspec` based on that.

RyanGlScott:
Aha, so _that's_ where the source code for `cabal-docspec` lives! When I have more time, I'd definitely like to take a closer look at it, but perhaps I should wait until it's in a more polished state first?

phadej:
@RyanGlScott it's quite polished. Manual is scarse, but functionality is there (to run `lens`, `optics`, `base`, `comonad` and few others doctests).

phadej:
This is ready for review now. Manual (https://github.com/phadej/cabal-extras/blob/master/cabal-docspec/MANUAL.md) is updated with `lens` as an example.

If this look ok, then I'll make a release of `cabal-docspec` (i.e. host it from GitHub releases, and not from my site), and update merge `cabal-docspec` support to `haskell-ci`.

phadej:
Because we change from `build-type: Simple` to `build-type: Custom`
- we don't need (cannot have) `custom-setup` stanza
- `Setup.lhs` is not required, but for backward compat we make it a default one
- `Warning.hs` is also not needed (was used by `Setup.lhs`).
- We can have `test-output-direct: True` (default) with `haskell-ci`
- flag `test-doctests` is meaningless

We however need to preserve `test-suite doctests` to have a way to pull `simple-reflect` into the install plan. But e.g. in `comonad` (and probably other packages') case that test-suite can be removed.

lens.cabal
```diff
@@ -412,40 +401,17 @@ test-suite hunit
       test-framework       >= 0.6,
       test-framework-hunit >= 0.2
 
--- Verify the results of the examples
+-- We need this dummy test-suite to add simple-reflect to the install plan
```
RyanGlScott, line 404: The fact that you need a dummy test suite for dependency purposes is interesting. Does it somehow pick up on the fact that this test suite is related to doctests?

lens.cabal
```diff
@@ -348,6 +335,8 @@ library
 
   default-language: Haskell2010
 
+  x-docspec-options: --extra-package=simple-reflect --extra-package=deepseq
```
RyanGlScott, line 338: I wonder if it would be cleaner to be able to say `--extra-packages="simple-reflect deepseq"`. I could imagine that having to duplicate `--extra-package=` for each dependency would become annoying.

.github/workflows/haskell-ci.yml
```diff
@@ -141,6 +141,16 @@ jobs:
...
+          cabal-plan --version
+      - name: install cabal-docspec
+        run: |
+          mkdir -p $HOME/.cabal/bin
+          curl -sL https://oleg.fi/cabal-docspec-0.0.0.20201228.xz > cabal-docspec.xz
+          echo 'edf15d3a0b46fc57fb1bdb0ae5e87abdc5c3b7dfff8655a958ef038d4c2f17ab  cabal-docspec.xz' | sha256sum -c -
+          xz -d < cabal-docspec.xz > $HOME/.cabal/bin/cabal-docspec
+          rm -f cabal-docspec.xz
+          chmod a+x $HOME/.cabal/bin/cabal-docspec
+          cabal-docspec --version
```
RyanGlScott, line 153: Admittedly, I was under the impression that `cabal-docspec` was something that you ran as a dependency in the test suite, not as an external executable. What advantages does this provide over simply running the `doctest` executable? (I apologize if this is a silly question—I still don't fully understand the internal workings of `cabal-docspec`.)

.github/workflows/haskell-ci.yml
```diff
@@ -200,10 +210,14 @@ jobs:
...
         run: |
-          $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all
+          $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --write-ghc-environment-files=always
       - name: tests
         run: |
-          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all
+          $CABAL v2-test $ARG_COMPILER $ARG_TESTS $ARG_BENCH all --test-show-details=direct
+      - name: docspec
+        run: |
+          $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all
```
RyanGlScott, line 219: Isn't this `$CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH all` invocation redundant? You already run it once above.

lens.cabal
```diff
@@ -412,40 +401,17 @@ test-suite hunit
...
 
--- Verify the results of the examples
+-- We need this dummy test-suite to add simple-reflect to the install plan
 test-suite doctests
   type:              exitcode-stdio-1.0
   main-is:           doctests.hs
-  ghc-options:       -Wall -threaded
   hs-source-dirs:    tests
   default-language:  Haskell2010
-  x-doctest-options: --fast
```
RyanGlScott, line 422: Does `--fast` have an equivalent in `cabal-docspec`? I remember this being a pain point for `doctest` at one point.

lens.cabal
```diff
@@ -412,40 +401,17 @@ test-suite hunit
...
-      parallel,
-      semigroups     >= 0.9,
-      simple-reflect >= 0.3.1,
-      text,
-      unordered-containers,
-      vector
+  build-depends:
+    base,
+    deepseq,
+    simple-reflect >= 0.3.1
```
RyanGlScott, line 414: Oof, the fact that you have to list dependencies on `deepseq` and `simple-reflect` _twice_ in the same `.cabal` file is rather painful. Is there a way that we can make it so that they only have to be listed once?

RyanGlScott (commented):
That was fast! I still have some questions about the design of `cabal-docspec` before we proceed further.

lens.cabal
phadej: > Does it somehow pick up on the fact that this test suite is related to doctests?

No. It just looks around in the install plan for `--extra-package`s you specify. If `cabal.project` `extra-packages` worked, we could just use that. (They might in 3.4, but I'm not sure).

lens.cabal
phadej: due how parsing in `.cabal` files work it would need to be
```
"--extra-package=simple-reflect deepseq"
```

Which I'm sure people won't get right. (Though it's consistent with `--ghc-options` etc.)

An alterntive is to make
```
x-docspec-extra-packages: simple-reflect deepseq
```
possible, but for now I like the open-word of `x-doctest-options`. (As said, I expect `lens` to be the only one needing two arguments!)

.github/workflows/haskell-ci.yml
phadej: `doctest` executable have to be compiled with GHC you want to test with. (So you cannot *simply* download it, problem that `ghc-mod`, `hie` etc have, AFAIK). `doctest` uses `ghc` library to both extract the docstrings as well as running own `ghci` (it doesn't call to external processes).

On the other hand, `cabal-docspec` just works with any GHC. A bit like `ghcid` doesn't really care about GHC version you use.

Another difference: because `cabal-docspec` isn't limited by "have to be fast to compile" (because we can download it), it can be fancier. Important fancy feature is that it can self configure. In short, the (kind of) logic we had in `cabal-doctest` custom build script is now moved into `cabal-docspec` executable. (We can learn enough from `plan.json` `cabal-install` generates).

.github/workflows/haskell-ci.yml
phadej: Unfortunately `v2-test` can mess up the local package db. Not in `lens` case, but I don't know how to know for sure it isn't required.

`cabal` does spend some seconds thinking (it doesn't end up building anything). And whole step takes just 40secs. So I rather not make it more complicated/fragile.

lens.cabal
phadej: That is discussed in the manual.

https://github.com/phadej/cabal-extras/blob/master/cabal-docspec/MANUAL.md#a-note-on-performance (looks like I miss a word there).

> Because cabal-docspec uses compiled library, calling :reload: after each group doesn't cause performance problems, reloading is fast. For that reason, cabal-docspec doesn't have --fast variant, it is not needed.

lens.cabal
phadej: deepseq is not strictly needed (because it's transitive dependency of about everything). But I think that repeating here is better, then having some non-obvious rule of thumbs.

Note: **if** in the bright future, `cabal-install` learns about `cabal-docspec`, it could learn `x-docspec-extra-packages` (or similar), which then would be `x-docspec-extra-packages`.

i have to think about `x-docspec-extra-packages`. Maybe I'll add it tomorrow if I still think it's not a bad idea.

.github/workflows/haskell-ci.yml
phadej: One concrete example of problems with `.ghc.environment` files and `doctest` is visibility of packages.

`haskell-ci` has a perl-hack to allow filtering packages, e.g. to remove `base-compat` from the environment files (which contains transitive dependencies) for packages which use `base-compat-batteries`.

`cabal-docspec` can be more precise (haskell vs. bash+perl :) )

lens.cabal
phadej: The `docspec` step on GHA takes 40seconds. I think it is reasonable. (`doctest` with `--fast` took about the same time).

phadej:
I pushed a commit using `cabal-docspec` with ` x-docspec-extra-packages` support. Does it look better?

lens.cabal
```diff
@@ -348,6 +335,8 @@ library
 
   default-language: Haskell2010
 
+  x-docspec-extra-packages: simple-reflect deepseq
```
phadej, line 338: currently it's commaless list. But now as I see it, it looks weird. Should we have commas?

.github/workflows/haskell-ci.yml
RyanGlScott: > Another difference: because `cabal-docspec` isn't limited by "have to be fast to compile" (because we can download it), it can be fancier.

To clarify, are you envisioning making binary distributions of `cabal-docspec` available in the future so that one can avoid needing to build it from source?

> Important fancy feature is that it can self configure. In short, the (kind of) logic we had in `cabal-doctest` custom build script is now moved into `cabal-docspec` executable. (We can learn enough from `plan.json` `cabal-install` generates).

Interesting. When does `plan.json` get generated in this script?

> One concrete example of problems with `.ghc.environment` files and `doctest` is visibility of packages.
> 
> `haskell-ci` has a perl-hack to allow filtering packages, e.g. to remove `base-compat` from the environment files (which contains transitive dependencies) for packages which use `base-compat-batteries`.
> 
> `cabal-docspec` can be more precise (haskell vs. bash+perl :) )

Ah, I had no idea that this hack was necessary. Indeed, that's a very good reason to use something more sophisticated than a bare `doctest` invocation.

lens.cabal
RyanGlScott: I'm fine with the way the current formatting.

lens.cabal
RyanGlScott: OK. I assume this means that this will only work if `--enable-tests` is set? In other words, `docspec: True` and `tests: False` are incompatible?

.github/workflows/haskell-ci.yml
RyanGlScott: I see. But shouldn't `$ARG_TESTS` be `--enable-tests` instead? I fail to see how this could possibly work unless the dependencies from the test suite are included in the build plan.

lens.cabal
RyanGlScott: That's what I get for not reading the freaking manual. Thanks for pointing that out!

lens.cabal
phadej: For `lens`, yes.

.github/workflows/haskell-ci.yml
phadej: > To clarify, are you envisioning making binary distributions of cabal-docspec available in the future so that one can avoid needing to build it from source?

Yes.

> Interesting. When does plan.json get generated in this script?

It's a side-effect of doing about anything with `cabal-install`.

Even `cabal build --dry` is enough, though then some units (e.g. local) will be missing.

.github/workflows/haskell-ci.yml
phadej: I think you are thinking too much into `lens` example.

The problematic situation I try to avoid here are related to `all` meaning different things for `v2-test` and `v2-build`.

Consider a multi-package project. If some packages don't have `test-suite`s, their libraries won't be built by `v2-test`. For example, in `lens` case `cabal v2-test all --enable-tests` *won't* build `lens-examples`. I think that it's unsafe to assume that `lens-example` will still be registered in a local (project) package db.

However, we build with `v2-build all --enable-tests` just before `v2-test`, so things *seems* to work.

Yet, I don't know (even I should), what `v2-test` and `v2-build` do differently, as `v2-build` after `v2-test` is not a no-op:

```
In order, the following will be built (use -v for more details):
 - lens-5 (lib) (configuration changed)
 - lens-5 (test:doctests) (configuration changed)
 - lens-properties-4.11.1 (lib) (dependency rebuilt)
 - lens-examples-0.1 (exe:lens-pong) (dependency rebuilt)
 - lens-examples-0.1 (lib) (dependency rebuilt)
 - lens-5 (test:templates) (configuration changed)
 - lens-5 (test:properties) (configuration changed)
 - lens-5 (test:hunit) (configuration changed)
Configuring library for lens-5..
Preprocessing library for lens-5..
Building library for lens-5..
Configuring test suite 'doctests' for lens-5..
Preprocessing test suite 'doctests' for lens-5..
Building test suite 'doctests' for lens-5..
Preprocessing library for lens-properties-4.11.1..
Building library for lens-properties-4.11.1..
Preprocessing executable 'lens-pong' for lens-examples-0.1..
Building executable 'lens-pong' for lens-examples-0.1..
Preprocessing library for lens-examples-0.1..
Building library for lens-examples-0.1..
Configuring test suite 'templates' for lens-5..
Warning: The package has an extraneous version range for a dependency on an
internal library: lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0
&& ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5.
This version range includes the current package but isn't needed as the
current package's library will always be used.
Preprocessing test suite 'templates' for lens-5..
Building test suite 'templates' for lens-5..
Configuring test suite 'properties' for lens-5..
Warning: The package has an extraneous version range for a dependency on an
internal library: lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0
&& ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5.
This version range includes the current package but isn't needed as the
current package's library will always be used.
Preprocessing test suite 'properties' for lens-5..
Building test suite 'properties' for lens-5..
Configuring test suite 'hunit' for lens-5..
Warning: The package has an extraneous version range for a dependency on an
internal library: lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0
&& ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5, lens >=0 && ==5.
This version range includes the current package but isn't needed as the
current package's library will always be used.
Preprocessing test suite 'hunit' for lens-5..
Building test suite 'hunit' for lens-5..
```

It kind of does a lot, but actually does nothing.

So for now, I'd error on the safe side and have an extra `v2-build`.

lens.cabal
phadej: No worries. Not all your questions have answers there (yet).

lens.cabal
RyanGlScott: OK. This all feels a bit delicate to me, but at the same time, I can't think of a better way to do this with the design considerations involved. At the very least, this use case is documented in the manual, so I'm content with that.

.github/workflows/haskell-ci.yml
RyanGlScott: I think you misinterpreted my comment, as I wasn't suggesting to use `v2-test`, but rather using `cabal v2-build --enable-tests` instead of `cabal v2-build --enable-tests`. But in light of https://github.com/ekmett/lens/pull/958/files#r549543219, perhaps the use of `--enable-tests` isn't a good idea.

.github/workflows/haskell-ci.yml
RyanGlScott: Huh, today I learned that `cabal-install` [automatically generates a `plan.json`](https://cabal.readthedocs.io/en/3.4/nix-local-build.html?highlight=plan.json#caching) file that is located in `dist-newstyle`. Neat!

.github/workflows/haskell-ci.yml
phadej: instead of `$ARG_TESTS`?

Unconditional `--enable-tests` may be very bad, e.g. when tests are disabled for genuine reasons, like "a testing lib doesn't work with GHC-x.y" and aren't needed for adding dependencies for `docspec`.

.github/workflows/haskell-ci.yml
phadej: Yes, it's what makes `cabal-plan` able to do all the things it does.

lens.cabal
phadej: I tried locally, `extra-packages: simple-reflect` seems to work with `cabal-install-3.4`.

So in near future (once it's officially released &dagger;, and support in `haskell-ci` is added, ...) this can be cleaned up a bit. I don't want go there yet though.

&dagger; because adding it to `cabal.project` will affect contributors, which may not use release candidates.

phadej:
Is there some answered questions still, or is Windows support the only blocking thing?

RyanGlScott:
Sorry! There's nothing blocking this, I just hadn't had a chance to respond in a while. Overall, this looks nice, and I'm definitely looking forward to being able to drop custom `Setup` scripts in the near future.

If you'd like to go ahead with cutting a release of `cabal-docspec` and moving forward with it, that's fine with me. Even if there are still undiscovered issues, the risk involved here is pretty low, given that we're only changing the behavior of a single test suite.

lens.cabal
```diff
@@ -348,6 +335,10 @@ library
 
   default-language: Haskell2010
 
+  -- future proof, whether the field will be comma separated or not.
+  x-docspec-extra-packages: simple-reflect
+  x-docspec-extra-packages: deepseq
```
RyanGlScott, line 340: I'm a bit confused—`deepseq` is listed as an `x-docspec-extra-package` here, but it's not listed in the `build-depends` stanza of the `doctests` test suite. What is the reason for this discrepancy?

phadej: You wanted less duplication :)

RyanGlScott: To be clear, my dream is not to have any duplication at all, but if I understand https://github.com/ekmett/lens/pull/958#discussion_r549519356 correctly, that isn't feasible unless `cabal-install` gains some additional functionality. Given that, I reluctantly accept the fact that we need to duplicate some information for the time being.

But the state that this `.cabal` file is in now is an uncomfortable medium between no duplication and duplicating every `x-docspec-extra-package`, as we duplicate one of these packages (`simple-reflect`) but not the other (`deepseq`). This is even more mysterious to me, and when I inevitably revisit this code months later, I'm probably not going to remember the subtleties that allow this to work. I would request that we either:

* Just duplicate both `simple-reflect` and `deepseq`, to make it more visually obvious that these parts of the `.cabal` file are meant to be in sync, or
* Leave a comment somewhere explaining why `deepseq` is omitted.

phadej: I duplicated both, and added a comment why we need `doctests`. And what needs to happen, so it can be removed.

PR #960, opened 2021-01-01
Remove unneccessary #ifndef HLINTs

yairchu:
Those were probably needed when added 8 years ago at 873e13904a2c4d48022467d56150405b59cf710d,
but testing now, they no longer seem necessary,
either hlint has evolved or the .hlint.yaml has?

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,4 +1,4 @@
-- arguments: [--cpp-define=HLINT, --cpp-ansi, --cpp-include=include]
+- arguments: [--cpp-ansi, --cpp-include=include]
 
 - ignore: {name: Reduce duplication}
 - ignore: {name: Redundant lambda}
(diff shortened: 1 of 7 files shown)


RyanGlScott:
Good catch!

Now that the `HLINT` directive is no longer used, can you remove the `--cpp-define=HLINT` bit from the `.hlint.yaml` files in this repo? There's in the following places:

https://github.com/ekmett/lens/blob/ec7357fc2a94875aabd97e9ed64b6b8b542402c6/.hlint.yaml#L1

https://github.com/ekmett/lens/blob/ec7357fc2a94875aabd97e9ed64b6b8b542402c6/examples/.hlint.yaml#L1

https://github.com/ekmett/lens/blob/ec7357fc2a94875aabd97e9ed64b6b8b542402c6/lens-properties/.hlint.yaml#L1

yairchu:
> Now that the `HLINT` directive is no longer used, can you remove the `--cpp-define=HLINT` bit from the `.hlint.yaml` files in this repo?

Sure, done!

PR #962, opened 2021-01-06
Use HashMap.alterF for HashMap's At instance

Taneb:


src/Control/Lens/At.hs
```diff
@@ -469,10 +469,14 @@ instance Ord k => At (Map k a) where
   {-# INLINE at #-}
 
 instance (Eq k, Hashable k) => At (HashMap k a) where
+#if MIN_VERSION_unordered_containers(0,2,10)
```
phadej, line 472: Just bump the lower bound on `unordered-containers`.

phadej: For reference, https://matrix.hackage.haskell.org/#/package/unordered-containers `-0.2.10.0` supports GHC-7.8, which is the oldes GHC next `lens` release would support. So there is little reason to extend support to older `unordered-containers`.

src/Control/Lens/At.hs
```diff
@@ -469,10 +469,7 @@ instance Ord k => At (Map k a) where
   {-# INLINE at #-}
 
 instance (Eq k, Hashable k) => At (HashMap k a) where
-  at k f m = f mv <&> \r -> case r of
-    Nothing -> maybe m (const (HashMap.delete k m)) mv
-    Just v' -> HashMap.insert k v' m
-    where mv = HashMap.lookup k m
+  at k f = HashMap.alterF f k
```
phadej, line 472: Note to self, I'll try after this is merged.

I think we can eta-reduce `f`, but I'm not sure about `k` (simplified subsumption in GHC-9.0).

phadej: Ah, we cannot. They are flipped. Never mind.

PR #963, opened 2021-01-08
Use Data.Set.alterF

treeowl:


treeowl:
@RyanGlScott, do you want me to bump the lower bound on `containers` instead of using CPP?

phadej:
>  do you want me to bump the lower bound on containers instead of using CPP?

We shouldn't. It is less pain to everyone if we support `containers` versions bundled with GHCs we support.

phadej:
Is there `IntSet.alterF`?

treeowl:
I don't remember. I'll take a look.

treeowl:
@phadej Yes, there is an `alterF` for `IntSet`, but it's based entirely on API functions:

```haskell
alterF :: Functor f => (Bool -> f Bool) -> Key -> IntSet -> f IntSet
alterF f k s = fmap choose (f member_)
  where
    member_ = member k s

    (inserted, deleted)
      | member_   = (s         , delete k s)
      | otherwise = (insert k s, s         )

    choose True  = inserted
    choose False = deleted
```

So I guess it's a matter of taste whether to use it when available or not. We probably *shouldn't* use it for `at`, since that leads to an extra `fmap isJust`.

phadej:
I'd be in favor of having similar implementations for `IntSet` and `Set` **or** a proper note explaining the difference (and perfectly illustrating it with non contrived example).

I don't think that extra `fmap` is noticeable for e.g. `set` implementation.

treeowl:
`Set` and `IntSet` are quite different in this regard. Going into a `Set` two independent times means performing (possibly expensive) comparisons along the way twice, so we avoid that. For `IntSet`, all the branches are cheap. There are all sorts of trade-offs; `containers` aims to be not-awful in the case of expensive `compare` and expensive `fmap`, while also trying to be pretty fast in easier cases.

phadej:
Expensive `fmap` + `at @(Set _)` is a speculation. (If someone cares, they should use `contains`). Please make implementation uniform. Or provide method with `at` signature in `containers`.

I won't remember this discussion few months from now. If you won't change it, the future me will.

treeowl:
I do not understand your last remark. I'm happy to comment the source in detail on the reasoning behind the implementation.

phadej:
> I do not understand your last remark. I'm happy to comment the source in detail on the reasoning behind the implementation.

That is fine too.

src/Control/Lens/At.hs
```diff
@@ -473,17 +496,32 @@ instance (Eq k, Hashable k) => At (HashMap k a) where
...
+  -- Unlike a `Set`, we don't have to worry about expensive
+  -- comparisons from descending multiple times into an
+  -- `IntSet`. We are careful to share the results of insertion
+  -- or deletion across multiple positions in the `Functor`.
+  at k f s = fmap choose (f (guard member_))
+    where
+      member_ = IntSet.member k s
+
+      (inserted, deleted)
+        | member_   = (s         , IntSet.delete k s)
```
phadej, line 509: aligning is off, probably better to eat whitespace.

src/Control/Lens/At.hs
```diff
@@ -473,17 +496,32 @@ instance (Eq k, Hashable k) => At (HashMap k a) where
   {-# INLINE at #-}
 
 instance At IntSet where
-  at k f m = f mv <&> \r -> case r of
-    Nothing -> maybe m (const (IntSet.delete k m)) mv
-    Just () -> IntSet.insert k m
-    where mv = if IntSet.member k m then Just () else Nothing
+  -- This is a copy of the implementation of `IntSet.alterF`.
```
phadej, line 499: *modified* copy?

phadej:
Looks good. Do you mind adding the changelog entry, then this can be merged.

treeowl:
@phadej, I removed useless reinsertion in the implementation of `ix` for `IntSet`. For `Set` and `HashSet`, things are a bit more subtle; should `ix` be able to replace a key with a different one that compares `==` to it? That's what it currently does.

phadej:
Re `ix`, it's behavior should be `ix i = at i . _Just`. That doesn't re-insert key, does it?

treeowl:
I don't believe so.

phadej:
Then `ix` should reinsert either (And I think it's better that way too).

treeowl:
@phadej, all good now?

treeowl:
Thanks. Now I should probably add an `adjustA` to `containers` and `unordered-containers` to deal with map `Ix` better. Bleh.

PR #965, opened 2021-01-12
Eager seq ix

treeowl:
* Make `ix` look up eagerly for `Seq`.
* Make `ix` use `!?` for vectors, which [should eventually](https://github.com/haskell/vector/pull/347) make those lookups eager too.

### DIFF
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -365,30 +365,30 @@ instance (IArray UArray e, Ix i) => Ixed (UArray i e) where
 
 type instance IxValue (Vector.Vector a) = a
 instance Ixed (Vector.Vector a) where
-  ix i f v
-    | 0 <= i && i < Vector.length v = f (v Vector.! i) <&> \a -> v Vector.// [(i, a)]
-    | otherwise                     = pure v
+  ix i f v = case v Vector.!? i of
+    Just x -> f x <&> \x' -> v Vector.// [(i, x')]
+    Nothing -> pure v
   {-# INLINE ix #-}
 
 type instance IxValue (Prim.Vector a) = a
 instance Prim a => Ixed (Prim.Vector a) where
-  ix i f v
-    | 0 <= i && i < Prim.length v = f (v Prim.! i) <&> \a -> v Prim.// [(i, a)]
-    | otherwise                   = pure v
+  ix i f v = case v Prim.!? i of
+    Just x -> f x <&> \x' -> v Prim.// [(i, x')]
+    Nothing -> pure v
   {-# INLINE ix #-}
 
 type instance IxValue (Storable.Vector a) = a
 instance Storable a => Ixed (Storable.Vector a) where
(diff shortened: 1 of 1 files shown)


src/Control/Lens/At.hs
```diff
@@ -286,9 +286,15 @@ instance Ixed (Tree a) where
 
 type instance IxValue (Seq a) = a
 instance Ixed (Seq a) where
+#if MIN_VERSION_containers(0,5,8)
+  ix i f m = case Seq.lookup i m of
+    Just x -> f x <&> \x' -> Seq.update i x' m
+    Nothing -> pure m
+#else
   ix i f m
     | 0 <= i && i < Seq.length m = f (Seq.index m i) <&> \a -> Seq.update i a m
```
phadej, line 295: Is here impossible to `seq` things to avoid thunk, Would

```
let x = Seq.index m i `seq` (f x <&> ...)
```

do anything?

phadej (commented):
Eager here means that we don't retain the original sequence/vector, i.e. `f x` is `f <element value>` and not `f <think to lookup element>`?

If my understanding is right, could you amend commit message(s) with that, so it would be clearer what these changes do.

src/Control/Lens/At.hs
treeowl: Using `seq` like that would force *too much*, forcing not only the lookup but also the element itself.

phadej: I see. as later then `0.5.8` `containers` are bundled with GHC-8.2, I don't think we need to do here anything. (If someone have problems due this lazyness, and they could upgrade GHC or/and containers)

phadej (approved):
Please add changelog entry. If you explain what this change means there (not just saying *eager seq ix*), then we don't need to repeat that in commit message.

src/Control/Lens/At.hs
treeowl: @phadej Do you mean we shouldn't worry about the leak for older `containers`? I agree. Trying to plug it in that context will be very inefficient when `f` is actually strict.

treeowl:
@phadej For `Data.Sequence`, lookup isn't `O(1)`, and this PR can make lookups happen when they're not actually necessary. I can write some fancy rewrite rules to take care of the case where `ix` is (obviously) used to set without inspecting. This involves a dance with non-linear patterns and carefully buried bottom values. But I'm kind of thinking all that gunk would fit better in `containers` than in `lens`.

RyanGlScott:
What is the status of this PR? If I'm reading correctly, this is very nearly done modulo documentation/changelog concerns. (I'm asking since I'm doing a roundup of recent PRs before an upcoming `lens` release.)

treeowl:
Haven't looked at this in ages. Can try to look today.

phadej:
ping @treeowl

treeowl:
Pong! I will try to look at this today.

PR #967, opened 2021-01-15
Add a note that gplate is not uniplate

### LINKED ISSUE #966: Control.Lens.rewriteM does not behave like Data.Generics.Uniplate.Data.rewriteM
Is the intention that Lens version of rewriteM should behave the same as Uniplate version?
The result from the Lens version still has places where the rewrite can be applied, whereas Uniplate version does all the rewrites.

If they are intended to be the same, I can make a proper bug report with an example.

phadej:
Closes #966 

@RyanGlScott do you think we need to add something here? For me it was obvious that `GHC.Generics` based approach cannot handle mutual recursive types (without very ugly tricks, in a way `gplate` would need). Yet it was a surprise to @augustss.

### DIFF
--- a/src/Control/Lens/Plated.hs
+++ b/src/Control/Lens/Plated.hs
@@ -118,6 +118,13 @@ import Data.Data.Lens
 import Data.Tree
 import GHC.Generics
 
+-- $setup
+-- >>> :set -XDeriveGeneric -XDeriveDataTypeable
+-- >>> import Control.Applicative
+-- >>> import Data.Data (Data, Typeable)
+-- >>> import GHC.Generics (Generic)
+-- >>> import Control.Lens
+
 -- | A 'Plated' type is one where we know how to extract its immediate self-similar children.
 --
 -- /Example 1/:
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Ultimately, I think @augustss is the only one who can say whether this is sufficient documentation. From my perspective, however, it may be helpful to include a small example that demonstrates where `gplate`'s treatment of recursive data types differs from `uniplate`. Perhaps a condensed version of the example in https://github.com/ekmett/lens/issues/966#issuecomment-760489484 would suffice.

phadej:
@RyanGlScott, I added a small example.

src/Control/Lens/Plated.hs
```diff
@@ -721,6 +728,41 @@ parts = partsOf plate
...
 
 -- | Implement 'plate' operation for a type using its 'Generic' instance.
+--
+-- Note: the behavior may be different than with 'uniplate' in some special cases.
+-- The 'gplate' doesn't look through other types in a group of mutually
+-- recusive types.
+--
+-- For example consider mutually recursive even and odd natural numbers:
+--
+-- >>> data Even = Z | E Odd deriving (Show, Generic, Typeable, Data); data Odd = O Even deriving (Show, Generic, Typeable, Data)
```
phadej, line 738: For some reason GHC-7.8 doesn't like multiple line definition.

src/Control/Lens/Plated.hs
```diff
@@ -721,6 +728,41 @@ parts = partsOf plate
 -------------------------------------------------------------------------------
 
 -- | Implement 'plate' operation for a type using its 'Generic' instance.
+--
+-- Note: the behavior may be different than with 'uniplate' in some special cases.
+-- The 'gplate' doesn't look through other types in a group of mutually
```
RyanGlScott, line 733:
```suggestion
-- 'gplate' doesn't look through other types in a group of mutually
```

src/Control/Lens/Plated.hs
```diff
@@ -721,6 +728,41 @@ parts = partsOf plate
...
+-- recusive types.
```
RyanGlScott, line 734:
```suggestion
-- recursive types.
```

src/Control/Lens/Plated.hs
```diff
@@ -721,6 +728,41 @@ parts = partsOf plate
...
+--
+-- For example consider mutually recursive even and odd natural numbers:
+--
+-- >>> data Even = Z | E Odd deriving (Show, Generic, Typeable, Data); data Odd = O Even deriving (Show, Generic, Typeable, Data)
+--
+-- Then 'Data' based 'uniplate' finds
```
RyanGlScott, line 740:
```suggestion
-- Then 'uniplate', which is based on `Data`, finds
```

RyanGlScott (commented):
Other than some minor grammar-related suggestions, LGTM.

Issue #969, opened 2021-01-31
Tests failed with vector 0.12.2.0


felixonmars:
```
Running 4 test suites...
Test suite doctests: RUNNING...
/build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:1334: failure in expression `ala Sum foldMap [1,2,3,4]'
expected: 10
 but got:
          ^
          <interactive>:1371:9: error:
              Ambiguous occurrence ‘foldMap’
              It could refer to
                 either ‘Foldable.foldMap’,
                        imported from ‘Data.Foldable’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:123:1-42
                     or ‘Vector.foldMap’,
                        imported from ‘Data.Vector’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:149:1-38
                     or ‘Unboxed.foldMap’,
                        imported from ‘Data.Vector.Unboxed’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:151:1-47
                     or ‘Storable.foldMap’,
                        imported from ‘Data.Vector.Storable’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:152:1-49
                     or ‘Prim.foldMap’,
                        imported from ‘Data.Vector.Primitive’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:150:1-46

/build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:1371: failure in expression `alaf Sum foldMap Prelude.length ["hello","world"]'
expected: 10
 but got:
          ^
          <interactive>:1379:10: error:
              Ambiguous occurrence ‘foldMap’
              It could refer to
                 either ‘Foldable.foldMap’,
                        imported from ‘Data.Foldable’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:123:1-42
                     or ‘Vector.foldMap’,
                        imported from ‘Data.Vector’ at /build/haskell-lens/src/lens-4.19.2/src/Control/Lens/Wrapped.hs:149:1-38
                     or ‘Unboxed.foldMap’,
[...]

PR #970, opened 2021-01-31
Always import from Data.Vector.* qualified

RyanGlScott:
This makes errors like the ones observed in #969 less likely to occur in the future.

### DIFF
--- a/src/Control/Lens/Internal/Iso.hs
+++ b/src/Control/Lens/Internal/Iso.hs
@@ -22,16 +22,19 @@ import Data.Profunctor
 import Data.Profunctor.Unsafe
 import Control.Lens.Internal.Coerce
 
-import Data.ByteString       as StrictB
-import Data.ByteString.Lazy  as LazyB
-import Data.List.NonEmpty    as NonEmpty
-import Data.Text             as StrictT
-import Data.Text.Lazy        as LazyT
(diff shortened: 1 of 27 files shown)


src/Control/Lens/Internal/Iso.hs
```diff
@@ -22,16 +22,19 @@ import Data.Profunctor
...
-import Data.ByteString.Lazy  as LazyB
-import Data.List.NonEmpty    as NonEmpty
-import Data.Text             as StrictT
-import Data.Text.Lazy        as LazyT
-import Data.Vector           as Vector
-import Data.Vector.Primitive as Prim
-import Data.Vector.Storable  as Storable
-import Data.Vector.Unboxed   as Unbox
-import Data.Sequence         as Seq
+import Data.ByteString                 as StrictB
```
phadej, line 25: FWIW, I'd import all of these qualified. But even without this is :+1:

RyanGlScott: True, there's even more suspicious imports besides the ones from `vector`. Do you think it's worth addressing all of them here, or would that be better suited for separate PRs?

RyanGlScott: I decided to push a series of additional commits that change other commonly-clashing imports to be qualified.

PR #974, opened 2021-03-02
Add 'beneath'

dminuoso:
This PR adds a new combinator `beneath`, a Fold version of `aside`.

In our code base, we wanted this often, and `aside` wasn't enough because the nested property was usually described by an (affine) fold like `someField . _Just`. I think wanting to filter/fold all objects based on the presence of a nested property, while keeping that property around, is not an exotic thing to do. `aside` exists for the same reason.

### DIFF
--- a/src/Control/Lens/Fold.hs
+++ b/src/Control/Lens/Fold.hs
@@ -65,6 +65,7 @@ module Control.Lens.Fold
   , backwards
   , repeated
   , replicated
+  , beneath
   , cycled
   , takingWhile
   , droppingWhile
@@ -278,6 +279,14 @@ replicated n0 f a = go n0 where
   go 0 = noEffect
   go n = m *> go (n - 1)
 {-# INLINE replicated #-}
+
+-- | Use a 'Fold' to work over part of a structure. This is a version of 'aside' that takes a Fold
+-- instead. The resulting 'Fold' has as many targets as the provided 'Fold'.
+--
+-- >>> [Just ('a', 10), Nothing, Just ('b', 20)] ^.. folded . beneath (_Just . _1)
+-- [(Just ('a',10),'a'),(Just ('b',20),'b')]
+beneath :: Fold s a -> Fold s (s, a);
+beneath fo f s = phantom $ foldrOf fo (\a' fa' -> f (s, a') *> fa') noEffect s
 
 -- | Transform a non-empty 'Fold' into a 'Fold1' that loops over its elements over and over.
 --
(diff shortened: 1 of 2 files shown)


src/Control/Lens/Fold.hs
```diff
@@ -279,6 +280,14 @@ replicated n0 f a = go n0 where
   go n = m *> go (n - 1)
 {-# INLINE replicated #-}
 
+-- | Use a 'Fold' to work over part of a structure. This is a version of 'aside' that takes a Fold
+-- instead. The resulting 'Fold' has as many targets as the provided 'Fold'.
+--
+-- >>> [Just ('a', 10), Nothing, Just ('b', 20)] ^.. folded . beneath (_Just . _1)
+-- [(Just ('a',10),'a'),(Just ('b',20),'b')]
+beneath :: Fold s a -> Fold s (s, a);
```
phadej, line 288: extra `;`.

The first argument should be first-order. (E.g. `foldrOf :: Getting (Endo r) s a -> (a -> r -> r) -> r -> s -> r`)

I don't understand how this is related to `aside`. Then the type should be

```haskell
beneath :: Fold s a -> Fold (e, s) (e, a)
```

phadej (changes requested):
I left technical comments.

However I'm not sure this combinator should be added in the first place. I don't see a proper motivation, neither I understand how it's related to `aside`. What problem it solves? How you would write this code given bar `s -> [a]` function and list comprehensions?

src/Control/Lens/Fold.hs
dminuoso: What do you mean by `the first argument should be first-order`?

You are right, the change of type with regards to `aside` was a bit accidental, I twisted it a bit further because it fit my usage pattern. Perhaps this is up-to-debate, since plumbing with some `to dup` (see example below) is possible too.

The main motivation is a sort of filtering a'la mapMaybe, say you have some
```haskell
let xs = [ Just (Right 'a', 10)
         , Just (Left 'b', 20)
         , Nothing
         , Just (Left 'c', 30)
         ]
```

And you want to fold over all the things that happen to have a Left deep inside, together with the matching value. 

Then you could write say
```haskell
matching = xs ^.. folded . beneath (_Just . _1 . _Left)
-- or if we assumed `beneath' :: Fold s a -> Fold (e, s) (e, a)`
matching' = xs ^.. folded . to dup . beneath' (_Just . _1 . _Left)

```

With `aside` we'd require repetitive labor:

```haskell
dup x = (x, x)
matching = xs ^.. folded . to dup . aside _Just . alongside id _1 . aside _Left
```


The path into the "sub-focus" (the Fold you'd pass to beneath) would require interleaving aside and alongside in appropriate numbers until you get to the focus like in the example above.

dminuoso:
As idnar pointed out in #haskell-lens, this already exists in a more general form:

filteredBy :: (Indexable i p, Applicative f) => Getting (First i) a i -> p a (f a) -> a -> f a

PR #977, opened 2021-04-30
Use GHC 8.0 as the minimum

### LINKED ISSUE #928: -Wstar-is-type
``` 
   Using ‘*’ (or its Unicode variant) to mean ‘Data.Kind.Type’
    relies on the StarIsType extension, which will become
    deprecated in the future.
    Suggested fix: use ‘Type’ from ‘Data.Kind’ instead.
```

I have used evil trick of 

```haskell
#define Type *
```

for pre GHC-8.0. And we can re-export `Type` from internal prelude. (Not sure if `TH` module will freak out on that).

### LINKED ISSUE #964: Should we (gradually) add -Wno-star-is-type to ekmett packages?
It is part of `-Wall` with GHC-9.0 it seems.

Alternatives are
- drop pre-GHC-8.0 support and use `import Data.Kind (Type)`
- `#define Type *` for older GHCs

RyanGlScott:
This drops support for GHC 7.8 and 7.10, which allows us to clean up lots of code in the process.

Fixes #928. Fixes #964.


phadej:
I run a dependency lower bound checker, and the lower bounds for following dependencies can be bumped.

```
bifunctors                     5.5.7
contravariant                  1.4 
distributive                   0.5.1 
exceptions                     0.8.2.1
hashable                       1.2.7.0
mtl                            2.2.1
parallel                       3.2.1.0
semigroupoids                  5.0.1
tagged                         0.8.6
template-haskell               2.11.1.0 
transformers                   0.5.0.0
transformers-compat            0.5.0.4 
vector                         0.12.1.2
```

I.e. solver was not able to find an install plan with lower versions for these dependencies (with any GHC>=8.0). I haven't checke whether we have CPP related to these versions, but it would help solver to be a slightly faster alone by pruning the search space earlier. (there is MIN_VERSION_mtl at least).

RyanGlScott:
Thanks for checking! I'm a bit confused as to how the lower bound checker arrived at `template-haskell-2.11.1.0`, however, since `template-haskell-2.11.0.0` is bundled with GHC 8.0.1.

src/Control/Lens/Plated.hs
```diff
@@ -102,9 +102,6 @@ import Control.Lens.Traversal
 import Control.Monad.Free as Monad
 import Control.Monad.Free.Church as Church
 import Control.Monad.Trans.Free as Trans
-#if !(MIN_VERSION_free(4,6,0))
```
Taneb, line 112: Shouldn't this have been made unconditional rather than removed?

RyanGlScott: The lower version bounds on `free` are currently:

https://github.com/ekmett/lens/blob/d47860c8498d4db20c8a17756a5d4c52be256683/lens.cabal#L187

As a result, this code can never be reached, unless I'm overlooking something.

phadej:
> Thanks for checking! I'm a bit confused as to how the lower bound checker arrived at `template-haskell-2.11.1.0`, however, since `template-haskell-2.11.0.0` is bundled with GHC 8.0.1.

I use GHC-8.0.2. (And wouldn't mind if GHC-8.0.1 is excluded - I don't want to remember what bugs it had)

src/Control/Lens/Plated.hs
Taneb: No, you're right.

RyanGlScott:
Very well. I'll update this to require GHC 8.0.2 (not 8.0.1) as the minimum soon.

RyanGlScott:
I really wish I knew what is causing the mysterious CI timeouts, such as the one observed in this PR [here](https://github.com/ekmett/lens/runs/4074286543?check_suite_focus=true#step:20:683). This isn't just limited to `lens`, as I've observed the same phenomenon in other repos [here](https://github.com/RyanGlScott/text-show/runs/4061687003?check_suite_focus=true#step:17:69) and [here](https://github.com/ekmett/eq/runs/4061798705?check_suite_focus=true#step:17:147), among others.

phadej:
@RyanGlScott yes, i have wondered whether it's `haskell-ci` fault or GHA flakyness.

RyanGlScott:
It's only one data point, but I haven't seen this happen in any CI jobs for `base-compat`, which doesn't use `haskell-ci`.

RyanGlScott:
In any case, one of the GHC 8.0.2 builds did succeed, so I'm going to just merge this.

PR #982, opened 2021-09-10
Add prisms for working with prefixed and suffixed text

### LINKED ISSUE #981: `prefixed` and `suffixed` prisms for Text
We have the `prefixed` and `suffixed` prisms for lists, but shouldn't we have them for both strict and lazy texts as well?

Taneb:
Closes #981


tests/properties.hs
```diff
@@ -32,7 +32,7 @@ import Test.QuickCheck
 import Test.Framework
 import Test.Framework.Providers.QuickCheck2
 import Data.Char (isAlphaNum, isAscii, toUpper)
-import Data.Text.Strict.Lens
+import qualified Data.Text.Strict.Lens as Text
```
phadej, line 35: This is a sign we might want a type-class. If we want these to begin with, it is.

Taneb: Good and interesting point!

RyanGlScott: If we did offer a type class for this, there's a question of what we should do with the existing `prefixed` and `suffixed` functions in `Data.List.Lens`. Would it be preferable to remove them in favor of a type class that provides a single API entrypoint for all types, or would we want to offer the type class alongside specialized `prefixed`/`suffixed` functions in `Data.List.Lens`, `Data.Text.Strict.Lens`, etc.?

phadej: only type-class seems sane. It could be re-exported by `Data.List.Lens` etc (c.f Cons).

src/Control/Lens/Prism.hs
```diff
@@ -361,3 +369,71 @@ _Show = prism show $ \s -> case reads s of
...
+
+instance Prefixed TL.Text where
+  prefixed p = prism' (p <>) (TL.stripPrefix p)
+  {-# INLINE prefixed #-}
+
+instance Prefixed BS.ByteString where
+  prefixed p = prism' (p <>) (BS.stripPrefix p)
+  {-# INLINE prefixed #-}
+
+instance Prefixed BL.ByteString where
```
phadej, line 403: Good. We have `Cons` instances for `ZipList`, `Seq`, various `Vector`s and lens-own `Deque`.

- ZipList should be easy to add
- `Seq` doen't have stripPrefix/stripSuffix, neither `Vector`, but they have `splitAt` and O(1) `length`, so should be  easy to add.
- I don't know `Deque`, so that could be left out.

RyanGlScott:
@Taneb, is this something you're planning to work on soon? I'm interested in landing this before the next `lens` release. If you don't have time, then no worries—I can pick it up (assuming your branch gives the appropriate permissions to `lens` contributors).

Taneb:
I'd forgotten about it, being honest. My main issue was I couldn't figure out how to run the doctests locally, so I'd have had to do a lot of noisy trial and error to get things passing. Other than that I'm happy to work on it. I think the branch is on this repo so anyone with permission for ekmett/lens should be able to push to it

RyanGlScott:
My apologies, I didn't realize you were blocked. Here is my workflow for running the doctests:

1. Build the library with `cabal build`.
2. Download `cabal-docspec` from https://github.com/phadej/cabal-extras/releases/download/cabal-docspec-0.0.0.20210111/cabal-docspec-0.0.0.20210111.xz
3. Run `cabal-docspec` in the top-level `lens` checkout directory. (Make sure the same GHC version you used in step (1) is on your `PATH`.)

If that doesn't work, feel free to shout.

RyanGlScott:
I pushed a commit which fixes the build errors on old GHCs. Or so I thought, at least. There's a separate issue [on GHC 7.8.4](https://github.com/ekmett/lens/runs/4074147313?check_suite_focus=true#step:21:68):

```
[29 of 86] Compiling Control.Lens.Prism ( src/Control/Lens/Prism.hs, /__w/lens/lens/dist-newstyle/build/x86_64-linux/ghc-7.8.4/lens-5.1/build/Control/Lens/Prism.o )

src/Control/Lens/Prism.hs:397:31:
    Not in scope: ‘BS.stripPrefix’
    Perhaps you meant one of these:
      ‘TS.stripPrefix’ (imported from Data.Text),
      ‘TL.stripPrefix’ (imported from Data.Text.Lazy),
      ‘TS.stripSuffix’ (imported from Data.Text)

src/Control/Lens/Prism.hs:401:31:
    Not in scope: ‘BL.stripPrefix’
    Perhaps you meant one of these:
      ‘TL.stripPrefix’ (imported from Data.Text.Lazy),
      ‘TS.stripPrefix’ (imported from Data.Text),
      ‘TL.stripSuffix’ (imported from Data.Text.Lazy)

src/Control/Lens/Prism.hs:431:33:
    Not in scope: ‘BS.stripSuffix’
    Perhaps you meant one of these:
      ‘TS.stripSuffix’ (imported from Data.Text),
      ‘TL.stripSuffix’ (imported from Data.Text.Lazy),
      ‘List.stripSuffix’ (imported from Data.List.Lens)

src/Control/Lens/Prism.hs:435:33:
    Not in scope: ‘BL.stripSuffix’
    Perhaps you meant one of these:
      ‘TL.stripSuffix’ (imported from Data.Text.Lazy),
      ‘TS.stripSuffix’ (imported from Data.Text),
      ‘List.stripSuffix’ (imported from Data.List.Lens)
```

`stripPrefix` was introduced in `bytestring-0.10.8.0`, and unfortunately, GHC 7.8.4 comes bundled with an older version of `bytestring`. This doesn't seem easy to work around unless we backport the entire implementation of `stripSuffix`, which doesn't sound fun. Perhaps we should land #977 first and rebase on top of that.

src/Data/List/Lens.hs
```diff
@@ -89,55 +95,17 @@
 --
 ----------------------------------------------------------------------------
 module Data.List.Lens
-  ( prefixed
-  , suffixed
-  , stripSuffix
+  ( stripSuffix
```
RyanGlScott, line 98: `Data.List.Lens` no longer exports any optics at all after this change, which makes the module Haddocks misleading.

CHANGELOG.markdown
```diff
@@ -7,6 +7,8 @@
   when building with `template-haskell-2.18` or later.
 * Define `_CharTyLit` in `Language.Haskell.TH.Lens` when building with
   `template-haskell-2.18` or later.
+* Add prisms for working with text with prefixes or suffices in
+  `Data.Text.Strict.Lens` and `Data.Text.Lazy.Lens`
```
RyanGlScott, line 11: This should also mention that `prefixed` and `suffixed` have been removed from `Data.List.Lens` and moved to `Control.Lens.Prism`, as this is a breaking change.

src/Data/List/Lens.hs
RyanGlScott: Alternatively, we could re-export `Prefixed` and `Suffixed` from this module for backwards-compatibility, although that would arguably be inconsistent with how we treat other list-adjacent classes (e.g., `Cons` and `Snoc`).

RyanGlScott: After looking at how many [uses of `Data.List.Lens` there are on Hackage](https://hackage-search.serokell.io/?q=Data%5C.List%5C.Lens), I'm less enthusiastic about simply breaking this outright. For the time being, I've decided to re-export `Prefixed` and `Suffixed` from `Data.List.Lens` to ensure that this change is mostly backwards compatible. I do think we should eventually change this (and I've left a note in the module Haddocks saying as much), but I'll wait until the ecosystem has caught up with `lens-5.1` first to ensure a smoother migration path.

src/Control/Lens/Prism.hs
RyanGlScott: I do think we should do this, although I don't want it to hold up this PR. I've opened #989 as a reminder to do this.

RyanGlScott:
I've pushed some commits which add more Haddocks, more changelog descriptions, and makes things more backwards-compatible. I think that takes care of the remaining review suggestions, so unless anyone notices something I've overlooked, I'll plan to merge this within a day or so.

PR #983, opened 2021-09-19
A few cleanups in GHC.Generics.Lens

treeowl:
* Use `EmptyCase` for `V1` instead of old-school forcing.
* Use `coerced` where applicable.

### DIFF
--- a/src/GHC/Generics/Lens.hs
+++ b/src/GHC/Generics/Lens.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE EmptyCase #-}
+{-# LANGUAGE LambdaCase #-}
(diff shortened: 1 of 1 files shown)


src/GHC/Generics/Lens.hs
```diff
@@ -78,28 +80,27 @@ generic1 = iso from1 to1
 {-# INLINE generic1 #-}
 
 _V1 :: Over p f (V1 s) (V1 t) a b
-_V1 _ = absurd where
-  absurd !_a = undefined
+_V1 _ = \case
 {-# INLINE _V1 #-}
 
 _U1 :: Iso (U1 p) (U1 q) () ()
 _U1 = iso (const ()) (const U1)
```
treeowl, line 87: Why is this so lazy? This converts between `undefined` and `()` and between `U1` and `undefined`. Weird.....

RyanGlScott (approved):

PR #992, opened 2021-11-10
Allow building with hashable-1.4.*

RyanGlScott:


phadej:
I restarted CI after I made GHC-9.2 compatible aeson release. The GHC-8.0 job will fail eventually (when it gets to 6 hour default timeout, we should regenerate the CI config). I don't know what might be causing this, as previously everything worked well.

The upstream issue is https://github.com/actions/runner/issues/1326, I assume it's a correct place to report them. I also pinged @emilypi, in hope that HF has some GitHub contacts.

RyanGlScott:
Thanks, @phadej! Let's go ahead and land this, then. I think this is the last change I wanted to make before the next release, so with luck, I'll be able to do that within a day or so.

phadej:
FWIW: there is https://hackage.haskell.org/package/mtl-2.3/candidate which hopefully be released soon.

Based https://github.com/haskell/mtl/issues/86#issuecomment-967759954 i'd extrapolate that the release would happen soon enough.

PR #994, opened 2021-12-09
Use Endo [a] for universe/On/Of implementation

phadej:
I added a benchmark which is maybe exaggerated but highlights the problem:

```
Benchmark                          plated-master  plated-branch   
universe                           0.823e-4       0.613e-4 -25.52%
universeOf (cloneTraversal plate)  0.824e-4       0.609e-4 -26.08%
universeOf Data.template           2.794e-4       3.370e-4 +20.61%
universeOf Data.tinplate           2.234e-4       3.050e-4 +36.53%
universeOf Data.uniplate           2.789e-4       3.472e-4 +24.47%
universeOf Data.uniplate fibExpr   0.840e-4       0.835e-4  -0.60%
universeOf plate                   0.830e-4       0.611e-4 -26.34%
universeOf plate fibExpr           0.453e-4       0.262e-4 -42.04%
Geometric mean                     1.179e-4       1.075e-4  -8.79%

```

I don't know why `Data` based plates are slower though, that is weird.


benchmarks/plated.hs
```diff
@@ -70,6 +70,8 @@ main = defaultMain
   , bench "universeOf Data.template"           $ nf (map (universeOf Data.template)) testsExpr
   , bench "universeOf Data.uniplate"           $ nf (map (universeOf Data.uniplate)) testsExpr
   , bench "universeOf (cloneTraversal plate)"  $ nf (map (universeOf (cloneTraversal plate))) testsExpr
+  , bench "universeOf plate fibExpr"           $ nf (universeOf plate) fibExpr
+  , bench "universeOf Data.uniplate fibExpr"   $ nf (universeOf plate) fibExpr
```
RyanGlScott, line 74: What is the difference between these two benchmarks? Perhaps I'm missing something, but they both seem to just use `nf (universeOf plate) fibExpr`.

src/Control/Lens/Plated.hs
```diff
@@ -385,13 +385,18 @@ universe = universeOf plate
 -- @
 -- 'universeOf' :: 'Fold' a a -> a -> [a]
 -- @
-universeOf :: Getting [a] a a -> a -> [a]
-universeOf l = go where
-  go a = a : foldMapOf l go a
+universeOf :: Getting (Endo [a]) a a -> a -> [a]
```
RyanGlScott, line 388: Presumably the changes to the type signatures in this module warrant a major version bump?

benchmarks/plated.hs
phadej: Fixed

src/Control/Lens/Plated.hs
phadej: Unfortunately yes.

RyanGlScott: OK. Can you mention these changes in the `CHANGELOG`, along with a description of how to migrate code that might be broken by this change? (I'd do so myself, but I'm not sure I understand which situations would break due to these changes.)

phadej: If people have `Getting [a]` values they pass around, they will need to use `Getting (Endo [a])`, but I doubt anyone does that. Or if they define own combinators (which take in `Getting [a] a a`), they would need to change signature. So in short: fix type errors GHC tells you to fix.

(It's a unfortunate that `Getting r s a` has that `r` argument, leaky implementation detail)

RyanGlScott: Here is a proposed addition to the CHANGELOG:

````
5.2 [????.??.??]
----------------
* The type of `universeOf` has changed:

  ```diff
  -universeOf :: Getting       [a]  a a -> a -> [a]
  +universeOf :: Getting (Endo [a]) a a -> a -> [a]
  ```

  In many cases, using `Endo [a]` over `[a]` improves performance. Most call
  sites to `universeOf` will not be affected by this change, although you may
  need to update your code if you define your own combinators in terms of
  `universeOf`
````

Does this sound right? If so, I can add this after merging the PR.

phadej: that's great! thanks.

RyanGlScott (approved):

PR #995, opened 2021-12-11
Add a hint about using cloneTraversal to the Haddock for Traversal1

### LINKED ISSUE #984: Add a function to turn a `Traversal1` into a `Traversal`
Normally we don't need such a thing due to the typeclass hierarchy, but in the case of `Traversal1`, `Apply` is sadly not a superclass of `Applicative`, so you can't just use a `Traversal1` as a `Traversal`.

It's easy to write a function to do the conversion explicitly, though:
```
traversal1ToTraversal :: Traversal1' a b -> Traversal' a b
traversal1ToTraversal t f = unwrapApplicative . t (WrapApplicative . f)
```

michaelpj:
Fixes #984.

### DIFF
--- a/src/Control/Lens/Type.hs
+++ b/src/Control/Lens/Type.hs
@@ -201,6 +201,11 @@ type Traversal s t a b = forall f. Applicative f => (a -> f b) -> s -> f t
 -- @
 type Traversal' s a = Traversal s s a a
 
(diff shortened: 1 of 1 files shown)


Taneb:
This reads well to me.

RyanGlScott (approved):

Issue #944, opened 2020-10-06
ix is able to insert undefined in strict (hash)map


noughtmare:
See [this reddit monthly hask anything question](https://new.reddit.com/r/haskell/comments/j2q5p8/monthly_hask_anything_october_2020/g7rc4tv). The question is specifically about hashmaps, but I've been able to reproduce the same behavior with normal maps.

The issue is that it is possible to insert undefined into a strict map:

```haskell
import Control.Lens
import qualified Data.Map.Strict as Map

main :: IO ()
main = Map.singleton "foo" 123 & ix "foo" .~ undefined `seq` return ()
```

This should produce an error because a strict Map that contains an undefined value is evaluated before returning. Instead it works without problems.

I tried to see what the issue was, so I manually defined `ix'` using [this definition of `ix`](https://hackage.haskell.org/package/lens-4.19.2/docs/src/Control.Lens.At.html#line-284):

```haskell
import Control.Lens
import qualified Data.Map.Strict as Map

ix' k f m = case Map.lookup k m of
  Just v  -> f v <&> \v' -> Map.insert k v' m
  Nothing -> pure m

main = Map.singleton "foo" 123 & ix' "foo" .~ undefined `seq` ()
```

This does produce the expected error. I wonder if this might be a GHC bug, but I thought I should try here first.

Taneb:
The issue here is that the `Map` types in `Data.Map.Strict` and `Data.Map.Lazy` are in fact the same type, it's just the functions in the module that differ. The `ix` function in `lens` does not have a way to distinguish between the two, as they're the same thing, and cannot tell which of the two modules you have imported (indeed, you may have both or neither in scope).

I'm not sure what the correct next step here is. Improved documentation in `containers` is part of it. Perhaps adding a strict version of `ix` or operators for using setters strictly?

noughtmare:
Ah, I hadn't considered that.

I personally feel like the best solution would be to split the types into two. That way type classes could actually differentiate the two. But I guess that would cause all sorts of breakage and it also reduces reusability of the data structure.

I think a strict `ix` would also be a very nice solution. I wonder how it would interact with multiple layers and other data structures. And I'm not an expert and I also don't use this kind of code very often, so I think some other users/developers should chime in.

arybczak:
FWIW `optics` has strict set variants (`set'` and `!~`) as well at `at'` (where even strict set doesn't help since the result is wrapped in a `Maybe`) that solve that problem.

```
[...]

PR #1017, opened 2022-12-21
Add strict over variants

### LINKED ISSUE #1016: And a strict over?
I don't see a strict version of `over`, which we could put in `Control.Lens.Traversal`.

```haskell
over' :: ATraversal s t a b -> (a -> b) -> s -> t
over' l f = getSolo . cloneTraversal l (\old -> Solo $! f old)
  where
    getSolo (Solo x) = x
```

I don't remember how indexed stuff works, but I'm sure that's available too.

treeowl:
Add `over'` and `iover'`, which force the result(s) of applying the passed function.

Closes #1016


treeowl:
@ekmett and @RyanGlScott , what do you think? Two questions in my mind:

1. Most things around are marked `INLINE`. Should these be too?
2. Should there be operator forms? `%~!` and `%@~!`, perhaps? Or would the bangs go in the middle somewhere?

treeowl:
Another question: as @phadej suggested, I wrote these to take specific traversals, rather than `ATraversal` and `AnIndexedTraversal`. I don't really understand the conventions around which style to use where.

ekmett:
Combinators can generally ask for any functor that they need, so that's fine.

The main uses for ATraversal/AnIndexedTraversal is to allow users to pass around lists of traversals and the like, and to reconstitute them into real Traversals / IndexedTraversals through cloneTraversal.

ekmett:
These should be marked INLINE, and the current morphology would likely lead to %!~ and %!@~ probably, (and %!= and %!@= respectively for state). The nice thing about the state one is you can use the usual strict function over a monad encoding I think using <$!>.

src/Control/Lens/Traversal.hs
```diff
@@ -1466,3 +1468,28 @@ traverseByOf l pur app f = reifyApplicative pur app (l (ReflectedApplicative #.
 -- @
 sequenceByOf :: Traversal s t (f b) b -> (forall x. x -> f x) -> (forall x y. f (x -> y) -> f x -> f y) -> s -> f t
 sequenceByOf l pur app = reifyApplicative pur app (l ReflectedApplicative)
+
+-- | A version of 'Control.Lens.Setter.over' that forces the result(s) of
+-- applying the function. This can prevent space leaks when modifying lazy
+-- fields.
+--
+-- @
+-- over traverse (const ⊥) [1,2] = [⊥, ⊥]
```
phadej, line 1477: These can be turned into proper doctests. `length $ ...`  and matching like

```
...
...Exception...
...
```

for an error. It's not the prettiest, but  it's an honest test.

src/Control/Lens/Traversal.hs
```diff
@@ -1466,3 +1468,28 @@ traverseByOf l pur app f = reifyApplicative pur app (l (ReflectedApplicative #.
...
+-- over' traverse (const ⊥) [1,2] = ⊥
+-- @
+--
+-- @
+-- over' :: 'Lens' s t a b -> (a -> b) -> s -> t
+-- over' :: 'Traversal' s t a b -> (a -> b) -> s -> t
+-- @
+over' :: LensLike Solo s t a b -> (a -> b) -> s -> t
+over' l f = getSolo . l (\old -> Solo $! f old)
```
phadej, line 1486: It maybe worth to add a quick note (or reference) why one needs a boxed type here, i.e. `Solo` and that using `Identity` won't work. (I actually don't know why, someone else looking how `lens` is implemented may wonder as well).

Secondly, we should document (probably publicly) that `Solo` can be given `Settable` instance, but probably shouldn't to prevent this combinator working with true setters and lazy functors (as they should be, I guess?)

```haskell
>>> length $ over' mapped (const undefined) [1,2]
2
```

(https://github.com/ekmett/distributive/issues/67 somewhat related).

src/Control/Lens/Traversal.hs
treeowl: To my mind doctests wouldn't do as good a job of documenting, but I can write them anyway. How do I run those locally?

src/Control/Lens/Traversal.hs
treeowl: `traverse @Identity` is operationally identical to `fmap` (see, e.g., `fmapDefault`). A simple example:

```haskell
data Pair a = Pair a a
  deriving (..., Traversable)

traverse @Pair @Solo f (Pair a b)
= liftA2 Pair (f a) (f b)
= case (f a, f b) of
      (Solo a', Solo b') -> Solo (Pair a' b')

traverse @Pair @Identity f (Pair a b)
= case (f a, f b) of
     (Identity a', Identity b') -> Identity (Pair a' b')
```

In the `Identity` case, the pattern matches on `Identity` don't force anything (it's just a newtype), but in the `Solo` case, they force the applications to WHNF to reveal the `Solo` constructors. That forces whatever computation `f` chooses to perform before revealing the constructor; in our case, it forces the result inside to WHNF.

treeowl: Yes, a `Settable` instance for `Solo` sounds bad from this perspective. I don't know enough about `lens` to say if there's ever any reason to want that, but it seems unlikely. Should we ban the instance as a `TypeError`?

treeowl:
> These should be marked INLINE, and the current morphology would likely lead to %!~ and %!@~ probably, (and %!= and %!@= respectively for state). The nice thing about the state one is you can use the usual strict function over a monad encoding I think using <$!>.

Are you suggesting traversing *in* the ambient monad? Or just pattern matching out of `Solo` into it?

src/Control/Lens/Traversal.hs
phadej: Don't explain it to *me*, explain it to the (future) reader of the code. This is why asked for a reference or a note.

That said, it probably should be explained in `base`'s `Solo` docs, so that could be used as reference; but there is only a side-comment mention about "non-leaky traversals".

ekmett:
It's a state monad variant, so in theory the <$!> trick should be usable in the ambient monad. Not wedded to doing it that way, was just indicating it's probably doable and quite clean.

src/Control/Lens/Traversal.hs
phadej:
```
cabal build all --enable-tests
cabal-docspec
```

phadej: > To my mind doctests wouldn't do as good a job of documenting, but I can write them anyway.

There is barely non-doctest examples in `lens`. They may do less good job as documenting, but at least they stay up to date. There are *hundreds* of examples in `lens` package. No-one would ever go through them by hand and verify they still type-check/work/etc.

src/Control/Lens/Setter.hs
```diff
@@ -342,6 +342,10 @@ cloneIndexedSetter l pafb = taintedDot (runIdentity #. l (Indexed $ \i -> Identi
 -- >>> over _1 show (10,20)
 -- ("10",20)
 --
+-- Like 'fmap', @over@ is normally lazy in the result(s) of calling the
```
phadej, line 345: I'm also on the edge with this "warning".

Shouldn't the strict version also have a similar warning that it *actually* forces all the computations up front. I.e. `over' traverse f aList` is most likely a wrong thing to do. (And the examples used are contrived).

treeowl: Yes, they're contrived. Should I give a datatype definition and `makeLenses` and all that jazz? A more realistic case might be a record with a lazy field. If you use `over` several times, you'll build a thunk chain in there. `over' traverse` isn't too far from what, say, `Data.Map.Strict.map` does, though it's probably less efficient.

phadej: That's a problem. I think in majority of relevant cases you'd rather make the field strict. In fact, there are people which would prefer `Data.Map.Strict` to be a separate datatype. (The type system doesn't help to keep the `Map` thunk free.)

i.e. while `over'` is useful, I don't think it should be commonly used.

phadej:
This solves https://github.com/ekmett/lens/issues/944

src/Control/Lens/Setter.hs
phadej: The only case I can think out of my head is `Seq`. Where `over' traverse` doesn't leave thunks, but AFAIK ruins asymptotics.

treeowl:
It occurs to me that there's another approach to this issue, and it's not entirely clear whether one or the other is better. First, an exceedingly simple monad transformer that probably belongs in `transformers`:

```haskell
newtype SoloT m a = SoloT
  { runSoloT :: m (Solo a) }

-- Note: in the below instances, clever implementations of <$,
-- <*, *>, or >> will almost certainly be wrong; the defaults are
-- correct.
instance Monad m => Functor (SoloT m) where
  fmap = liftM
instance Monad m => Applicative (SoloT m) where
  pure = SoloT . pure . Solo
  (<*>) = ap
instance Monad m => Monad (SoloT m) where
  m >>= f = SoloT $ runSoloT m >>= \(Solo a) -> runSoloT (f a)
instance MonadTrans SoloT where
  lift = SoloT . fmap Solo
```

Now, we can strictify monadic `LensLike`s:

```haskell
strictly :: Monad m => LensLike (SoloT m) s t a b -> LensLike m s t a b
strictly l f s = runSoloT (l (\a -> SoloT $ f a >>= \ !b -> pure $ Solo b) s) >>= \(Solo t) -> pure t
```

In particular, we can write

```haskell
overly' :: LensLike (SoloT Identity) s t a b -> (a -> b) -> s -> t
overly' l = over (strictly l)
```

I still think a direct implementation of `over'` is better for its exact purpose, but maybe `strictly` is also a good thing to have on hand?

treeowl:
`strictly` seems stricter, actually, which may be a bad thing. Let's stick with `over'` stuff for now, and think through the rest later.

treeowl:
Sorry, no it's not. This stuff is confusing! Let's consider both.

treeowl:
Ah, now I understand why that felt off. The monad transformer is the wrong thing, but there's a very similar (slightly lazier) *applicative* transformer that I believe does what we mean.

```haskell
type BoxT :: (Type -> Type) -> Type -> Type
newtype BoxT m a = BoxT
  { runBoxT :: m (Solo a) }
  deriving (Functor)

instance Apply m => Apply (BoxT m) where
  liftF2 f (BoxT m) (BoxT n) = BoxT (liftF2 (liftA2 f) m n) 
  {-# INLINE liftF2 #-}
instance Applicative m => Applicative (BoxT m) where
  pure = BoxT . pure . Solo
  {-# INLINE pure #-} 
  BoxT m <*> BoxT n = BoxT (liftA2 (<*>) m n)
  {-# INLINE (<*>) #-}
#if MIN_VERSION_base(4,10,0)
  liftA2 f (BoxT m) (BoxT n) = BoxT (liftA2 (liftA2 f) m n)
  {-# INLINE liftA2 #-}
#endif
```

Now, we can be exceedingly general:

```haskell
strictly :: (Functor m, Profunctor p, Profunctor q) => Optical p q (BoxT m) s t a b -> Optical p q m s t a b
strictly l f = rmap (fmap getSolo .# runBoxT) $ l (rmap (\mb -> BoxT ((Solo $!) <$> mb)) f)
{-# INLINE strictly #-}
```

src/Control/Lens/Setter.hs
treeowl: @phadej `over'` could quite reasonably be used with `ix` for `Array`, `SmallArray` (if that has an instance), `Vector`, `Seq`, `Map`, or `IntMap`. I imagine some folks just use `at` instead, but that's semantically overkill for just forcing.

src/Control/Lens/Traversal.hs
treeowl: I'm getting errors from `cabal-doctest` about mismatched GHC and plan versions. Unfortunately, `cabal-doctest` doesn't seem to have much in the way of documentation, so I can't figure out where to go from there.

src/Control/Lens/Traversal.hs
```diff
@@ -1466,3 +1478,82 @@ traverseByOf l pur app f = reifyApplicative pur app (l (ReflectedApplicative #.
...
+{-# INLINE strictly #-}
+
+-- | A very simple applicative transformer that gives us more control over when
+-- things get forced. Note: this type /should not/ be made an instance of
+-- @Settable@, because then users could accidentally use 'strictly' with a
+-- 'Setter', which will not work at all. There is no way to strictify a
+-- 'Setter'.
+newtype BoxT m a = BoxT
+  { runBoxT :: m (Solo a) }
+  deriving (Functor, Foldable, Traversable)
```
treeowl, line 1544: Where should this type be defined and exported?

treeowl:
I don't know how much to overgeneralize things. We could, for example, write

```haskell
over' :: (Profunctor p, Profunctor q) => Optical p q Solo s t a b -> p a b -> q s t
over' l f = rmap getSolo $ l (rmap (Solo $!) f)
```

Is that more useful or more confusing?

src/Control/Lens/Traversal.hs
phadej: So you don't want to wait until `SoloT` is added somewhere, as you proposed in private mail to me and Ross (maintainer of `transformers`)?

I understand that you have time to work on stuff during holidays, but I honestly don't, so will politely ignore you until the next year.

phadej (commented):
This pull request has overgrown what I feel comfortable to comment about. I'd ask @ekmett to review this.

src/Control/Lens/Traversal.hs
treeowl: `SoloT` turned out to be the wrong sort of transformer for this purpose. I made a mistake and changed course.

phadej: Please tell that to others in that email message too. I wasn't aware that I don't need to think about that anymore.

phadej:
> I don't know how much to overgeneralize things. We could, for example, write
> 
> ```haskell
> over' :: (Profunctor p, Profunctor q) => Optical p q Solo s t a b -> p a b -> q s t
> over' l f = rmap getSolo $ l (rmap (Solo $!) f)
> ```
> 
> Is that more useful or more confusing?

Will `over' traverse (+1) [1,2,3]` still work?

treeowl:
> > I don't know how much to overgeneralize things. We could, for example, write
> > ```haskell
> > over' :: (Profunctor p, Profunctor q) => Optical p q Solo s t a b -> p a b -> q s t
> > over' l f = rmap getSolo $ l (rmap (Solo $!) f)
> > ```
> > Is that more useful or more confusing?
> 
> Will `over' traverse (+1) [1,2,3]` still work?

Yes. `p` and `q` both specialize to `->`, so `rmap` becomes `(.)`.

phadej:
So can `over` be of `Optical p q Identity s t a b -> p a b -> q s t` form? I'd say it's both or neither.

treeowl:
In that scheme, `over` would become

```haskell
over :: (Profunctor p, Profunctor q) => Optical p q Identity s t a b -> p a b -> q s t
over l f = runIdentity #. l (Identity #. f)
```

src/Control/Lens/Traversal.hs
treeowl: I learned yesterday that this idea has been invented before (rather unsurprisingly); I've edited, expanded, and adapted a note about it in `optics` for inclusion here.

src/Control/Lens/Traversal.hs
treeowl: I learned yesterday that this idea has been invented before (rather unsurprisingly); I've edited, expanded, and adapted a note about it in `optics` for inclusion here.

src/Control/Lens/Traversal.hs
treeowl: Done.

treeowl:
It occurs to me that when used in a non-trivial functor, the traversal produced by `strictly` will *always* leave thunks in it. Only forcing those thunks from the outside will actually force the targets. This can be avoided if the ambient functor is *either* a `Traversable` or a `Monad`:

```haskell
strictlyT :: (Traversable f, Profunctor p, Profunctor q) => Optical p q (BoxT f) s t a b -> Optical p q f s t a b
strictlyT l f = rmap (getSolo . sequenceA .# runBoxT) $ l (rmap (BoxT #. fmap (Solo $!)) f)

strictlyM :: (Monad f, Profunctor p, Profunctor q) => Optical p q (BoxT f) s t a b -> Optical p q f s t a b
strictlyM l f = rmap ((>>= \(Solo r) -> pure r) .# runBoxT) $ l (rmap (BoxT #. fmap (Solo $!)) f)
```

Is either of these worth having?

ekmett:
i think they are worth having, i hate the names, but not the functionality.

treeowl:
@ekmett should the monadic version use the `SoloT` monad transformer instead of the `BoxT` applicative transformer? My guess is yes, but I probably need to play with some benchmarks.

PR #1021, opened 2023-02-01
Add `_TypeDataD` for `template-haskell-2.20.*` (GHC 9.6)

RyanGlScott:


RyanGlScott:
Huh, the GHC 8.2 CI job is timing out while running Haddock of all things. I have absolutely no idea what is going on there.

The GHC 9.6 CI job fails due to picking a rather silly build plan. My hope is that this will be resolved with haskell/aeson#1005.

PR #1027, opened 2023-03-15
Allow these-1.2, strict-0.5, assoc-1.1

phadej:
Note: there is no problem with `Bifunctor` as lens requires GHC-8.0, for which `Bifunctor` is always unique.

OTOH, `swap :: Bitraversable p =>`... has such problem, so maybe it's easiest to drop GHC-8.0 support.

Also update CI to not use head.hackage as deps are all updated

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.15.20230227
+# version: 0.15.20230313
 #
-# REGENDATA ("0.15.20230227",["--config=cabal.haskell-ci","cabal.project"])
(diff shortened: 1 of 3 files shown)


RyanGlScott:
Unfortunately, I think we need to give the GHC 9.6.1 job _some_ kind of special treatment, as not all of `lens`'s dependencies have upper version bounds that support 9.6 yet (e.g., `vector-th-unbox`). I don't have a strong opinion on how we accomplish this: we could continue to use `head.hackage` in the CI job, or we could try some `--allow-newer`-related tweaks.

phadej:
@RyanGlScott I think few selected `allow-newer` is better for now, if we don't actually need any patches from `head.hackage`. I'll amend this PR shortly.

RyanGlScott:
Thanks! I've applied a Hackage revision [here](https://hackage.haskell.org/package/lens-5.2.1/revisions/).

PR #1041, opened 2023-07-12
Remove docs saying toNonEmptyOf works on a Prism'

sjshuck:
Was working nearby and happened to notice it.

### DIFF
--- a/src/Control/Lens/Fold.hs
+++ b/src/Control/Lens/Fold.hs
@@ -639,7 +639,6 @@ toListOf l = foldrOf l (:) []
 -- 'toNonEmptyOf' :: 'Lens'' s a       -> s -> NonEmpty a
 -- 'toNonEmptyOf' :: 'Iso'' s a        -> s -> NonEmpty a
 -- 'toNonEmptyOf' :: 'Traversal1'' s a -> s -> NonEmpty a
--- 'toNonEmptyOf' :: 'Prism'' s a      -> s -> NonEmpty a
 -- @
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Good catch.

Most `Getting` arguments can be instantiated with a `Prism`, but `toNonEmptyOf` is an exception. Can you make a note of this in the documentation, along with an explanation of why it doesn't work for `toNonEmptyOf`?

sjshuck:
> make a note of this

under `Getting`, 
> along with an explanation

under `toNonEmptyOf`?

RyanGlScott:
I was mainly thinking of the documentation for `toNonEmptyOf`, but I suppose it wouldn't hurt to mention this in the documentation for `Getting` as well. I'll leave that part up to you.

sjshuck:
So, looking through the rest of the file, there are a bunch of other combinators that _aren't_ shown with `Prism'` but _could_ be, and there are a bunch more like `maximum1Of` that deserve a note as much as `toNonEmptyOf` does.

I'd be happy to drop in lots of comments about `Prism'` and `Apply`/`Semigroup` but I'm starting to get worried about the total amount of comments&mdash;I wonder if some of these charts could be deduplicated, just saying "See `first1Of`", "See `first1Of`" etc., or "Accepts `Lens`, `Fold`, ..."

Perhaps a thorough going-over should be another issue, and this is just to delete wrong info?  Let me know if I'm making a big deal about this.

PR #1042, opened 2023-07-22
Disambiguate Haddock references

ncfavier:
Disambiguate references to `Lens`, `Iso`, etc. by adding the type [namespace marker](https://haskell-haddock.readthedocs.io/en/latest/markup.html#hyperlinked-identifiers), so that Haddock doesn't confuse them with the homonymous data constructors in `Control.Lens.Reified` (among other trivial fixes). Removes most of the "ambiguous" warnings from `cabal haddock`'s output.

Please remember to use `t'...'` where needed when writing future documentation!

Before:
screenshot-2023-07-22-181356

After:
screenshot-2023-07-22-181408


phadej:
This seems to be relatively new haddock feature? Is it documented, where? When (in which `haddock` version) it was added?

When building with haddocks bundled with GHC-8.0.2 and GHC-8.6.5 (which I still occasionally use), the output is broken:

Screenshot from 2023-07-22 19-47-54

ncfavier:
Added in Haddock 2.23 https://github.com/haskell/haddock/commit/dd47029cb29c80b1ab4db520c9c2ce4dca37f833

RyanGlScott:
Good catch. As noted in https://github.com/ekmett/lens/pull/1042#issuecomment-1646628545, we should consider how we want to handle things in old GHCs, but I'm optimistic that we can figure out something.

Out of curiosity, why does `'Lens'` get rendered as `RenderedLens` in the Haddock? I get that `Lens` is a data constructor for `RenderedLens`, but I would have still expected the Haddocks to still say "Lens", even if it links to `RenderedLens`. Outright changing the way it's rendered in the Haddocks almost feels like a Haddock bug.

RyanGlScott:
Since I had to look this up...

> Added in Haddock 2.23 [haskell/haddock@dd47029](https://github.com/haskell/haddock/commit/dd47029cb29c80b1ab4db520c9c2ce4dca37f833)

...the earliest GHC that bundled this version of Haddock appears to be GHC 8.8:

```
$ haddock-8.8.4 --version
Haddock version 2.23.0, (c) Simon Marlow 2006
Ported to use the GHC API by David Waern 2006-2008
```

phadej:
> In such a case, Haddock defaults to pointing to the type.

Is documentation in conflict with the implementation? I read that the (way fewer) `v'Lens'` would be required to annotated instead of `t'Lens'` across all of the package?

phadej:
@RyanGlScott it does seem that it is outright haddock bug, as it currently renders

```
-- | Bitwise '.|.' the target(s) of a 'Lens' or 'Setter'.
--
-- >>> _2 .|.~ 6 $ ("hello",3)
-- ("hello",7)
--
-- @
-- ('.|.~') :: 'Bits' a             => 'Setter' s t a a    -> a -> s -> t
-- ('.|.~') :: 'Bits' a             => 'Iso' s t a a       -> a -> s -> t
-- ('.|.~') :: 'Bits' a             => 'Lens' s t a a      -> a -> s -> t
-- ('.|.~') :: ('Data.Monoid.Monoid' a, 'Bits' a) => 'Traversal' s t a a -> a -> s -> t
-- @
(.|.~):: Bits a => ASetter s t a a -> a -> s -> t
```

with `ReifiedSetter` etc.

---

But according to the docs, when ambiguous it should prefer **the type**, so that is another bug in the implementation.

RyanGlScott:
Haddock bugs notwithstanding, we should figure out a way to get the Haddocks to render the way they were intended to. I don't suppose that there is a way to make `t'Lens'` render in a somewhat sensible way on old versions of Haddock? If not, we'll probably just have to include a disclaimer somewhere that you need a sufficiently new version of Haddock to build the documentation.

ncfavier:
Maybe fully qualifying the module (`'Control.Lens.Lens.Lens'`) would be enough. I'll try later

phadej:
> If not, we'll probably just have to include a disclaimer somewhere that you need a sufficiently new version of Haddock to build the documentation.

I'd rather have `haddock` do what is specified, and not crap everything with unneeded (according to docs) `t` prefix. And then we can say that use (much more recent, but still recent) haddock version to avoid incorrect links.

The qualifying is ugly too. And it makes aligning even more difficult. (IMO if you go for `t'` prefix then it should be added *for all type constructors* uniformly).

RyanGlScott:
I share your concern about needing to attach `t` prefixes (or full qualification) to a substantial number of identifiers in `lens`'s Haddocks. As a developer, I would find that annoying (and easy to forget).

On the other hand, the current status quo is pretty broken. I wouldn't be opposed to adopting an ugly (but serviceable) fix of having more verbose Haddock comments with the understanding that we will pursue a proper fix for this issue upstream in Haddock, which we would adopt later.

ncfavier:
So apparently, on both GHC 8.6 and 9.2, `'Control.Lens.Lens.Lens'` causes haddock to generate a link to `Control-Lens-Lens.html#v:Lens` instead of `#t:Lens`...

RyanGlScott:
I've submitted https://github.com/haskell/haddock/issues/1608 about the underlying Haddock issue.

Issue #1043, opened 2023-08-01
Build failure with `template-haskell-2.21` (GHC 9.8)


andreasabel:
```
src/Language/Haskell/TH/Lens.hs:763:33: error: [GHC-83865]
    • Couldn't match type ‘BndrVis’ with ‘()’
      Expected: TypeFamilyHead -> [TyVarBndrUnit]
        Actual: TypeFamilyHead -> [TyVarBndr BndrVis]
    • In the first argument of ‘lens’, namely ‘g’
      In the expression: lens g s
      In an equation for ‘typeFamilyHeadTyVarBndrs’:
          typeFamilyHeadTyVarBndrs
            = lens g s
            where
                g (TypeFamilyHead _ tvbs _ _) = tvbs
                s (TypeFamilyHead n _ rs ia) tvbs = TypeFamilyHead n tvbs rs ia
    |
763 | typeFamilyHeadTyVarBndrs = lens g s where
    |             
...
```
This could be fixed by using ideas from the patch on head.hackage: https://gitlab.haskell.org/ghc/head.hackage/-/blob/master/patches/lens-5.2.2.patch

RyanGlScott:
Sorry for the delay on this—things always tend to become busy around a new GHC release. I'll plan to fix this within the coming days.

RyanGlScott:
I've uploaded [`lens-5.2.3`](https://hackage.haskell.org/package/lens-5.2.3) to Hackage with a fix.

PR #1046, opened 2023-08-22
Allow building with `template-haskell-2.21.*` (GHC 9.8)

RyanGlScott:
This is made quite simple by using the `TyVarBndrVis`/`bndrVis` compatibility shims from `th-abstraction-0.6.*`. When that version of `th-abstraction` is not available, I provide these definitions using CPP.

Fixes https://github.com/ekmett/lens/issues/1043.

While I was in town, I also made sure that the `Prism`s that `Language.Haskell.TH.Lens` offers are up to date with recent versions of the `template-haskell` library.


RyanGlScott:
`docspec` is [failing on the GHC 9.8.1 job](https://github.com/ekmett/lens/actions/runs/5939537105/job/16106134731?pr=1046#step:24:176), which reveals that there are some uses of `head` and `tail` in the doctests which I failed to notice in #1045:

```
[  20.23108] error: [1,2,3,4] & upon head .~ 0
[  20.23164] error: in comment at 160:1
+
+<interactive>:9088:18: warning: [GHC-63394] [-Wx-partial]
+    In the use of ‘head’
+    (imported from Prelude, but defined in GHC.List):
+    "This is a partial function, it throws an error on empty lists. Use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
 [0,2,3,4]
```
```
[  20.26376] error: upon' (tail.tail) .~ [10,20] $ [1,2,3,4]
[  20.26495] error: in comment at 204:1
+
+<interactive>:9098:8: warning: [GHC-63394] [-Wx-partial]
+    In the use of ‘tail’
+    (imported from Prelude, but defined in GHC.List):
+    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
+
+<interactive>:9098:13: warning: [GHC-63394] [-Wx-partial]
+    In the use of ‘tail’
+    (imported from Prelude, but defined in GHC.List):
+    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
 [1,2,10,20]
```
```
[  20.30274] error: onceUpon (tail.tail) .~ [10,20] $ [1,2,3,4] -- BAD
[  20.30804] error: in comment at 225:1
+
+<interactive>:9108:11: warning: [GHC-63394] [-Wx-partial]
+    In the use of ‘tail’
+    (imported from Prelude, but defined in GHC.List):
+    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
+
+<interactive>:9108:16: warning: [GHC-63394] [-Wx-partial]
+    In the use of ‘tail’
+    (imported from Prelude, but defined in GHC.List):
+    "This is a partial function, it throws an error on empty lists. Replace it with drop 1, or use pattern matching or Data.List.uncons instead. Consider refactoring to use Data.List.NonEmpty."
 [1,10,20]
```

I'll have to think about the best way to adapt these examples.

RyanGlScott:
I took a look at the `Data.Data.List` doctests that use `head`/`tail`, and remarkably, some of these actually _rely_ on `head`/`tail` being partial functions, such as this one:

https://github.com/ekmett/lens/blob/72678199d5e722233f7b8f906f4c34a26294d934/src/Data/Data/Lens.hs#L181-L182

It turns out that there is some grimy code [here](https://github.com/ekmett/lens/blob/72678199d5e722233f7b8f906f4c34a26294d934/src/Data/Data/Lens.hs#L152-L156) that catches all exceptions thrown (in pure code!) when invoking `upon` in order to make this work... wild stuff.

We could certainly define our own custom versions of `head` and `tail` to avoid triggering `-Wx-partial` warnings, but that would arguably make the doctests less useful, since readers would have to figure out what our custom functions are doing. `head` and `tail`, dangerous as they may be, are well-known among Haskellers, and using them has the advantage that readers don't have to spend mental energy figuring out what they're meant to do.

For now, I'm inclined to just disable `-Wx-partial` when running the doctests in `Data.Data.Lens`. It's possible that we may have to revisit this choice some day (e.g., if a future version of the `Prelude` removes `head`/`tail` altogether), but that is a problem for future me.

RyanGlScott:
As noted in https://github.com/ekmett/lens/commit/850b587e23110c92b9ca5b33264ac8d738af9831#commitcomment-125292040, it's not entirely straightforward to disable a warning only in the doctests, nor is it even guaranteed to work reliably. As such, I've pushed an alternative approach that instead redefines `head`/`tail` in an internal module and replaces the use of `head`/`tail` from the `Prelude` (which emit `-Wx-partial` warnings) with our custom, warning-free versions. Luckily, the end result is that the doctests look exactly the same to readers of the Haddocks.

PR #1052, opened 2023-12-06
Separate workflow for checks

phadej:
I'm experimenting with an idea of generating a separate workflow for checks which are enough to do once. E.g. hlint is run once. `cabal check` does the same for every GHC.

Having such checks in a separate workflow, would:
- allow per-GHC builds to pass even if there is e.g HLint issue
- allow more parallism
- simplify structure

The goal is that `haskell-ci` would be able to generate `lints.yml` file too, as it looks there are potentially a lot of setup to do.


phadej:
Checks are tricky, I'll make another PR if I make any progress.

PR #1058, opened 2024-03-01
feat(Setter): add: prepends setter

ncaq:
Add a setter that prepends the element, like `cons`. For balance, we also add a setter that also appends the element to the tail, like `snoc`. We often use this when adding elements to list structures and other order-conscious things. When we are doing natural language processing, for example, and we are adding processed data, it is more natural to add the resulting processed data to the head side if the original input source is closer to the head. Naturally we can also construct the processing by adding the tail if we think hard enough, but we don't want to work hard at that since we only need to prepare a function. The naming of the `<>:~` function was lost with `~<>`, but considering that the `~` symbol in the lenses setter operator is basically on the right, I used the `:` operator to indicate that it is closer to the `cons` of the list, rather than focusing on the contrast.


RyanGlScott:
Thanks, @ncaq!

Looking at the [`lens` wiki page on infix operators](https://hackage.haskell.org/package/lens-5.2.3/docs/Control-Lens-Operators.html#v:-60--62--126-), I wonder if it would make sense to add "with result" variants of each of these operators. That is, introduce a `(<<>:~)` operator alongside `(<>:~)`, and so on for the other operators.

ncaq:
@RyanGlScott 
"with result" variants have been added to each operator.
Cons operators are now written in the Cons module to avoid circular imports.

ncaq:
I made a mistake in describing the test...
I didn't feel the significance of leaving this in the commit log, so I force pushed it.

RyanGlScott (approved):
Thanks, @ncaq!

PR #1060, opened 2024-03-10
Magnify instance for CPS RWST

BebeSparkelSparkel:
mtl added MonadReader instance for CPS RWST in version 2.3 https://github.com/haskell/mtl/blob/6be8cb55ca9a8bfa6111e3004e4b56b91ed5edcc/Control/Monad/Reader/Class.hs#L117

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -185,7 +185,7 @@ library
     contravariant                 >= 1.4      && < 2,
     distributive                  >= 0.5.1    && < 1,
     exceptions                    >= 0.8.2.1  && < 1,
-    filepath                      >= 1.2.0.0  && < 1.6,
+    filepath                      >= 1.2.0.0  && < 1.5,
     free                          >= 5.1.5    && < 6,
     ghc-prim,
     hashable                      >= 1.2.7.0  && < 1.5,
--- a/lens.cabal
+++ b/lens.cabal
@@ -192,7 +192,7 @@ library
     indexed-traversable           >= 0.1      && < 0.2,
     indexed-traversable-instances >= 0.1      && < 0.2,
     kan-extensions                >= 5        && < 6,
-    mtl                           >= 2.3      && < 2.4,
+    mtl                           >= 2.2.1    && < 2.4,
     parallel                      >= 3.2.1.0  && < 3.3,
     profunctors                   >= 5.5.2    && < 6,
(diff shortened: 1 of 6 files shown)


lens.cabal
```diff
@@ -192,7 +192,7 @@ library
     indexed-traversable           >= 0.1      && < 0.2,
     indexed-traversable-instances >= 0.1      && < 0.2,
     kan-extensions                >= 5        && < 6,
-    mtl                           >= 2.2.1    && < 2.4,
+    mtl                           >= 2.3      && < 2.4,
```
RyanGlScott, line 195: This approach is not going to work, as `mtl` is a boot library that is shipped with GHC. By raising the lower version bounds here, you're cutting off support for GHCs that ship an older version of `mtl` (or, at the very least, making it much harder to construct build plans).

I would advocate for keeping the version bounds as they currently are, but instead guarding the instances behind `#if MIN_VERSION_mtl(2,3,0)` CPP.

src/Control/Lens/Zoom.hs
```diff
@@ -46,6 +46,7 @@ import Control.Monad.Trans.Writer.Lazy as Lazy
 import Control.Monad.Trans.Writer.Strict as Strict
 import Control.Monad.Trans.RWS.Lazy as Lazy
 import Control.Monad.Trans.RWS.Strict as Strict
+import Control.Monad.Trans.RWS.CPS as CPS
```
RyanGlScott, line 49: Make sure to guard this import behind CPP...

src/Control/Lens/Zoom.hs
```diff
@@ -102,6 +103,7 @@ type instance Magnified (ReaderT b m) = Effect m
 type instance Magnified ((->)b) = Const
 type instance Magnified (Strict.RWST a w s m) = EffectRWS w s m
 type instance Magnified (Lazy.RWST a w s m) = EffectRWS w s m
+type instance Magnified (CPS.RWST a w s m) = EffectRWS w s m
```
RyanGlScott, line 106: ...as well as this instance.

RyanGlScott (approved):
Thanks!

RyanGlScott:
I've uploaded [`lens-5.3.1`](https://hackage.haskell.org/package/lens-5.3.1) to Hackage with these changes.

PR #1066, opened 2024-04-21
Add separate HLint workflow

phadej:


phadej:
Ping @RyanGlScott.  I'm planning to remove HLint support from `haskell-ci`. Having a separate workflow for HLint is nowadays very easy (as this PR illustrates). OTOH, having HLInt support in `haskell-ci` is not that easy and convenient.

RyanGlScott (approved):
Thanks, @phadej!

PR #1068, opened 2024-05-11
Re-export #1058 operators from `Control.Lens.Operators`, fix incorrect documentation

### LINKED ISSUE #1067: New operators not exported by Control.Lens.Operators
These operators, introduced in `lens-5.3`, are not re-exported through `Control.Lens.Operators`:

*    (<>:~): prepend an element to the front via (<>).
*    (<<>:~): ~~append an element to the back~~ prepend an element to the front via (<>).
*    (<|~): cons an element to the front via (<|).
*    (<<|~): cons an element to the front via (<|) and return the result.
*    (|>~): snoc an element to the back via (|>).
*    (<|>~): snoc an element to the back via (|>) and return the result.

`(<<>:~)` appears misdocumented in the changelog: it is actually left-side `<>` and-return-result

Should these lenses also have `<<` variants which return the old value?

Ping @ncaq

RyanGlScott:
Fixes #1067.


src/Control/Lens/Lens.hs
```diff
@@ -1192,18 +1192,18 @@ l <<>~ m = l <%~ (<> m)
...
 -- When you do not need the result of the operation, ('Control.Lens.Setter.<>:~') is more flexible.
 (<<>:~) :: Semigroup m => LensLike ((,)m) s t m m -> m -> s -> (m, t)
 l <<>:~ m = l <%~ (m <>)
 {-# INLINE (<<>:~) #-}
 
--- | ('<>') a 'Semigroup' value onto the end of the target of a 'Lens' into
+-- | ('<>') a 'Semigroup' value onto the front of the target of a 'Lens' into
 -- your 'Monad''s state and return the result.
--- However, unlike '<<>=', it is prepend to the head side.
+-- However, unlike ('<<>='), it is prepend to the head side.
```
endgame, line 1206:
```suggestion
-- However, unlike ('<<>='), it is prepended to the head side.
```

endgame (commented):
Thanks for fixing this up, I can only see one doc nit. While we're here, I think we should add the `<<` variants (which return the old result), because the consistency of lens' operator language is one of its greatest strengths. They're also fairly simple to implement.

RyanGlScott:
> While we're here, I think we should add the `<<` variants (which return the old result)

What names would you propose for these? We're in a bit of a weird situation as some of these operators already begin with `<` and `<<`, so the usual convention of adding `<<` to the front of the name to indicate returning the old result doesn't work for all of these operators.

endgame:
I don't think that's a blocker: we already have [`(<<<>=)`](https://hackage.haskell.org/package/lens-5.2.3/docs/Control-Lens-Operators.html#v:-60--60--60--62--61-) for "`<>` onto part of a `MonadState` and return the old value".

There are only really three new operators in `5.3.1`:

* `(<>:~)` (prepend front via `(<>)`)
* `(<|~)` (cons via `(<|)`)
* `(|>~)` (snoc via `(|>)`)

The PR that added them added the the "return new result variants" (beginning with `<`) and the `MonadState` variants (`s/~/=`). I am just suggesting we fill out the rest of the established pattern:

* `(<<<>:~)` (prepend front via `(<>)` and return old result)
* `(<<<>:=)` (prepend front via `(<>)` and return old result, `MonadState` version)
* `(<<<|~)` (prepend front via `(<|)` and return old result)
* `(<<<|=)` (prepend front via `(<|)` and return old result, `MonadState` version)
* `(<<|>~)` (append back via `(|>)` and return old result)
* `(<<|>=)` (append back via `(|>)` and return old result, `MonadState` version)

Personally, I have never really needed the operators in `Control.Lens.Cons` and would have expected `<|>~` to mean "combine the target of the lens with a new value according to its `Alternative` instance", but since it's in `lens` now it's probably best to build out the matching operator variants for consistency.

RyanGlScott:
I've pushed a commit which implements the operators suggested in https://github.com/ekmett/lens/pull/1068#issuecomment-2106086583. Let me know if this covers everything.

src/Control/Lens/Lens.hs
```diff
@@ -1192,24 +1192,42 @@ l <<>~ m = l <%~ (<> m)
...
 {-# INLINE (<<>=) #-}
 
--- | ('<>') a 'Semigroup' value onto the end of the target of a 'Lens' and
+-- | ('<>') a 'Semigroup' value onto the front of the target of a 'Lens' and
 -- return the result.
--- However, unlike '<<>~', it is prepend to the head side.
+-- However, unlike ('<<>~'), it is prepended to the head side.
 --
 -- When you do not need the result of the operation, ('Control.Lens.Setter.<>:~') is more flexible.
 (<<>:~) :: Semigroup m => LensLike ((,)m) s t m m -> m -> s -> (m, t)
```
endgame, line 1200:
```suggestion
(<<>:~) :: Semigroup m => LensLike ((,) m) s t m m -> m -> s -> (m, t)
```

src/Control/Lens/Lens.hs
```diff
@@ -1192,24 +1192,42 @@ l <<>~ m = l <%~ (<> m)
...
 l <<>:~ m = l <%~ (m <>)
 {-# INLINE (<<>:~) #-}
 
--- | ('<>') a 'Semigroup' value onto the end of the target of a 'Lens' into
+-- | ('<>') a 'Semigroup' value onto the front of the target of a 'Lens' and
+-- return the /old/ result.
+-- However, unlike ('<<>~'), it is prepended to the head side.
+--
+-- When you do not need the result of the operation, ('Control.Lens.Setter.<>:~') is more flexible.
+(<<<>:~) :: Semigroup m => LensLike' ((,)m) s m -> m -> s -> (m, s)
```
endgame, line 1209:
```suggestion
(<<<>:~) :: Semigroup m => LensLike' ((,) m) s m -> m -> s -> (m, s)
```

src/Control/Lens/Lens.hs
```diff
@@ -1192,24 +1192,42 @@ l <<>~ m = l <%~ (<> m)
...
+l <<<>:~ m = l <<%~ (m <>)
+{-# INLINE (<<<>:~) #-}
+
+-- | ('<>') a 'Semigroup' value onto the front of the target of a 'Lens' into
 -- your 'Monad''s state and return the result.
--- However, unlike '<<>=', it is prepend to the head side.
+-- However, unlike ('<<>='), it is prepended to the head side.
 --
 -- When you do not need the result of the operation, ('Control.Lens.Setter.<>:=') is more flexible.
 (<<>:=) :: (MonadState s m, Semigroup r) => LensLike' ((,)r) s r -> r -> m r
```
endgame, line 1218:
```suggestion
(<<>:=) :: (MonadState s m, Semigroup r) => LensLike' ((,) r) s r -> r -> m r
```

src/Control/Lens/Lens.hs
```diff
@@ -1192,24 +1192,42 @@ l <<>~ m = l <%~ (<> m)
...
 l <<>:= r = l <%= (r <>)
 {-# INLINE (<<>:=) #-}
 
+-- | ('<>') a 'Semigroup' value onto the front of the target of a 'Lens' into
+-- your 'Monad''s state and return the /old/ result.
+-- However, unlike ('<<<>='), it is prepended to the head side.
+--
+-- When you do not need the result of the operation, ('Control.Lens.Setter.<>:=') is more flexible.
+(<<<>:=) :: (MonadState s m, Semigroup r) => LensLike' ((,)r) s r -> r -> m r
```
endgame, line 1227:
```suggestion
(<<<>:=) :: (MonadState s m, Semigroup r) => LensLike' ((,) r) s r -> r -> m r
```

endgame (approved):
Seems right to me. Just some whitespace lint which you can take or leave.

Thanks for tidying this up!

PR #1070, opened 2024-06-07
Add makeFieldsId for -XNoFieldSelectors users

### LINKED ISSUE #1024: No accessors are generated when using `NoFieldSelectors` to remove prefixes.
With the new `NoFieldSelectors` feature from GHC 9.2, we don't have to prefix every single field with a constructor prefix.

However, lens `makeFields` and others expect the field to start with a constructor prefix or `_`, so if I lose the prefix, they won't create the accessor.

Anyway, my workaround is written below, but I think it would be better to generate accessors by default, given the existence of `NoFieldSelectors` in the first place, than to simply generate this function.
I think that making it the target of `makeLenses` means they expect it to generate accessors.

~~~hs
module Control.Lens.TH.NoFieldSelectors (makeLensesId) where

import           Control.Lens
import qualified Data.Char           as C
import           Language.Haskell.TH

-- | `NoFieldSelectors`を有効にした環境では、
-- フィールドにプレフィクスを付けなくて済みますが、
-- lensがアクセサを作る条件を満たさなくなるため雑に許可します。
makeLensesId :: Name -> DecsQ
makeLensesId = makeLensesWith idRules

idRules :: LensRules
idRules = defaultFieldRules & lensField .~ idNamer

idNamer :: FieldNamer
idNamer _ _ field =
  let fieldPart = nameBase field
      classNamePart = case fieldPart of
        x : xs -> C.toUpper x : xs
        []     -> error "field is empty"
  in [MethodName (mkName $ "Has" ++ classNamePart) (mkName fieldPart)]
~~~

rhendric:
Closes #1024.

### DIFF
--- a/lens.cabal
+++ b/lens.cabal
@@ -333,6 +333,9 @@ test-suite templates
     T799
     T917
     T972
+  if impl(ghc >= 9.2)
+    other-modules:
+      T1024
   ghc-options: -Wall -threaded
   hs-source-dirs: tests
   default-language: Haskell2010
--- a/src/Control/Lens/TH.hs
+++ b/src/Control/Lens/TH.hs
@@ -28,6 +28,7 @@ module Control.Lens.TH
   , makeClassy, makeClassyFor, makeClassy_
   , makeFields
   , makeFieldsNoPrefix
+  , makeFieldsId
   -- ** Prisms
   , makePrisms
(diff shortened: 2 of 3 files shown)


RyanGlScott (approved):
Thanks!

PR #1072, opened 2024-07-05
Update CI

phadej:


lens.cabal
```diff
@@ -402,7 +402,7 @@ test-suite doctests
   hs-source-dirs:    tests
   default-language:  Haskell2010
 
-  build-depends: base, deepseq, simple-reflect >= 0.3.1
+  build-depends: base <5, deepseq, simple-reflect >= 0.3.1
```
phadej, line 405: The `cabal check` (somewhat correctly) complains about not having bound here, as this test-suite doesn't depend on library component, so doesn't inherit dependencies.

RyanGlScott (approved):

PR #1077, opened 2024-09-19
lens.cabal: un-rot spj talk link

yuuko-partyvan:
good talk, got nuked, saved from death, can confirm it really helped get an intuition for the library


RyanGlScott (approved):
Thanks!

PR #1085, opened 2025-01-20
Add instances for strict boxed vector added in vector-0.13.2.0

Shimuuar:
They're protected by CPP. I added all instance other vector types have.

Also I noticed that primitive vector lacks AsEmpty instance so I added it too.


RyanGlScott (approved):
Thanks!

Issue #1084, opened 2025-01-20
foldlOf' is not properly inlined


Shimuuar:
While trying to optimize traverse for vector I started investigate poor performance of `sumOf` and noticed that `foldlOf'` is inlined too late and fails to specialize. This is source code of functions in question:

```haskell
foldlOf' :: Getting (Endo (Endo r)) s a -> (r -> a -> r) -> r -> s -> r
foldlOf' l f z0 xs = ...

sumOf :: Num a => Getting (Endo (Endo a)) s a -> s -> a
sumOf l = foldlOf' l (+) 0
```
GHC won't inline `foldlOf'` body unless it has 4 arguments and `sumOf` and other combinators pass only 3 so they fail specialize properly. Solution is to move `xs` from LHS to lambda on RHS. This greatly improves performance (al produces much more compact core). Following benchmarks demonstrate this:

```haskell
module Main where

import Data.Monoid
import Control.Lens
import Test.Tasty.PAPI

list :: [Double]
list = [1 .. 10000]
{-# NOINLINE list #-}

sumListLens :: [Double] -> Double
sumListLens = sumOf traverse
{-# NOINLINE sumListLens #-}

xsumListLens :: [Double] -> Double
xsumListLens = xsumOf traverse
{-# NOINLINE xsumListLens #-}

foldlListLens :: [Double] -> Double
foldlListLens = foldlOf' traverse (+) 0
{-# NOINLINE foldlListLens #-}

foldlListLensEtaExpand :: [Double] -> Double
foldlListLensEtaExpand xs = foldlOf' traverse (+) 0 xs
{-# NOINLINE foldlListLensEtaExpand #-}

----------------------------------------------------------------

xfoldlOf' :: Getting (Endo (Endo r)) s a -> (r -> a -> r) -> r -> s -> r
xfoldlOf' l f z0 = \xs ->
  let f' x (Endo k) = Endo $ \z -> k $! f z x
  in foldrOf l f' (Endo id) xs `appEndo` z0
{-# INLINE xfoldlOf' #-}

xsumOf :: Num a => Getting (Endo (Endo a)) s a -> s -> a
xsumOf l = xfoldlOf' l (+) 0
{-# INLINE xsumOf #-}

main :: IO ()
main = defaultMain
  [ bench "sumOf"    $ whnf  sumListLens list
  , bench "foldlOf η-contracted" $ whnf foldlListLens list
[...]

PR #1086, opened 2025-01-20
Eta reduce foldlOf' and related functions

Shimuuar:
This is done in order to make GHC inliner more eager to inline them and do that earlier. Standard heuristic is to inline function when all arguments are applied. For example this means foldlOf' will be inlined into body of sumOf and this gives GHC more opportunities for optimization

Simple benchmark for summong list using sumOf see 8x performance improvement

In addition to `foldlOf'` I applied same transformation to other fold variant. If it's good for one of them it's probably good for all of them. I  also added GHC-style note attempting to explain reason for that. It _is_ weird way of writing code and should be explained.


benchmarks/folds.hs
```diff
@@ -31,6 +31,7 @@ main = defaultMainWith config
       [ bench "native"     $ nf (V.toList . V.indexed) v
       , bench "itraversed" $ nf (itoListOf itraversed) v
       ]
+    , bench "sum" $ whnf (sumOf each) v
```
RyanGlScott, line 34: For consistency with the other benchmarks, can you turn this into a `bgroup "sum"` that also includes benchmarks for `whnf sum v`? (And similarly for the other `sumOf` benchmarks.)

src/Control/Lens/Fold.hs
```diff
@@ -1902,10 +1909,26 @@ foldrMOf l f z0 xs = foldlOf l f' return xs z0
...
+foldlMOf l f z0 = \xs -> foldrOf l f' return xs z0
   where f' x k z = f z x >>= k
 {-# INLINE foldlMOf #-}
 
+-- NOTE: [Inlining and arity]
+-- ~~~~~~~~~~~~~~~~~~~~~~~~~~
+--
+-- GHC uses following inlining heuristic. Function body is inlined if
+-- all its arguments on LHS are applied. So following two definitions
+-- are not equivalent from inliner PoV:
```
RyanGlScott, line 1922:
```suggestion
-- GHC uses the following inlining heuristic: a function body is inlined if
-- all its arguments on the LHS are applied. So the following two definitions
-- are not equivalent from the inliner's PoV:
```

src/Control/Lens/Fold.hs
```diff
@@ -1902,10 +1909,26 @@ foldrMOf l f z0 xs = foldlOf l f' return xs z0
...
+--
+-- > foldlOf' l f z0 xs = ...
+-- > foldlOf' l f z0 = \xs -> ...
+--
+-- GHC will be less eager to inline first one and this results in
+-- worse code. Simple list summation using `sumOf` will be 8x slower
+-- with first version.
```
RyanGlScott, line 1929:
```suggestion
-- GHC will be less eager to inline the first one and this results in
-- worse code. For example, a simple list summation using `sumOf` will be 8x slower
-- with the first version.
```

RyanGlScott (commented):
Thanks!

Please make sure to mention https://github.com/ekmett/lens/issues/1084 somewhere the commit message so that the commit is linked to the issue.

Shimuuar:
I updated benchmarks as you suggested. Thank you for quick review!

P.S. curiously lens variant outperforms native sum for boxed vector. But that looks like a vector's problem.

phadej:
FWIW, this is not eta reduction (nor expansion). It's just wiggling argument around *just for* simplifier (I don't think anything else than GHC inliner cares whether we have `foo x = ...` or `foo = \x -> ...`; please correct me if I'm wrong here).

RyanGlScott:
Good point, @phadej. I agree that "eta reduce" isn't the right terminology here, although it only really impacts the commit title—the Note uses the phrase "arity", which I believe is a more accurate description of what actually matters with respect to inlining.

With that in mind, what if we just changed the commit title to "Reduce arity of foldlOf' and related functions"?

Shimuuar:
Done

RyanGlScott (approved):
Thanks again!

Issue #590, opened 2015-08-09
Benchmark moving the arguments to the right 


ekmett:
#551 looks interesting, but it'd be a much easier sell if we had some benchmarks showing it actually helped in some situations. It makes the code a fair bit harder to read, so it'd be nice to know we're getting something for all that effort.

If someone wanted to contributed benchmarks or dumped core/stg here that either confirmed or denied that it could lead to performance wins that would be incredibly helpful.

shachaf:
For what it's worth, the original motivation for #183 was a few cases where this transformation made a difference in the Core. I don't remember the specific cases anymore, but I don't think it should just be applied indiscriminately.

ekmett:
Which leaves us with precisely no movement forward on the issue.

It isn't clear where to draw the line.

An "always do it" transformation is easy to apply. The "never do it" status quo is even easier to maintain.

PR #1103, opened 2026-05-30
Reduce arity of set, partsOf, both, and related combinators

### LINKED ISSUE #183: Move arguments to the right side of the =?

jozanek:
Closes #183 and #590

Move commonly-under-applied trailing arguments right of the `=` so the `INLINE` fires at under-applied call sites, as was done for the strict folds in 5.3.4 (see `NOTE: [Inlining and arity]` in `Control.Lens.Fold`). A loop over a partially-applied `set _1` is ~19x faster; the rest are neutral but keep each family consistent. Adds `benchmarks/arity.hs`.

### Which combinators changed, and why

The transform only changes runtime behaviour when a combinator is *held partially applied* (stored/reused below its full arity), so it's applied where both of these hold:
1. the combinator is plausibly held partially applied in client code, and
2. the transform is strictness-neutral — moving the argument forces nothing earlier than before.

When one function in a family changed, its same-shape siblings changed too, so no family is left half-transformed:
- `set`, `set'` — the motivating case (#183/#590).
- `partsOf`, `partsOf'`, `unsafePartsOf`, `unsafePartsOf'` - the `let b = l sell s` is still only forced once `s` arrives.
- `mapAccumLOf`, `imapAccumLOf`.
- `both`, `both1` — used first-class as optic values; the arity-0 form lets the optic inline at the consumer.
- `auf` — brought in line with its siblings `au`/`under`.

Already-point-free siblings (`mapAccumROf`, `imapAccumROf`) and already-arity-1 indexed variants (`ipartsOf`/`iunsafePartsOf`, … via `conjoined`) need no change.

### One non-neutral case, called out

`auf` is the exception to (2): like `au`/`under` it is now strict in the iso (`auf undefined` diverges instead of returning a function) in exchange for decomposing the iso once. This is documented as an intentional tradeoff at the definition.

### DIFF
--- a/benchmarks/arity.hs
+++ b/benchmarks/arity.hs
@@ -10,7 +10,7 @@
 -- saturates it, the combinator inlines, and worker/wrapper fuses the hot loop.
 -- See NOTE: [Inlining and arity] in Control.Lens.Fold.
 --
--- Coverage mirrors the change set: set/set', partsOf, mapAccumLOf, both/both1,
+-- Coverage includes set/set', partsOf, mapAccumLOf, both/both1,
 -- and auf. (mapAccumLOf is exercised both directly and via scanl1Of.)
 module Main (main) where
 
--- a/benchmarks/arity.hs
+++ b/benchmarks/arity.hs
@@ -1,0 +1,58 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# OPTIONS_GHC -O2 #-}
+-- Benchmark for issue #183: combinators used *under-applied*.
+--
+-- The setter wrappers are NOINLINE so the partially-applied form (`set _1`)
+-- survives as a real binding. With `set` at arity 2 it cannot inline into the
+-- wrapper and each application pays for the `Identity` newtype path; with `set`
+-- at arity 1 the wrapper saturates it, `set` inlines, and worker/wrapper fuses
+-- it to a tight unboxed-tuple worker. See NOTE: [Inlining and arity].
+module Main (main) where
+
+import Control.Lens
+import Criterion.Main
+
+-- `set _1` held under-applied at the definition site.
+setFst :: Int -> (Int, Int) -> (Int, Int)
+setFst = set _1
+{-# NOINLINE setFst #-}
+
+set'Snd :: Int -> (Int, Int) -> (Int, Int)
+set'Snd = set' _2
+{-# NOINLINE set'Snd #-}
+
+-- `partsOf traverse` held under-applied (arity 3 -> 1).
+reverseParts :: [Int] -> [Int]
+reverseParts = over (partsOf traverse) reverse
+{-# NOINLINE reverseParts #-}
+
+-- scanl1Of drives mapAccumLOf under-applied (arity 4 -> 3).
+scanSum :: [Int] -> [Int]
+scanSum = scanl1Of traverse (+)
+{-# NOINLINE scanSum #-}
+
+-- Hammer the under-applied setter in a tight loop.
+loopSet :: (Int -> (Int, Int) -> (Int, Int)) -> Int -> Int
+loopSet f = \n ->
+  let go !i !p@(a, b)
+        | i >= n    = a + b
+        | otherwise = go (i + 1) (f i p)
+  in go 0 (0, 0)
+{-# INLINE loopSet #-}
+
+xs :: [Int]
+xs = [1 .. 100]
+{-# NOINLINE xs #-}
+
+main :: IO ()
+main = defaultMain
+  [ bgroup "set-underapplied"
+    [ bench "set _1"  $ whnf (loopSet setFst)  1000000
+    , bench "set' _2" $ whnf (loopSet set'Snd) 1000000
+    ]
+  , bgroup "traversal-underapplied"
+    [ bench "partsOf traverse" $ nf (\v -> sum (concatMap reverseParts (replicate 1000 v))) xs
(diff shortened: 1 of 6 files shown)


jozanek:
@ekmett sorry for tagging, who shall I contact for review of my PRs?

RyanGlScott:
Sorry for the delay on this, @jozanek. I'd normally be reviewing these PRs, but I've been away on vacation for a while.

I recently got back, so I'm hoping to have a chance to catch up on my review backlog soon. Do note that it will take some time, however---there are 10 (!) pull requests to go through, and some of them are pretty substantial. I'll do my best to review one or two at a time over the coming days.

benchmarks/arity.hs
```diff
@@ -0,0 +1,58 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# OPTIONS_GHC -O2 #-}
+-- Benchmark for issue #183: combinators used *under-applied*.
+--
+-- The setter wrappers are NOINLINE so the partially-applied form (`set _1`)
+-- survives as a real binding. With `set` at arity 2 it cannot inline into the
+-- wrapper and each application pays for the `Identity` newtype path; with `set`
+-- at arity 1 the wrapper saturates it, `set` inlines, and worker/wrapper fuses
+-- it to a tight unboxed-tuple worker. See NOTE: [Inlining and arity].
```
RyanGlScott: To make it clear where the note lives:

```suggestion
-- it to a tight unboxed-tuple worker. See NOTE: [Inlining and arity] in Control.Lens.Fold.
```

benchmarks/arity.hs
```diff
@@ -0,0 +1,58 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# OPTIONS_GHC -O2 #-}
```
RyanGlScott: This is redundant, as the `.cabal` file already enables `-O2`.

RyanGlScott (approved):
Two minor suggestions inline, but otherwise LGTM.

benchmarks/arity.hs
jozanek: updated

benchmarks/arity.hs
jozanek: updated

jozanek:
@RyanGlScott no worries! I admit I was a bit impatient 😊 Learning Haskell is fun and I am looking for a job so have a lot of time, let's see how many bugs I can squash in the meantime.

N.B. Added few more enhancements (see 3 last commits).

src/Control/Lens/Iso.hs
```diff
@@ -209,7 +209,14 @@ au k = withIso k $ \ sa bt f -> fmap sa (f bt)
 -- with the behavior of 'au'. For the old behavior use 'xplatf' or for a version that is compatible
 -- across both old and new versions of @lens@ you can just use 'coerce'!
 auf :: (Functor f, Functor g) => AnIso s t a b -> (f t -> g s) -> f b -> g a
-auf k ftgs fb = withIso k $ \sa bt -> sa <$> ftgs (bt <$> fb)
+-- Pushing the remaining arguments inside 'withIso' (so only @k@ is to the left
+-- of the @=@) matches 'au'/'under': the iso is decomposed once, when @auf k@ is
+-- forced, and the body inlines at partial applications. Unlike the other arity
+-- reductions in this change this is *not* strictness-neutral -- as with
+-- 'au'/'under', @auf@ is now strict in the iso, so @auf undefined@ diverges
```
RyanGlScott: A minor thing, but I generally prefer the documentation to describe the _current_ state of the code rather than talking about how it used to be implemented. Unless you follow the commit history of `lens` closely, you aren't going to have context for what "this change" and "now" refer to. Consider dropping these phrases:


```suggestion
-- forced, and the body inlines at partial applications. Unlike the other arity
-- reductions, this is *not* strictness-neutral -- as with
-- 'au'/'under', @auf@ is strict in the iso, so @auf undefined@ diverges
```

benchmarks/arity.hs
```diff
@@ -0,0 +1,110 @@
...
+--
+-- Each binding below fixes a combinator's leading arguments and is marked
+-- NOINLINE, so the partially-applied form (`set _1`, `over both (+1)`, ...)
+-- survives as a real binding. With the combinator at full arity it cannot
+-- inline into that binding and each application pays for the newtype/decompose
+-- path; with the trailing arguments moved right of the `=` the binding
+-- saturates it, the combinator inlines, and worker/wrapper fuses the hot loop.
+-- See NOTE: [Inlining and arity] in Control.Lens.Fold.
+--
+-- Coverage mirrors the change set: set/set', partsOf, mapAccumLOf, both/both1,
```
RyanGlScott: Similarly, avoid using the phrase "the change set" here. Perhaps:


```suggestion
-- Coverage includes set/set', partsOf, mapAccumLOf, both/both1,
```

jozanek: done

src/Control/Lens/Iso.hs
jozanek: updated

RyanGlScott (approved):

PR #1104, opened 2026-05-31
Add ReifiedFoo vs AFoo benchmark

### LINKED ISSUE #217: Benchmarks for Control.Lens.Reified would be nice
We might be able to remove a lot of the `ReifiedFoo` types in favour of `AFoo`, which compose and should be almost as fast. It'd be nice to have some benchmarks for this.

jozanek:
Answers/Closes #217

Adds benchmarks/reified.hs and a `benchmark reified` stanza, answering the long-standing question in #217:  Both can be stored in a container, so each optic is compared three ways -- Reified (runX), AFoo recovered via cloneX, and AFoo through a direct consumer (^#, over, withIso, withPrism, view, review).

Result: the AFoo path is as fast or much faster everywhere except Lens (~10%) and Traversal (~3%), which marginally favor Reified -- e.g. Review ~19x, Prism ~6.6x, Iso ~2.7x, Fold/Getter ~2.5x, Setter ~2x. Unwrapping a Reified yields a polymorphic optic that can't be specialized across the NOINLINE boundary, so it pays runtime dictionary dispatch; the AFoo alias is pinned to one functor. The case for keeping ReifiedFold/Getter/Review is their combining instances and result polymorphism, not speed.

### DIFF
--- a/benchmarks/reified.hs
+++ b/benchmarks/reified.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE BangPatterns #-}
-{-# OPTIONS_GHC -O2 #-}
--- Benchmark for ReifiedFoo newtypes be replaced by the monomorphic AFoo aliases.
+-- Benchmark for ReifiedFoo newtypes being replaced by the monomorphic AFoo aliases.
 -- Both can be stored in a container; this measures whether storing an AFoo
 -- (then recovering an optic via clone*, or using a dedicated A* consumer) costs
 -- the same as storing a ReifiedFoo and unwrapping. Each stored optic is NOINLINE
--- a/benchmarks/reified.hs
+++ b/benchmarks/reified.hs
@@ -1,0 +1,123 @@
+{-# LANGUAGE BangPatterns #-}
+{-# OPTIONS_GHC -O2 #-}
+-- Benchmark for ReifiedFoo newtypes be replaced by the monomorphic AFoo aliases.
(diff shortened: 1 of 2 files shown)


benchmarks/reified.hs
```diff
@@ -0,0 +1,123 @@
+{-# LANGUAGE BangPatterns #-}
+{-# OPTIONS_GHC -O2 #-}
+-- Benchmark for ReifiedFoo newtypes be replaced by the monomorphic AFoo aliases.
```
RyanGlScott:
```suggestion
-- Benchmark for ReifiedFoo newtypes being replaced by the monomorphic AFoo aliases.
```

benchmarks/reified.hs
```diff
@@ -0,0 +1,123 @@
+{-# LANGUAGE BangPatterns #-}
+{-# OPTIONS_GHC -O2 #-}
```
RyanGlScott: This is redundant, as `-O2` is specified in the `.cabal` file.

RyanGlScott (commented):
LGTM, aside from two minor suggestions.

benchmarks/reified.hs
jozanek: fixed

benchmarks/reified.hs
jozanek: udpated

RyanGlScott (approved):

PR #1105, opened 2026-05-31
Space leaks

### LINKED ISSUE #233: Space leaks
Various traversals in lens have a space leak problem. It seems to be related to GHC generating tail-recursive functions that build up big case-analysis thunks which they pass to themselves.

Some examples:
- `lastOf (traverse.traverse) :: [Maybe a] -> Maybe a`
- `almostAnythingOf bytes :: Strict.ByteString -> ...`

For the second one, note that e.g. `foo b = lastOf traverse (unpack b)` has the problem, whereas `foo b = lastOf folded (unpack b)` doesn't. This seems to be related to a RULE firing.

jozanek:
Closes #233 

### Problem

`lastOf bytes` (and other `Rightmost` folds) over a **lazy** `ByteString` retains memory proportional to the number of chunks instead of running in constant space. This is the lazy-`ByteString` case that #233 was reopened to track. It still reproduces on GHC 9.14.1:

```haskell
lastOf bytes (BL.replicate n 7)   -- peak residency grows ~linearly with n
```

### Root cause

The rewrite rules on `traversedLazy`/`traversedLazy8` route lazy-`ByteString` folds to a streaming `BL.foldr`-based implementation only at type `Getting (Endo r)` — i.e. the right-fold-shaped getters (`lengthOf`, `sumOf`, `toListOf`, …), which therefore already stream. `lastOf` uses the `Rightmost` monoid, which is **not** `Endo`-shaped, so it misses the rule and falls through to the general `traversedLazy`, whose `foldrChunks`/`<*>` builds a lazy thunk chain over the chunk spine.

(Lazy `Text` is unaffected: its `text` traversal routes through `unpacked . traversed`, i.e. the list traversal, which streams since GHC #7556.)

### Fix

Add a `Rightmost` rule mirroring the existing `Endo` one, dispatching `lastOf`-style folds to a strict `BL.foldl'`:

```haskell
"lastOf lazy bytestring"
  traversedLazy = lazyBytesRightmost :: Getting (Rightmost a) BL.ByteString Word8;
```

One subtlety: `Rightmost`'s `(<>)` always wraps its result in `RStep <thunk>`, and the thunk closes over the accumulator, so a naïve `foldl' (<>)` is actually *worse* (an O(n) thunk chain). Forcing the `RStep` payload one extra level drops that reference, giving constant space (`stepRightmost`).

**Soundness:** `Rightmost`'s `(<>)` is *not* value-level associative (left- vs right-nesting insert different numbers of `RStep` layers), so the strict left fold does not literally rebuild the same `Rightmost` value as the lawful right fold. It is sound nonetheless: the only consumer of `Getting (Rightmost a)` is `lastOf`, which observes the result solely through `getRightmost` — and `getRightmost` is invariant under re-association and extra `RStep` layers. The `seq` only forces, so the rewrite agrees with the lawful traversal on every observation `lastOf` can make. Short-circuiting folds (`firstOf`/`anyOf`, i.e. `Leftmost`/`Any`/`All`) are deliberately **not** rerouted: their laziness is load-bearing and they already stream by stopping early.

### Results (GHC 9.14.1, `lastOf bytes` over a lazy `ByteString`)

| input | before | after |
|---|---|---|
| 40 MB | ~100 KB | ~45 KB |
| 160 MB | ~1.1 MB | ~45 KB |
| 640 MB | ~4.2 MB | ~45 KB |

i.e. O(chunks) → O(1). `lengthOf`/`firstOf`/`anyOf` and lazy `Text` are unchanged (already constant-space).

### Regression test (`tests/space.hs`)

New `space` test suite that folds a lazy `ByteString` with `lastOf bytes` at two sizes (8k vs 128k chunks) and asserts peak residency (`GHC.Stats.max_live_bytes`) doesn't scale. Details:

- The input is built with `BL.fromChunks` from fixed-size 4 KiB chunks, so the chunk count — the quantity the leak actually scales with — is exact and independent of bytestring's internal chunking heuristics (which have varied across releases).
- The leak's peak is transient, so reliable sampling needs frequent GCs: the cabal stanza bakes in `-with-rtsopts=-T -A256k` (`-T` enables `GHC.Stats`, the small nursery makes GCs frequent enough to sample the peak).
- Because the suite would silently measure nothing if those options went missing, it verifies at startup (via `GHC.Stats`/`GHC.RTS.Flags`) that they are actually in effect and fails loudly otherwise. This also catches the case where an incremental build skips the relink after a link-flag-only change, leaving stale baked options.
- Validated both ways: with the fix, residency growth is **~1 KB** → PASS; with the fix reverted, **~4 MB** → FAIL.
- Behind a `test-space` cabal flag (default on), like the other suites, and gated to GHC ≥ 8.2 (where `GHC.Stats.getRTSStats` exists); depends on `base`, `bytestring`, `lens`.

### Caveats

- Targeted at `Rightmost` (the one strict-to-end built-in fold not covered by the `Endo` rule, i.e. `lastOf`). A user folding a lazy `ByteString` via `foldMapOf` with their *own* strict-to-end lazy monoid would still hit the general path; the built-in surface is covered.

### DIFF
--- a/src/Control/Lens/Internal/ByteString.hs
+++ b/src/Control/Lens/Internal/ByteString.hs
@@ -32,6 +32,7 @@ import Prelude ()
 import Control.Lens.Type
 import Control.Lens.Getter
 import Control.Lens.Fold
+import Control.Lens.Internal.Fold (Rightmost(..))
 import Control.Lens.Indexed
 import Control.Lens.Internal.Prelude
 import Control.Lens.Setter
@@ -140,7 +141,33 @@ traversedLazy pafb = \lbs -> BL.foldrChunks go (\_ -> pure BL.empty) lbs 0
     traversedLazy = foldring BL.foldr :: Getting (Endo r) BL.ByteString Word8;
   "igets lazy bytestring"
     traversedLazy = ifoldring ifoldrBL :: IndexedGetting Int (Endo r) BL.ByteString Word8;
+  "lastOf lazy bytestring"
+    traversedLazy = lazyBytesRightmost :: Getting (Rightmost a) BL.ByteString Word8;
  #-}
+
+-- | 'lastOf' and other strict-to-end 'Rightmost' folds are not covered by the
+-- @Endo@ rule above, so without this they fall through to the general
+-- 'traversedLazy', whose @foldrChunks@/@<*>@ builds a lazy thunk chain over the
+-- chunk spine and retains memory proportional to the chunk count.
+--
+-- 'Rightmost' is strict-to-end (unlike the short-circuiting 'Leftmost'/'Any'),
+-- so a strict 'BL.foldl'' is sound: @acc <> 'RLeaf' w@ discards @acc@, so the
+-- fold runs in constant space. We deliberately do /not/ do this for
+-- 'firstOf'/'anyOf', whose laziness is load-bearing.
+lazyBytesRightmost :: Getting (Rightmost a) BL.ByteString Word8
+lazyBytesRightmost f = Const #. BL.foldl' (\acc w -> stepRightmost acc (getConst (f w))) mempty
+{-# INLINE lazyBytesRightmost #-}
+
+-- | One step of a strict 'Rightmost' left fold. @(<>)@ for 'Rightmost' always
+-- wraps its result in 'RStep' and is lazy in the left argument, so a plain
+-- @foldl' (<>)@ retains an O(n) thunk chain (each thunk closing over the prior
+-- accumulator). Forcing the 'RStep' payload one extra level drops that
+-- reference, giving constant space. This does not change the value: @foldl' (<>)
+-- mempty@ equals the lawful @foldr (<>) mempty@ by associativity, and the 'seq'
+-- only forces, so the rewrite stays sound.
+stepRightmost :: Rightmost a -> Rightmost a -> Rightmost a
+stepRightmost acc r = case acc <> r of s@(RStep x) -> x `seq` s; s -> s
+{-# INLINE stepRightmost #-}
 
 imapBL :: (Int -> Word8 -> Word8) -> BL.ByteString -> BL.ByteString
 imapBL f = snd . BL.mapAccumL (\i a -> i `seq` (i + 1, f i a)) 0
@@ -171,7 +198,14 @@ traversedLazy8 pafb = \lbs -> BL.foldrChunks go (\_ -> pure BL.empty) lbs 0
     traversedLazy8 = foldring BL8.foldr :: Getting (Endo r) BL8.ByteString Char;
   "igets lazy bytestring"
     traversedLazy8 = ifoldring ifoldrBL8 :: IndexedGetting Int (Endo r) BL8.ByteString Char;
+  "lastOf lazy bytestring"
+    traversedLazy8 = lazyBytesRightmost8 :: Getting (Rightmost a) BL8.ByteString Char;
  #-}
+
+-- | See 'lazyBytesRightmost'; this is the @Char@ analogue for 'traversedLazy8'.
+lazyBytesRightmost8 :: Getting (Rightmost a) BL8.ByteString Char
+lazyBytesRightmost8 f = Const #. BL8.foldl' (\acc c -> stepRightmost acc (getConst (f c))) mempty
+{-# INLINE lazyBytesRightmost8 #-}
 
 imapBL8 :: (Int -> Char -> Char) -> BL8.ByteString -> BL8.ByteString
 imapBL8 f = snd . BL8.mapAccumL (\i a -> i `seq` (i + 1, f i a)) 0
--- a/src/Control/Lens/Internal/ByteString.hs
+++ b/src/Control/Lens/Internal/ByteString.hs
@@ -159,12 +159,19 @@ lazyBytesRightmost f = Const #. BL.foldl' (\acc w -> stepRightmost acc (getConst
 {-# INLINE lazyBytesRightmost #-}
 
 -- | One step of a strict 'Rightmost' left fold. @(<>)@ for 'Rightmost' always
--- wraps its result in 'RStep' and is lazy in the left argument, so a plain
--- @foldl' (<>)@ retains an O(n) thunk chain (each thunk closing over the prior
--- accumulator). Forcing the 'RStep' payload one extra level drops that
--- reference, giving constant space. This does not change the value: @foldl' (<>)
--- mempty@ equals the lawful @foldr (<>) mempty@ by associativity, and the 'seq'
--- only forces, so the rewrite stays sound.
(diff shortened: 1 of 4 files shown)


tests/space.hs
```diff
@@ -0,0 +1,74 @@
...
+import           Data.ByteString.Lazy.Lens (bytes)
+import           GHC.Stats (getRTSStats, getRTSStatsEnabled, max_live_bytes)
+import           System.Environment (getExecutablePath, lookupEnv, setEnv)
+import           System.Exit (exitFailure, exitWith)
+import           System.Mem (performMajorGC)
+import           System.Process (rawSystem)
+import           Text.Printf (printf)
+
+reexecVar :: String
+reexecVar = "LENS_SPACE_TEST_REEXEC"
```
RyanGlScott: Please document somewhere what this environment variable is doing.

src/Control/Lens/Internal/ByteString.hs
```diff
@@ -140,8 +141,34 @@ traversedLazy pafb = \lbs -> BL.foldrChunks go (\_ -> pure BL.empty) lbs 0
...
+
+-- | One step of a strict 'Rightmost' left fold. @(<>)@ for 'Rightmost' always
+-- wraps its result in 'RStep' and is lazy in the left argument, so a plain
+-- @foldl' (<>)@ retains an O(n) thunk chain (each thunk closing over the prior
+-- accumulator). Forcing the 'RStep' payload one extra level drops that
+-- reference, giving constant space. This does not change the value: @foldl' (<>)
+-- mempty@ equals the lawful @foldr (<>) mempty@ by associativity, and the 'seq'
+-- only forces, so the rewrite stays sound.
+stepRightmost :: Rightmost a -> Rightmost a -> Rightmost a
+stepRightmost acc r = case acc <> r of s@(RStep x) -> x `seq` s; s -> s
```
RyanGlScott: Minor style nitpick: rather than using semicolons here, I'd prefer to separate these cases using whitespace:


```suggestion
stepRightmost acc r = case acc <> r of
                        s@(RStep x) -> x `seq` s
                        s -> s
```

tests/space.hs
```diff
@@ -0,0 +1,74 @@
...
+
+main :: IO ()
+main = do
+  reentered <- lookupEnv reexecVar
+  case reentered of
+    Just _  -> runTest
+    Nothing -> do
+      setEnv reexecVar "1"
+      setEnv "GHCRTS" "-T -A256k"
```
RyanGlScott: If the user sets the `GHCRTS` environment variable themselves, then this will override those settings, correct?

As an alternative, I wonder if we should instead pass `+RTS -T -A256k -RTS` as arguments to the executable. This should avoid needing to muck around with environment variables like `GHCRTS`.

tests/space.hs
```diff
@@ -0,0 +1,74 @@
...
+-- 'lastOf' over a lazy 'ByteString' must run in /constant/ space. A leaking
+-- traversal retains roughly one thunk per chunk, so peak residency grows with
+-- the input length. We fold a lazily-generated 'ByteString' at two sizes (16x
+-- apart) and assert that peak residency (from "GHC.Stats") does not grow: a
+-- streaming fold stays flat, a leaking one jumps by megabytes.
+--
+-- The peak is transient (it lives only during the fold), so it is only sampled
+-- if GCs are frequent enough; we therefore need a small nursery (@-A256k@) plus
+-- @-T@ to enable "GHC.Stats". Baking a multi-word @-with-rtsopts@ through cabal
+-- is unreliable, so on first entry we set @GHCRTS@ and re-exec ourselves -- the
```
RyanGlScott: Out of curiosity, in what sense is `-with-rtsopts` unreliable? Is there an upstream issue in `cabal`'s issue tracker about this?

RyanGlScott (commented):
Thanks for fixing this long-standing issue!

tests/space.hs
```diff
@@ -0,0 +1,74 @@
...
+--
+-- 'lastOf' over a lazy 'ByteString' must run in /constant/ space. A leaking
+-- traversal retains roughly one thunk per chunk, so peak residency grows with
+-- the input length. We fold a lazily-generated 'ByteString' at two sizes (16x
+-- apart) and assert that peak residency (from "GHC.Stats") does not grow: a
+-- streaming fold stays flat, a leaking one jumps by megabytes.
+--
+-- The peak is transient (it lives only during the fold), so it is only sampled
+-- if GCs are frequent enough; we therefore need a small nursery (@-A256k@) plus
+-- @-T@ to enable "GHC.Stats". Baking a multi-word @-with-rtsopts@ through cabal
```
phadej: Is there a cabal issue, please open if there aren't and reference.

src/Control/Lens/Internal/ByteString.hs
jozanek: done

tests/space.hs
jozanek: This is gone entirely now (see the other thread)

tests/space.hs
jozanek: Yes, `setEnv "GHCRTS"` would have clobbered a user-set `GHCRTS`. Switched to your suggestion: the test now re-execs itself passing `+RTS -T -A256k -RTS` as plain arguments, so the environment is never touched. The child is recognized by a sentinel argument (the RTS strips the `+RTS ... -RTS` block from its argv).

tests/space.hs
jozanek: I tried to substantiate the claim and couldn't. It was wrong, so I removed it along with the re-exec machinery (back to a plain baked `-with-rtsopts`, `process` dependency gone).

What actually bit me: GHC doesn't relink when only link-time flags change, so my edit-test loop kept running a binary with stale baked RTS options. The quoting works fine. Since the suite silently measures nothing if the options go missing, the test now verifies at startup (via `GHC.Stats`/`GHC.RTS.Flags`) that they're in effect, and fails loudly otherwise.

RyanGlScott (approved):
Looks great! Could you resolve the merge conflicts?

tests/space.hs
jozanek: Sorry for the confusion — there's no cabal quoting bug.

What I was seeing was GHC not relinking after a -with-rtsopts-only change in the cabal file: GHC's recompilation checker doesn't track link-time flags, so with no source changes the old binary (without the baked RTS options) was reused. cabal build still reported "Building" but produced nothing new.

The cabal stanza -rtsopts "-with-rtsopts=-T -A256k" works correctly — a clean build produces a binary with those options in effect. To guard against the stale-binary scenario (and against anyone accidentally removing the stanza in the future), the test now verifies at startup via GHC.RTS.Flags.getGCFlags and GHC.Stats.getRTSStatsEnabled that the options are actually active, and fails loudly with a diagnostic if they're not. So there's no re-exec, no process dependency, and no cabal issue to file.

PR #1106, opened 2026-06-01
Fix `upon` looping forever when nested via per-probe exception tokens

### LINKED ISSUE #235: `upon` needs more magic.
```
[10:33] shachaf:	 > [1..10] & (upon.view.upon) tail %~ reverse
[10:33] lambdabot:	  mueval-core: Time limit exceeded
```

jozanek:
Fixes #235.

## Root cause

`upon`'s "magic" runs through `lookupon`, which booby-traps the structure so each candidate target throws a `FieldException` carrying the index it read, runs the user's accessor, and catches that exception to learn which field was touched. The exception was distinguished only by its focus type:

```haskell
data FieldException a = FieldException !Int a
instance Typeable a => Exception (FieldException a)
```

When two `upon` layers share a focus type (here both `[Int]`), the inner layer's handler catches the **outer** layer's exception and mistakes it for its own. Its recursive search never reaches its base case, so it spins forever.

## Fix

Give each probe a per-invocation `Unique` token, and make `FieldException` existential so a handler can catch any field exception and *then* decide ownership:

```haskell
data FieldException = forall a. Typeable a => FieldException !Unique !Int a
```

`lookupon` now dispatches on the token: own token → recover the payload (via `cast`); foreign token → re-`throw` so it bubbles to the layer that owns it (this breaks the loop); not a `FieldException` → `Nothing`, preserving partial-accessor handling such as `"" ^? upon tail == Nothing`.

The token is minted once per `upon`/`upon'`/`onceUpon`/`onceUpon'` call by a `NOINLINE`, structure-keyed helper and threaded through. Minting it inside the `INLINE` `lookupon` would let `-O2` CSE share a single token across layers and silently re-break the fix.

### DIFF
--- a/src/Data/Data/Lens.hs
+++ b/src/Data/Data/Lens.hs
@@ -61,6 +61,7 @@ import qualified Data.HashSet as S
 import           Data.HashSet (HashSet)
 import           Data.IORef
 import           Data.Monoid
+import           Data.Unique (Unique, newUnique)
 import           GHC.Exts (realWorld#)
 import           Prelude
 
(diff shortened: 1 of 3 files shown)


src/Data/Data/Lens.hs
```diff
@@ -136,20 +137,35 @@ biplate = biplateData (fromOracle answer) where
...
+instance Exception FieldException
 
-lookupon :: Typeable a => LensLike' (Indexing Identity) s a -> (s -> a) -> s -> Maybe (Int, Context a a s)
-lookupon l field s = case unsafePerformIO $ E.try $ evaluate $ field $ s & indexing l %@~ \i (a::a) -> E.throw (FieldException i a) of
+-- | A fresh identity per 'upon' invocation. Keyed on the structure, 'NOINLINE',
+-- and relying on the module's @-fno-full-laziness@ so GHC cannot share one token
+-- across invocations (which would defeat the nesting fix).
+newFieldToken :: s -> Unique
+newFieldToken s = s `seq` unsafePerformIO newUnique
+{-# NOINLINE newFieldToken #-}
```
RyanGlScott: Wow, this is magical indeed. I almost think that we should name this function "`unsafeNewFieldToken`" so that it's more obvious that this function is (intentionally) not referentially transparent.

tests/hunit.hs
```diff
@@ -436,4 +446,9 @@ main = defaultMain $
   , testCase "correct indexing lazy text" case_correct_indexing_lazy_text
   , testCase "correct indexing strict bytestring" case_correct_indexing_strict_bytestring
   , testCase "correct indexing lazy bytestring" case_correct_indexing_lazy_bytestring
+    -- timeout so a future non-termination regression fails fast instead of hanging CI
```
RyanGlScott: Can you say what the timeout here is in seconds/milliseconds/whatever the appropriate unit of time is? I am always a bit worried about applying blanket timeouts, since different machines will run this slower or faster depending on their specs. If we ever need to increase the timeout window to support slower machines, it will be helpful to know exactly what the timeout is.

src/Data/Data/Lens.hs
jozanek: Good idea — renamed to `unsafeNewFieldToken` and the haddock now calls out the intentional non-referential-transparency.

tests/hunit.hs
jozanek: It's 10 seconds (`mkTimeout` takes microseconds) — now a named `uponTimeout` binding with a comment, so there's one obvious place to bump if slower machines need headroom.

RyanGlScott (approved):

RyanGlScott:
Can you resolve the merge conflicts?

jozanek:
@RyanGlScott merged

PR #1107, opened 2026-06-01
Add lens-build: type-level completeness-checked record builder

### LINKED ISSUE #286: QuasiQuoter for record construction
So lenses currently suck a bit for constructing record values from scratch because you get no compile-time verification that all fields are set, which leaves unset fields as bottoms.

Currently the better options are to either construct using the constructor function, or the record notation with the underscored field names, but I don't like either of those options in every case and the latter will not work if we implement and use #283.

My straw man proposal then is something like this:

``` haskell
data Foo = Foo { _a, _b :: Int }
makeLenses ''Foo

-- These are equivalent
foo = [record| Foo { a = 1, b = 2 } |]
foo' = Foo {} & a .~ 1 & b .~ 2
foo'' = Foo { _a = 1, _b = 2 }

-- These warn or error at compile time
bar = [record| Foo { a = 1 } |]
bar'' = Foo { _a = 1 }  -- at least with -Wall

-- But this doesn't, which is the main problem
bar' = Foo {} & a .~ 1
```

Sadly we can't use expression quotes for this, because you can't quote invalid record constructions:

``` haskell
>>> [| Just { foo = "bar" } |]

<interactive>:4:11:
    `foo' is not a (visible) field of constructor `Just'
```

Which means we'd have to use `QuasiQuotes` for this and that we have to depend on `haskell-src-meta`.  Is that acceptable or should this be a separate package?

Other issues:
- I don't know for sure if TH is powerful enough to take a constructor by string name and reify it to find all its record fields
- We'd have to have a way to map record fields to lenses which is easy assuming people use the `defaultLensRules`, otherwise we could perhaps have a variant that takes the lens rules as an argument
  - (We need this to error at compile-time if any field is left out)
- If the record has non-lensed fields we need to set those with normal record notation, like `Foo { a = 1 } & b .~ 2`

Please critique and suggest alternative solutions to the problem stated at the top!

jozanek:
Closes #286 

## How

- `Builder s (set :: [Symbol])` carries the set of assigned fields in a phantom type-level list. `build` only type-checks once a closed type family confirms `set` covers every field; otherwise it reduces to a hand-written `TypeError`.
- `makeBuild` (Template Haskell) reifies a record via `th-abstraction` and emits the `RecordBuilder` / `SetField` instances. It is independent of `makeLenses` - field names use the same `_field -> field` convention, so they line up.

## Scope and limitations

Single-constructor, monomorphic records with lazy fields. GHC >= 8.8 (older compilers are skipped via `buildable: False`).

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -219,6 +219,7 @@ jobs:
           echo "packages: $GITHUB_WORKSPACE/source/." >> cabal.project
           echo "packages: $GITHUB_WORKSPACE/source/./examples" >> cabal.project
           echo "packages: $GITHUB_WORKSPACE/source/./lens-properties" >> cabal.project
+          echo "packages: $GITHUB_WORKSPACE/source/./lens-build" >> cabal.project
           cat cabal.project
       - name: sdist
         run: |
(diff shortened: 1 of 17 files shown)


RyanGlScott (commented):
I'll need some time to look at this more thoroughly, but I like the direction of travel.

A question about testing: I see that there is a test case that demonstrates `lens-build` when it works correctly, which is nice. On the other hand, part of the original motivation for https://github.com/ekmett/lens/issues/286 was to ensure that proper error messages would be raised if you attempt to build a lens for something invalid (e.g., `Foo{}`). Does `lens-build` do this as well? If so, can we test this?

jozanek:
Yes, it does. An incomplete `build` is rejected at compile time with a custom message, e.g. `build: missing field "v"`.

I've added negative tests that prove it, covering both error surfaces:
- the builder itself — an incomplete `build` (`tests/negative/TMissingField.hs`); and
- the `makeBuild` generator's guards on unsupported input — a sum type, a positional (non-record) constructor, and a parameterized type — each of which fails with its own message (e.g. `makeBuild: single-constructor records only`).

`tests/negative/run.sh` (wired up in a small `negative-test.yml` workflow) builds each fixture and asserts both that it fails to compile and that the failure is the expected message. There's also a check confirming a fixture starts passing-as-compiled if you complete it, so the tests can't silently pass.

One implementation note in case it's useful: I went with compile-and-check-stderr rather than the usual `-fdefer-type-errors`/`should-not-typecheck` runtime approach, because `Complete` is a phantom constraint. `build` never forces its dictionary, so a deferred type error would never actually throw at runtime. So I just check that it fails to compile instead.

RyanGlScott:
Ah, thank you for adding tests. But I was thinking of something simpler: if I understand correctly, attempting to construct an invalid record field using `lens-build` will cause an exception to be thrown in the `Q` monad. Could we instead intercept these exceptions using [`recover`](https://hackage-content.haskell.org/package/template-haskell-2.24.0.0/docs/Language-Haskell-TH-Syntax.html#v:recover) to ensure that the error occurs? This seems like it would be simpler than having external scripting for running these tests.

phadej:
Sorry for a possibly stupid question but: Why this needs to be in `lens`? I don't see anything lens/optics specific? It's not **lens**-build, it's (if you will) just "build", isn't it?

EDIT: To expand, I don't think #286 can be fixed. "set of field selectors" is just not a "thing" to construct a record. The `OverloadedRecordDot` in GHC doesn't work with construction syntax **at all**. Dually, "set of prisms" is not enough to make a complete pattern match. The `COMPLETE` pragma in GHC is ad-hoc fix, but IMO it's not principled at all; a hack. And this patch isn't using lenses at all, so if it's merged it's not solving #286.

jozanek:
@phadej You're right, it uses no lenses and it doesn't solve #286 as posed, which is why I fell back to type-level field tracking.

That does reframe the issue, though. Construction is the job of the *dual* optics, which lens already has - you build a sum through a `Prism` (`review _Just 5`). The missing analogue for products is a record <-> field-tuple `Iso`:

```haskell
-- generated by TH (or via generic-lens):
fooIso :: Iso' Foo (Int, Bool)

-- construct through the product iso; completeness is enforced by the tuple type:
foo :: Foo
foo = review fooIso (1, True)        -- or  (1, True) ^. from fooIso
```

This actually *uses* optics, and the "`Foo{}` should warn" motivation falls out for free. Tuple type demands every field. Trade-off: it's positional, so you lose named fields (re-adding names pulls the type-level machinery back in).

So two lens-native directions that would genuinely bear on #286
- TH helper that generates this product `Iso` plus a `review`-style builder around it
- documenting the `review`/`Iso` recipe (and why a set of lenses can't construct) 

Would either be of interest, or would you rather this just live as a standalone package?

PR #1108, opened 2026-06-01
Add homogeneousOf/homogeneous and their aliases sameOf/same

### LINKED ISSUE #685: Add ‘allEqualOf’
Even though it's not in `Data.List` it would be cool to have a `Fold` for testing if all elements are the same. Sketch of an implementation, comment if you have a nicer way of constructing this `Monoid`, maybe with `First` and `All`

```haskell
data Same a = SeenNone | SeenSome a | Nope

instance Eq a => Semigroup (Same a) where
  SeenNone   <> SeenSame a = SeenSame a
  SeenSame a <> SeenNone   = SeenSame a
  SeenNone   <> SeenNone   = SeenNone
  SeenSame a <> SeenSame b
    | a == b = SeenSame a
  _ <> _ = Nope
 
instance Eq a => Monoid (Same a) where
  mempty = SeenNone
  mappend = (<>)

theSame :: Same a -> Bool
theSame = \case
  SeenNone   -> True
  SeenSame{} -> True
  Nope       -> False

-- allEqualOf :: Eq a => Fold s a -> s -> Bool
allEqualOf :: Getting (Same a) s a -> s -> Bool
allEqualOf ℓ = theSame . foldMapOf ℓ SeenSame
```

Other names `allSameOf`, `allEqualOf`, `equalOf`, `allTheSameOf`.

```
>>> __ = error "..."
>>> allEqualOf both (Left __)
True
>>> allEqualOf both ('a', 'a')
True
>>> allEqualOf both ('a', 'b')
False
>>> allEqualOf (folded.folded.both.folded) [Just ("aaaa", "aa"), Nothing, Just ("a", "")]
False
```

jozanek:
Closes #685.

## Summary

Adds a `Fold` consumer that tests whether **all targets are equal to one another** - the optics analogue of `Data.List.Extra.allSame`.

Four exports, all surfaced through the umbrella `Control.Lens`:

| Name | Type | Role |
|------|------|------|
| `homogeneousOf` | `Eq a => Getting (Same a) s a -> s -> Bool` | primary optic consumer |
| `homogeneous` | `(Foldable f, Eq a) => f a -> Bool` | `Foldable` convenience (`= homogeneousOf folded`) |
| `sameOf` | `Eq a => Getting (Same a) s a -> s -> Bool` | short alias for `homogeneousOf` |
| `same` | `(Foldable f, Eq a) => f a -> Bool` | short alias for `homogeneous` |

```haskell
>>> homogeneousOf both ('a','a')
True
>>> homogeneousOf both ('a','b')
False
>>> homogeneous [1,1,1]
True
>>> homogeneousOf both (Left (error "boom") :: Either Int Int)  -- single target never forced
True
```

## Naming

Both stems are single words: `homogeneous` (precise) and `same` (matches FP precedent). The `*Of` forms are the optic consumers; the bare forms are the `Foldable` conveniences @Icelandjack originally requested.

## Implementation

- New `Same` monoid in `Control.Lens.Internal.Fold`, beside `Leftmost`/`Rightmost`: a three-state accumulator `SameEmpty | SameOne a | SameDiffer`.
- The `Eq a =>` `Semigroup` orders `SameDiffer <> _ = SameDiffer` first so a negative result short-circuits and stays lazy in its right argument (per @Zemyla's note in the thread).
- The `Eq a` constraint flows into the consumer signatures, consistent with `elemOf`/`notElemOf`.

### DIFF
--- a/src/Control/Lens/Internal/Fold.hs
+++ b/src/Control/Lens/Internal/Fold.hs
@@ -212,13 +212,13 @@ data Same a
   | SameDiffer  -- ^ at least two targets differed
 
 instance Eq a => Semigroup (Same a) where
-  SameDiffer <> _          = SameDiffer        -- short-circuit, lazy in the right
-  SameEmpty  <> y          = y
-  x          <> SameEmpty  = x
-  SameOne a  <> SameOne b
-    | a == b               = SameOne a
-  _          <> _          = SameDiffer
-  {-# INLINE (<>) #-}
+  SameDiffer    <> _          = SameDiffer       -- short-circuit, lazy in the right
+  SameEmpty     <> y          = y
+  x@(SameOne _) <> SameEmpty  = x
+  SameOne _     <> SameDiffer = SameDiffer
+  x@(SameOne a) <> SameOne b
+    | a == b                  = x                -- reuse x; no new allocation
+    | otherwise               = SameDiffer
 
 instance Eq a => Monoid (Same a) where
   mempty = SameEmpty
@@ -232,4 +232,3 @@ instance Eq a => Monoid (Same a) where
 getSame :: Same a -> Bool
 getSame SameDiffer = False
 getSame _          = True
-{-# INLINE getSame #-}
--- a/src/Control/Lens/Internal/Fold.hs
+++ b/src/Control/Lens/Internal/Fold.hs
@@ -23,6 +23,7 @@ module Control.Lens.Internal.Fold
   , Sequenced(..)
   , Leftmost(..), getLeftmost
   , Rightmost(..), getRightmost
+  , Same(..), getSame
   , ReifiedMonoid(..)
   -- * Semigroups for folding
   , NonEmptyDList(..)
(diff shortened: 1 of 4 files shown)


RyanGlScott (commented):
Thanks! I suppose my only question is whether we want to have both `homogeneous*` and `same*` names for the same functions. Personally, I find `same*` to be more concise and easier to read, so my inclination would be to remove the `homogeneous*` names in favor of just the `same*` ones.

jozanek:
@RyanGlScott honestly, I don't have a strong preference on the naming here, so I'm happy to defer to your call. Agreed that exposing two names for the same function is redundant surface area. If `same*` reads cleaner to you, let's go with that and drop the `homogeneous*` pair, keeping just `sameOf`/`same`.

src/Control/Lens/Internal/Fold.hs
```diff
@@ -198,3 +199,37 @@ getRightmost :: Rightmost a -> Maybe a
...
+  | SameDiffer  -- ^ at least two targets differed
+
+instance Eq a => Semigroup (Same a) where
+  SameDiffer <> _          = SameDiffer        -- short-circuit, lazy in the right
+  SameEmpty  <> y          = y
+  x          <> SameEmpty  = x
+  SameOne a  <> SameOne b
+    | a == b               = SameOne a
+  _          <> _          = SameDiffer
+  {-# INLINE (<>) #-}
```
phadej:
```haskell
SameDiffer <> _          = SameDiffer        -- short-circuit, lazy in the right
SameEmpty  <> y          = y
x@(SameOne _)  <> SameEmpty  = x
SameOne _  <> SameDiffer = SameDiffer
x@(SameOne a)  <> SameOne b
    | a == b             = x -- no new allocation
    |                    = SameDiffer
```

Maybe GHC would figure that out itself, but not having overlapping patterns in the first place is better style, IMHO.

I'm also quite sure that `INLINE` is pessimisation

phadej:
`allSame` is a special case of 

```haskell
allPairwiseBy f = and $ zipWith f xs (tail xs)
allSame = allPairwiseBy (==)
```

with a small optimisation that it's enough to compare with first element (so we can have `x@(Same a) <> ... = x`) . But OTOH `allPairwiseBy` generalises e.g. `isSorted = allPairwiseBy (<=)` etc.

I feel that if `allSame` (under whatever name) is added to `lens` the more general combinator will have to follow.

jozanek:
@phadej good point — `same` is `allPairwiseBy (==)`. One subtlety: the `Same` monoid can compare against just the first element because `==` is transitive; a general `pairwiseBy` (e.g. `<=` for `isSorted`) needs adjacent comparisons, so a fatter monoid and the relation threaded through the fold — a separate, larger design. I'd rather not block this small combinator on it. Happy to open a follow-up issue for `pairwiseOf`/`isSorted` — sound good?

src/Control/Lens/Internal/Fold.hs
jozanek: Done. I rewrote the `Same` `Semigroup` with non-overlapping patterns (reusing the existing `SameOne` box on the equal case), and dropped the `INLINE` pragmas to match `Leftmost`/`Rightmost`. Thanks for the review!

PR #1109, opened 2026-06-06
Add haddock examples for IndexedSetter combinators (#565)

### LINKED ISSUE #565: example use request for IndexedSetter methods.
I've been tweaking the docs on the operator page at

https://github.com/ekmett/lens/wiki/Operators

As someone with recent exposure to lens, I think it would be good if every operator linked there had examples of use in the hackage.  This is the case for the left hand column of Control.Lens section but I hit my first confusion in "Control.Lens (Indexed)"

Could you add example usages in the haddock for iover at least, and perhaps other methods in this section there, in Control.Lens.Setter.

Specifically, I can't even figure out how to produce a value of the right type for iover (an IndexedSetter).

Thanks.

My hope is to make the operators page a friendlier jumping off point.

jozanek:
Fixes #565

Combinators covered: `iover`, `iset`, `isets`, `(%@~)`, `(.@~)`, `(%@=)`, `imodifying`, `(.@=)`.

The examples deliberately show *where the index comes from* (the original confusion): they use `imapped` (Int position for lists, key for `Map`) and one `isets`/`zipWith` example that builds an `IndexedSetter` by hand. `imapOf` is left untouched since it's deprecated.

All examples use `String` elements with `show`/`++` so they're type-clean (avoiding the `Int`-index vs `Integer`-literal defaulting mismatch) and produce deterministic `Show` output. Each `>>>` was extracted from the source and verified against its expected line in `cabal repl lens`. Doc-only change; builds `-Wall`-clean and `hlint` reports no hints.

### DIFF
--- a/src/Control/Lens/Setter.hs
+++ b/src/Control/Lens/Setter.hs
@@ -1192,6 +1192,15 @@ ilocally l f = Reader.local (iover l f)
 -- 'iover' :: 'IndexedLens' i s t a b      -> (i -> a -> b) -> s -> t
 -- 'iover' :: 'IndexedTraversal' i s t a b -> (i -> a -> b) -> s -> t
 -- @
+--
+-- An 'IndexedSetter' such as 'imapped' supplies the index: for a list it is the
+-- @Int@ position, for a 'Data.Map.Map' it is the key.
+--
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):
Thanks!

Issue #662, opened 2016-06-07
Document Ixed restriction


treeowl:
Along with the general law that `ix` should produce a valid `Traversal`, I suspect we also want to require that `ix` never use the `<*>` it gets from the `Applicative` constraint. Conceptually, it should only need `Pointed`. I believe this follows from the `Ixed/At` relationship when we have `At`.

treeowl:
This would rule out an `ix` that traverses all values associated with a single key in a multimap, requiring it to produce a single container with all of them. Of course, it's also possible that such a thing should be considered valid. In any case, some clarity would be nice.

scshunt:
I can't think of any reason to disallow it, although I'm not sure that, in the case of a multimap, it would be particularly useful.

treeowl:
One reason to disallow it is to allow `Ixed` users to rely on an `ix` traversal traversing at most one element. This allows fake `Applicative`s:

```haskell
newtype Fudge f a = Fudge (f a)
instance Pointed f => Applicative (Fudge f) where
  pure = point
  _ <*> _ = error "Not really Applicative"

ix' :: (Ixed m, Pointed f) => Index m -> (IxValue m -> f (IxValue m)) -> m -> f m
ix' = ... -- Use Fudge.
```

The downside, of course, is that someone might *want* their `Ixed` instance to traverse multiple values. If that's permitted, I think it might be nice to add an example of such an instance to the documentation.

RyanGlScott:
After some discussion in #1110, we have decided not to include this restriction in the Haddocks.

PR #1110, opened 2026-06-08
Document that `ix` focuses at most one element

jozanek:
Addresses #662.

Adds a law to the `Ixed` class Haddock: `ix k` should focus **at most one** element — the value at the given index, when present. Equivalently, `ix` uses only `pure`/`fmap` and never `<*>`, so it could in principle run under any `Pointed` functor rather than a full `Applicative`. For any type with an `At` instance this already follows from the existing law `ix k ≡ at k . traverse`.

Every existing `Ixed` instance in the library already satisfies this, so the change is documentation-only (no behaviour change). CHANGELOG updated.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,6 +1,7 @@
 next [????.??.??]
 -----------------
 * Add `ReifiedReview` to `Control.Lens.Reified`.
+* Document that `ix` focuses at most one element.
 
 5.3.6 [2026.01.10]
 ------------------
--- a/src/Control/Lens/At.hs
+++ b/src/Control/Lens/At.hs
@@ -213,6 +213,13 @@ type family IxValue (m :: Type) :: Type
 
 -- | Provides a simple 'Traversal' lets you 'traverse' the value at a given
 -- key in a 'Map' or element at an ordinal position in a list or 'Seq'.
+--
+-- An instance of 'Ixed' should provide an 'ix' that focuses /at most one/
+-- element: the value at the given index, when it is present. Equivalently,
+-- 'ix' should use only 'pure' and 'fmap', never '<*>', so that it could in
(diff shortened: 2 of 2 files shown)


phadej:
I'm not so sure about this. You can't really take advantage of the additional restriction, can you?

`lens` really should have affine traversals, so we can enforce that requirement in types. But the way `lens` works and `base` type-classes are, it's not really possible.

jozanek:
You're right that the proper way to express this is an affine traversal, and that lens can't give it a distinct type — the van Laarhoven encoding would need a "`Functor` with `pure` but no `<*>`" constraint, which `base` doesn't have.

On taking advantage of it: the only mechanism I'm aware of is the one from #662 — wrap into a `newtype` with a fake `Applicative` whose `<*>` is `error` and whose `pure` comes from `Pointed`, giving `ix' :: (Ixed m, Pointed f) => ...`. That relies on `ix` never touching `<*>`, but it's unsafe and not something lens should expose, so I take your point that there's no real, safe payoff inside lens itself.

So I didn't mean this as an enforceable law — just to answer #662 and signal to instance authors what's expected (e.g. ruling out an `ix` that traverses every value under a key in a multimap). If that's worth a note at all, I'm happy to reframe it as an explicit "convention, not enforced by the type" rather than a law; and if you don't think it's worth documenting given there's no enforcement, I'm equally happy to close.

phadej:
TBH, I kind of like the idea that `ix` would traverse every key&value in multimap. But OTOH I don't remember when was the last time I used multimap.

So I'd rather not do anything. "Focus on a single element" is IMO stating the obvious.

jozanek:
@phadej Sounds good. I agree it's not worth pinning down, and leaving the multimap `ix` open is a fair call. You can close #662, thanks for the look.

PR #1111, opened 2026-06-08
Document the relationship between `filtered` and `noneOf`/`anyOf` (#690)

### LINKED ISSUE #690: Document similarities between ‘filtered’ and ‘noneOf’
Should the documentation mention the similarities between `filtered` and `noneOf`

```haskell
noneOf id    :: (a -> Bool) -> (a -> Bool)
has.filtered :: (a -> Bool) -> (a -> Bool)

none                            :: (a -> Bool) -> (Foldable f => f a -> Bool)
noneOf folded                   :: (a -> Bool) -> (Foldable f => f a -> Bool) 
(\p -> has (folded.filtered p)) :: (a -> Bool) -> (Foldable f => f a -> Bool)
```

jozanek:
Closes #690.

Documents the similarity between `filtered` and `noneOf`/`anyOf` requested in the issue, while honoring @ekmett's caveat from the thread that `filtered` is stronger than a predicate test and is only narrowed to a fold "for real" by `has`.

### DIFF
--- a/src/Control/Lens/Fold.hs
+++ b/src/Control/Lens/Fold.hs
@@ -336,9 +336,9 @@ iterated f g = go where
 -- >>> [1..10]^..folded.filtered even
 -- [2,4,6,8,10]
 --
--- Composing 'filtered' with a 'Fold'-consuming combinator such as 'has' or 'anyOf'
--- turns it into a predicate test: @'has' (l '.' 'filtered' p)@ asks whether /any/
--- target of @l@ satisfies @p@, and 'noneOf' is its negation.
+-- Composing 'filtered' with the 'Fold'-consuming combinator 'has' turns it into a
+-- predicate test: @'has' (l '.' 'filtered' p)@ asks whether /any/ target of @l@
+-- satisfies @p@ (the same as @'anyOf' l p@), and 'noneOf' is its negation.
 --
 -- @
 -- 'has' (l '.' 'filtered' p) ≡ 'anyOf' l p
@@ -771,6 +771,8 @@ orOf l = getAny #. foldMapOf l Any
 -- >>> anyOf biplate (== "world") (((),2::Int),"hello",("world",11::Int))
 -- True
 --
+-- See 'filtered' for a related operation: @'anyOf' l p ≡ 'has' (l '.' 'filtered' p)@.
+--
 -- @
 -- 'Data.Foldable.any' ≡ 'anyOf' 'folded'
 -- @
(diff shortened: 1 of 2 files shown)


src/Control/Lens/Fold.hs
```diff
@@ -807,6 +825,8 @@ allOf l f = getAll #. foldMapOf l (All #. f)
 -- >>> noneOf (folded.folded) (<10) [[13,99,20],[3,71,42]]
 -- False
 --
+-- See 'filtered' for the relationship @'noneOf' l p ≡ 'not' '.' 'has' (l '.' 'filtered' p)@.
```
RyanGlScott: This sentence reads a bit strangely to me. I think something like "`See 'filtered' for a related operation @'noneOf' l p ≡ 'not' '.' 'has' (l '.' 'filtered' p)@`" would read more naturally.

Also, would it be worth including a similar reference in the Haddocks for `anyOf`?

src/Control/Lens/Fold.hs
```diff
@@ -336,6 +336,24 @@ iterated f g = go where
...
 -- [2,4,6,8,10]
 --
+-- Composing 'filtered' with a 'Fold'-consuming combinator such as 'has' or 'anyOf'
+-- turns it into a predicate test: @'has' (l '.' 'filtered' p)@ asks whether /any/
+-- target of @l@ satisfies @p@, and 'noneOf' is its negation.
+--
+-- @
+-- 'has' (l '.' 'filtered' p) ≡ 'anyOf' l p
+-- 'noneOf' l p           ≡ 'not' '.' 'has' (l '.' 'filtered' p)
+-- @
```
RyanGlScott: Strictly speaking, I don't think the code here shows an example of `anyOf` being _composed_ with `filtered`, just that the composition of `has` and `filtered` behaves the same as `anyOf`.

src/Control/Lens/Fold.hs
jozanek: well, not an easy task to be writing documentation as a non-native speaker...

src/Control/Lens/Fold.hs
jozanek: reworded

src/Control/Lens/Fold.hs
RyanGlScott: Please don't interpret my review comments as me criticizing your English skills! On the contrary: writing documentation is tricky even for native English speakers, so it is all the more impressive when someone can do it when English is a second language.

RyanGlScott (approved):
LGTM. Can you resolve the merge conflicts?

RyanGlScott (approved):

Issue #710, opened 2017-02-09
Incorporate some flavor of one-off lens & prism construction via TH


ekmett:
For example @chrisdone's

https://gist.github.com/chrisdone/c50e2ffc9e71a38d4fed9813fa90bbf4

We talked about this early on, but it seems to have fallen to the wayside.

jozanek:
**Proposal: `makeLens` / `makePrism` for one-off optic construction**

Picking this up — the [linked gist](https://gist.github.com/chrisdone/c50e2ffc9e71a38d4fed9813fa90bbf4) only covered lenses. Proposed addition to `Control.Lens.TH`:

```haskell
makeLens  :: Name -> ExpQ   -- a record-field selector name, e.g. '_field
makePrism :: Name -> ExpQ   -- a data-constructor name,       e.g. 'Ctor
```

Each splices a *single* optic as an **expression**, with its type inferred at the use site — no top-level declaration, no signature:

```haskell
over    $(makeLens '_field) f x
preview $(makePrism 'Ctor)  x
```

Semantics deliberately match the bulk generators (they reuse the same internals — `buildScaffold`/`makeFieldClauses` and `computeOpticType`/`makeConOpticExp`):

- `makeLens` yields the optic `makeLenses` would declare for that field: a `Lens`, a `Traversal` (field absent from some constructors), or an `Iso` (sole single-field constructor).
- `makePrism` yields the optic `makePrisms` would declare for that constructor: a `Prism`, an `Iso` (single-constructor type), or a `Review` (existential/GADT constructor).

Both reify their argument, so the focused type must already be in scope (same staging rule as `makeLenses`/`makePrisms`).

I have a branch ready (implementation + `tests/templates.hs` coverage + CHANGELOG). **The one open question is naming** — `makeLens`/`makePrism` (singular ↔ the existing plural forms) reads naturally to me, but new public names are your call. Alternative considered: `fieldLens`/`ctorPrism`. Happy to rename before opening the PR.

RyanGlScott:
I don't have a particularly strong opinion about the name, provided that the documentation clearly indicates that it expects a specific field or constructor name. I'm happy with the names `makeLens` and `makePrism`.

PR #1112, opened 2026-06-08
Add makeLens and makePrism for one-off optic construction (#710)

jozanek:
Implements #710 ([API proposed here](https://github.com/ekmett/lens/issues/710#issuecomment-4653656014)): build a single optic via Template Haskell as an *expression*, rather than declaring optics for a whole type.

Two new exports in `Control.Lens.TH`:

```haskell
makeLens  :: Name -> ExpQ   -- $(makeLens '_field)  — a record-field selector
makePrism :: Name -> ExpQ   -- $(makePrism 'Ctor)   — a data constructor
```

```haskell
over    $(makeLens '_field) f x
preview $(makePrism 'Ctor)  x
```

Each reuses the existing generation internals (`buildScaffold`/`makeFieldClauses` and `computeOpticType`/`makeConOpticExp`), so the spliced optic matches what `makeLenses`/`makePrisms` would declare for that field or constructor — a `Lens`/`Traversal`/`Iso`, or a `Prism`/`Iso`/`Review`.

### DIFF
--- a/src/Control/Lens/Internal/PrismTH.hs
+++ b/src/Control/Lens/Internal/PrismTH.hs
@@ -188,6 +188,8 @@ makePrism conName =
 makeConsPrisms :: Type -> [NCon] -> Maybe Name -> DecsQ
 
 -- special case: single constructor, not classy -> make iso
+-- ('makePrism' has the corresponding single-constructor Iso case; keep the two
+-- in sync.)
 makeConsPrisms t [con@(NCon _ [] [] _)] Nothing = makeConIso t con
 
 -- top-level definitions
--- a/src/Control/Lens/Internal/PrismTH.hs
+++ b/src/Control/Lens/Internal/PrismTH.hs
@@ -20,6 +20,7 @@ module Control.Lens.Internal.PrismTH
   ( makePrisms
   , makeClassyPrisms
   , makeDecPrisms
+  , makePrism
   ) where
 
 import Control.Applicative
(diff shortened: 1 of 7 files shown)


tests/templates.hs
```diff
@@ -476,5 +476,41 @@ declareFields [d|
...
+
+oneOffL :: Prism (Sum710 a b) (Sum710 c b) a c
+oneOffL = $(makePrism 'L710)
+
+oneOffR :: Prism (Sum710 a b) (Sum710 a c) b c
+oneOffR = $(makePrism 'R710)
+
+-- ...and an 'Iso' for a single-constructor type.
+oneOffWrap :: Iso (T997A a) (T997A b) a b
+oneOffWrap = $(makePrism 'MkT997A)
```
RyanGlScott: For more complete test coverage, it would be worth including an example that produces a `Review` for an existentially quantified data constructor.

src/Control/Lens/Internal/PrismTH.hs
RyanGlScott, line 193: In order to keep this code in sync with the corresponding code in `makePrism`, please expand the comment here to include a forward reference to `makePrism`'s code. (There is already a forward reference in `makePrism`'s code, so this is the only place that needs one.)

RyanGlScott (commented):
Thanks for the patch, @jozanek! Two suggestions inline.

tests/templates.hs
jozanek: Added `Review` case for `ReviewTest`, and broadened coverage with simple/multi-field prisms plus runtime checks in `hunit`.

src/Control/Lens/Internal/PrismTH.hs
jozanek: Added the sync comment.

RyanGlScott (approved):
Thanks!

PR #1113, opened 2026-06-09
Document that Seq is an instance of Cons and Snoc

### LINKED ISSUE #780: Link to Control.Lens.Cons from Data.Sequence.Lens
I suspect most people who look into `Data.Sequence.Lens` will also want to know that `Seq` is an instance of `Cons` and of `Snoc`, and that they can therefore use various facilities in `Control.Lens.Cons`. I know re-exports are widely frowned upon, but I think it would pay to at least include a note in the documentation. Perhaps something like

> `Seq` is an instance of `Cons` and of `Snoc`. This means you can use the prisms `_Cons` and `_Snoc` with sequences, along with the other functions in `Control.Lens.Cons`.

Unfortunately, my Haddock-fu is not very strong, so I don't know the right place/way to add that documentation. But I feel fairly strongly that it should go right at the top of the module.

jozanek:
Closes #780.

`Data.Sequence.Lens` had an empty module description. This adds a note to the Haddock header pointing out that `Seq` is an instance of `Cons` and `Snoc`, so the `_Cons`/`_Snoc` prisms and the rest of `Control.Lens.Cons` work on sequences.

The wording and link style mirror the existing Cons/Snoc documentation in `Data.List.Lens`, and two doctest examples are included.

### DIFF
--- a/src/Data/Sequence/Lens.hs
+++ b/src/Data/Sequence/Lens.hs
@@ -9,6 +9,19 @@
 -- Stability   :  experimental
 -- Portability :  non-portable
 --
+-- This module provides lenses, traversals, and isomorphisms for
+-- working with 'Seq'.
+--
+-- 'Seq' is an instance of 'Control.Lens.Cons.Cons' and
+-- 'Control.Lens.Cons.Snoc', so the prisms '_Cons' and '_Snoc', along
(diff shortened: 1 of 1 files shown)


RyanGlScott (approved):
Thanks!

PR #1114, opened 2026-06-11
Fail early when TH namers generate keyword names; add avoidKeywords namers

### LINKED ISSUE #762: Should `classUnderscoreNoPrefixNamer` and the like replace builtin names like `type`, `data`, etc. using some kind of common mangling scheme?
They used to work when qualified with a module due to some old quirks in GHC, but those quirks have since been fixed. We could chase them with an _ or a ' or something. We _know_ the current state will fail. The alternative merely might.

### LINKED ISSUE #774: makeFields fails if field is a keyword
Consider the following snippet
```Haskell

data Connection = Connection
  { _connectionType :: Int
  }

makeFields ''Connection
```
it fails with
```
    Illegal variable name: ‘type’
    When splicing a TH declaration:
      class HasType s a | s -> a
    where type :: Control.Lens.Type.Lens' s a
```

I'm not sure if it's a bug or feature, but it'd be nice to have it rename by default or provide an informative error message that instructs you to specify a special `lensField` namer.

jozanek:
Fixes #762, fixes #774.

When a field namer derives a name that is a Haskell keyword (e.g. `makeFields` on a record with a `_connectionType` field derives a class method named `type`), GHC rejects the spliced declarations with an `Illegal variable name: ‘type’` error that names neither the field nor the data type responsible. The same happens for `makeClassy` on a type whose name lowercases to a keyword (`makeClassy ''Where` → method `where`).

Following the discussion in #762 (fail by default, ship the mangling rule as an opt-in):

* `makeFieldOpticsForDatatype` now validates every generated `DefName`, and the classy method name, before splicing. It fails in `Q` with an error naming the data type, offending field, and generated reserved name, and suggests remedies. On `template-haskell >= 2.12`, this also accounts for enabled extensions that reserve additional generated function or class method names.
* New `avoidKeywordsNamer :: FieldNamer -> FieldNamer` and `avoidKeywordsClassyNamer :: ClassyNamer -> ClassyNamer` append `_` to keyword names.
* New `avoidNamesNamer :: Set String -> FieldNamer -> FieldNamer` and `avoidNamesClassyNamer :: Set String -> ClassyNamer -> ClassyNamer` generalize this to arbitrary reserved-name sets. The PR also exports `haskellKeywords`, per-extension keyword sets, and `ghcExtensionKeywords` for broad opt-in use.

Defaults are unchanged; users only get mangling when they opt into the new namer wrappers.

Tests: positive tests for the keyword wrappers and `ghcExtensionKeywords`, including `forall`, `by`, `using`, `static`, and `role`; recover-based negative tests guard that invalid generated names fail early inside `Q`.

One note for review: `recover` catching `fail` should hold across the CI matrix, but if an old GHC misbehaves the two negative splices can be CPP-gated (`__GLASGOW_HASKELL__ >= 806`) or dropped without losing the positive coverage.

A second note: `forall` is included unconditionally. On GHC < 9.10 with no `forall`-reserving extension a generated method literally named `forall` would now be rejected though it would have compiled there — the only dent in "nothing newly rejected." I kept one shared list rather than CPP-gating on `__GLASGOW_HASKELL__ >= 910`; happy to gate it if you'd prefer the invariant exact on old GHCs.


src/Control/Lens/Internal/TH.hs
```diff
@@ -164,6 +164,31 @@ isDataFamily D.NewtypeInstance = True
...
+--
+-- This mirrors the check GHC applies to spliced declarations
+-- (@okVarOcc@\/@reservedIds@ in "GHC.Utils.Lexeme" from @ghc-boot@, on which
+-- we do not wish to depend). Like GHC's @reservedIds@, it omits @forall@.
+haskellKeywords :: Set String
+haskellKeywords = Set.fromList
+  [ "case", "class", "data", "default", "deriving", "do", "else"
+  , "foreign", "if", "import", "in", "infix", "infixl", "infixr"
+  , "instance", "let", "module", "newtype", "of", "then", "type"
+  , "where", "_"
```
phadej, line 182: When some GHC extensions are enabled, GHC steals few more symbols, see e.g. list in https://github.com/well-typed/hs-bindgen/blob/1a1d5e029dc98a02e7b177d1cda5982676252fc9/hs-bindgen/src-internal/HsBindgen/Config/MangleCandidate/ReservedNames.hs#L98

I'd highlight `forall`. I don't remember which extrensions steal that keyword in term namespace, but I recall libraries renaming `forall` to `forAll`.

phadej: Actually, it's just a keyword in recent GHCs:

```
% ghci-9.14.1 -XHaskell2010
GHCi, version 9.14.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/phadej/.ghci
Prelude> forall = 'x'
<interactive>:1:1: error: [GHC-58481]
    parse error on input ‘forall’
```

jozanek: Good catch, thanks. Please also check 2 notes I left on PR description.

RyanGlScott: @phadej raises a valid point: this list of keywords is likely not exhaustive, as what GHC considers to be a keyword depends on what language extensions you have enabled. As such, we might want to give users the ability to extend this list of keywords with ones that are extension-dependent. We could take inspiration from https://github.com/well-typed/hs-bindgen/blob/1a1d5e029dc98a02e7b177d1cda5982676252fc9/hs-bindgen/src-internal/HsBindgen/Config/MangleCandidate/ReservedNames.hs#L98 and provide pre-assembled lists of keywords that users can pick from depending on their needs.

jozanek: Generalized `avoidKeywordsNamer`/`avoidKeywordsClassyNamer` and shipped the lists:

```haskell
avoidNamesNamer       :: Set String -> FieldNamer  -> FieldNamer
avoidNamesClassyNamer :: Set String -> ClassyNamer -> ClassyNamer
avoidKeywordsNamer       = avoidNamesNamer       haskellKeywords
avoidKeywordsClassyNamer = avoidNamesClassyNamer haskellKeywords
haskellKeywords, recursiveDoKeywords, arrowsKeywords, patternSynonymsKeywords :: Set String
```

So you can pick and combine: `avoidNamesNamer (haskellKeywords <> recursiveDoKeywords)`. Verified the per-extension contents: `RecursiveDo` => `{mdo, rec}`, `Arrows` => `{proc, rec}`, `PatternSynonyms` => `{pattern}`.

I left the fail-early validation on `haskellKeywords` alone, since it runs at every splice and can't reject extension-keywords without false-positiving users who lack the extension. Could make it exhaustive via `isExtEnabled` if you want.

RyanGlScott: Thanks, I like having the ability to customize this further. I think my only remaining quibbles are:

* I think this is missing some keywords (e.g., `by`, `role`, and `static`).
* It might be nice to offer something akin to `hs-bindgen`'s `ghcExtensionKeywords` that groups all of the extension-specific keywords into a single definition. In order to achieve the same thing at present, you'd have to write `recursiveDoKeywords <> arrowsKeywords <> patternSynonymsKeywords <> ...`, which is a bit verbose.

jozanek: @RyanGlScott I pushed an update addressing your points.

Changes:
* Added the missing extension keyword sets and `ghcExtensionKeywords` as the aggregate opt-in set.
* Made `forall` part of `haskellKeywords` only on GHC >= 9.10; older GHCs can use `explicitForAllKeywords`.
* Extended early validation to account for enabled extensions via `isExtEnabled` when `template-haskell >= 2.12`.

RyanGlScott: Thanks, I'm happy with the range of options available now.

RyanGlScott (approved):

PR #1116, opened 2026-06-13
Re-add Control.Monad.Primitive.Lens

### LINKED ISSUE #819: Control.Monad.Primitive.Lens
`Control.Monad.Primitive` offers quite a few isomorphisms involving `PrimBase` types. I think the only other optics available there are `Getter`s, which don't seem to deserve inclusion.

jozanek:
Closes #819.

`Control.Monad.Primitive.Lens` was removed in lens 4.9 only to drop the `primitive` dependency — which is now pulled in transitively via `vector` anyway. This re-adds it with three lawful `Iso`s over the conversions in `Control.Monad.Primitive`:

- `prim` — between a `PrimBase` monad and its underlying `State# -> (# State#, a #)` representation (`iso internal primitive`). This is the original optic, now generalized to be type-changing.
- `st` — between a `PrimBase` monad and `ST` with the same state token (`iso primToST stToPrim`).
- `io` — between a `PrimBase` monad whose state token is `RealWorld` and `IO` (`iso primToIO ioToPrim`).

### Notes / decisions (happy to change any of these)

- `unsafe*` excluded. Those conversions coerce between different `PrimState` tokens and aren't lawful isos, so they're deliberately left out (noted in the module haddock).
- No name for the generic `PrimBase`↔`PrimBase` conversion: it's exactly `prim . from prim`, so it's documented on `prim` rather than exported separately. Easy to add a named optic if you'd prefer.
- Naming. `st`/`io` follow the "name the iso after what `view` yields" convention (`vector`, `generic`, …). `primST`/`primIO` is the obvious alternative if the short names feel too grabby.
- `prim` is type-changing (`Iso (m a) (m b) … …`) rather than the historical `Iso'`. Strictly more general; no back-compat concern after its long absence.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -4,8 +4,7 @@ next [????.??.??]
   `lens` 4.9. It offers `Iso`s for the conversions in
   `Control.Monad.Primitive`: `prim` (between a `PrimBase` monad and its
   underlying `State#` representation, now type-changing), and the new `st`
-  and `io` (between a `PrimBase` monad and `ST`/`IO`, respectively). This
-  reintroduces the dependency on the `primitive` package.
+  and `io` (between a `PrimBase` monad and `ST`/`IO`, respectively).
 * Add `ReifiedReview` to `Control.Lens.Reified`.
 * Fix `Data.Data.Lens.upon` (and its variants) looping forever when nested,
   e.g. `(upon.view.upon) tail`.
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,11 @@
 next [????.??.??]
 -----------------
+* Re-add the `Control.Monad.Primitive.Lens` module, which was removed in
+  `lens` 4.9. It offers `Iso`s for the conversions in
+  `Control.Monad.Primitive`: `prim` (between a `PrimBase` monad and its
+  underlying `State#` representation, now type-changing), and the new `st`
+  and `io` (between a `PrimBase` monad and `ST`/`IO`, respectively). This
+  reintroduces the dependency on the `primitive` package.
 * Add `ReifiedReview` to `Control.Lens.Reified`.
 * Add `makeLens` and `makePrism` to `Control.Lens.TH`. These build a single
   optic, as an expression, for one record field or one data constructor (e.g.
--- a/src/Control/Monad/Primitive/Lens.hs
+++ b/src/Control/Monad/Primitive/Lens.hs
@@ -1,0 +1,100 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE TypeFamilies #-}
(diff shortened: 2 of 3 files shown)


src/Control/Monad/Primitive/Lens.hs
```diff
@@ -0,0 +1,100 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UnboxedTuples #-}
+#ifdef TRUSTWORTHY
+{-# LANGUAGE Trustworthy #-}
+#endif
+
+#include "lens-common.h"
```
RyanGlScott: Is this `#include` required? I don't think this module uses anything from `lens.common.h`.

src/Control/Monad/Primitive/Lens.hs
```diff
@@ -0,0 +1,100 @@
...
+-- converting directly between any two 'PrimBase' monads that share a
+-- 'PrimState' token. 'st' and 'io' compose the same way.
+--
+-- @
+-- 'prim' '.' 'from' 'prim'
+--   :: ('PrimBase' m, 'PrimBase' n, 'PrimState' m ~ 'PrimState' n)
+--   => 'Iso' (m a) (m b) (n a) (n b)
+-- @
+--
+-- >>> (pure 5 :: IO Int) ^. prim . from prim :: IO Int
```
RyanGlScott: Perhaps this example would be more compelling if it showed how one can change types (e.g., changing `ST` to `IO`).

CHANGELOG.markdown
```diff
@@ -1,9 +1,12 @@
 next [????.??.??]
 -----------------
+* Re-add the `Control.Monad.Primitive.Lens` module, which was removed in
+  `lens` 4.9. It offers `Iso`s for the conversions in
+  `Control.Monad.Primitive`: `prim` (between a `PrimBase` monad and its
+  underlying `State#` representation, now type-changing), and the new `st`
+  and `io` (between a `PrimBase` monad and `ST`/`IO`, respectively). This
+  reintroduces the dependency on the `primitive` package.
```
RyanGlScott: As you've noted, `lens` was already depending on `primitive` due it being a transitive dependency via `vector`. Perhaps it's not worth including the last sentence.

src/Control/Monad/Primitive/Lens.hs
jozanek: nope, left from my experiments

src/Control/Monad/Primitive/Lens.hs
jozanek: updated

CHANGELOG.markdown
jozanek: dropped

treeowl:
Cool. Thanks.

Issue #1078, opened 2024-10-07
Type signature of outside on cheat sheet is wrong


turion:
The type signature on the cheat sheet says:

```haskell
outside :: Prism s t a b -> Lens (s -> r) (t -> r) (a -> r) (b -> r)
```

But in fact `s` and `t`, and `a` and `b` should be swapped:

```haskell
outside :: Prism s t a b -> Lens (t -> r) (s -> r) (b -> r) (a -> r)
```

turion:
See https://hackage.haskell.org/package/lens-5.3.2/docs/Control-Lens-Prism.html#v:outside and specialise `p = (->)`.

RyanGlScott:
Fixed in https://github.com/ekmett/lens/pull/1117.

PR #1117, opened 2026-06-13
Recreate the lens hierarchy diagram from an editable Graphviz source

### LINKED ISSUE #828: class diagram has wrong type signatures for Review
The `Review` box of the class diagram says

    Review s a
    ---
    unto :: (b -> t) -> Review s t a b
    ---
    re :: Review s t a b -> Getter b t
    review :: MonadReader b m => Review s t a b -> m t
    reuse :: MonadState b m => Review s t a b -> m t

But there's no such thing as a `Review s t a b`, only a `Review t b`. Also, `Review` typically uses the type parameters `t` and `b`, not `s` and `a`. So the diagram should instead say:

    Review t b
    ---
    unto :: (b -> t) -> Review t b
    ---
    re :: Review t b -> Getter b t
    review :: MonadReader b m => Review t b -> m t
    reuse :: MonadState b m => Review t b -> m t

### LINKED ISSUE #1012: creately.com diagram template broken, blocks #828
https://creately.com/diagram/h5nyo9ne1/QZ9UBOtw4AJWtmAKYK3wT8Mm1HM%3D

This seems broken and to just result in an example diagram unrelated to the actually UML, which blocks fixing #828

jozanek:
## What & why

The "field guide" hierarchy diagram (`images/Hierarchy.png`) had two long-standing problems:

- **#828** — the `Review` box showed a four-parameter `Review s t a b`, but the type has had only two parameters (`Review t b`) for years.
- **#1012** — the diagram was authored on creately.com, whose source link is now dead, so it couldn't be edited to fix #828. Only the flat PNG lived in the repo.

This PR resolves both by bringing an **editable source** into the repo — a Graphviz file (`images/Hierarchy.dot`) that renders the PNG, plus a `scripts/hierarchy` wrapper to regenerate it. With the source in hand, the stale signatures are corrected. Each optic is one box; each combinator is one `<TR>` row in its HTML-label table, so future corrections are one-line edits.

Fixes #828.
Fixes #1012.
Fixex #1078.

## The diagram

Lens Hierarchy

Same lattice as before — the most capable optic (`Equality`) at the bottom, hollow "use-as" arrows pointing up to the weaker optics each can stand in for. It's a Graphviz reproduction, so the layout isn't pixel-identical to the old hand-drawn art, but the structure and (now-correct) signatures match the library.

## Signature corrections

The old art predated several API changes, so a faithful transcription surfaced more than just `Review`. Every combinator row was audited against current source (each module's documented "simpler type", or the real type). Corrections:

| Combinator | Was | Now |
| --- | --- | --- |
| `Review` (box) | `Review s t a b` (4 params) | `Review t b` |
| `au` | `… ((s -> a) -> e -> b) -> e -> t` | `… ((b -> t) -> e -> s) -> e -> a` (direction changed in 4.18) |
| `outside` | `Lens (s->r) (t->r) (a->r) (b->r)` | `Lens (t->r) (s->r) (b->r) (a->r)` |
| `elementOf` | `Traversal s t a b -> Int -> Traversal s t a b` | `Traversal' s a -> Int -> IndexedTraversal' Int s a` |
| `folded` | `Fold (f a) a` | `IndexedFold Int (f a) a` |

## Note on the Hackage description image (maintainer call)

The README embeds the diagram with a repo-relative path, so it tracks whatever ref you view. The Cabal `description:` can't use a relative path (Hackage won't resolve it — that's what #1023 fixed), so it needs an absolute URL. I changed it from the `master` raw URL to a **release-tagged** one (`…/v5.3.7/…`) so each release's Hackage page shows that release's diagram rather than a moving `master` asset.

Trade-off worth your call: the tag in that URL must then be **bumped alongside `version:` at each release** (there's a reminder in `scripts/README.md`). If you'd rather keep the zero-maintenance `master` URL, easy to switch back.


RyanGlScott:
Thanks for doing this! Having a way to reproducibly build diagrams is very helpful.

I am not terribly bothered by using the `v5.3.7` diagram in the `.cabal` file for now, but I do think it would be nice to switch to the new diagram at some point (perhaps after the next `lens` release).

Some various suggestions I have for the new diagram itself:

* The `s = t`/`a = b` captions on some of the arrows overlap with the arrows themselves (especially on the arrow from `Traversable s t a b` to `Fold s a`). Can you adjust the positions of the caption to tidy this up?
* Unless I'm missing something, the boxes now appear to be missing the definitions of the type synonyms (e.g., notice that `type Lens s t a b = forall f. Functor f => (a -> f b) -> s -> f t` is now missing from the `Lens s t a b` box). Is this intentional?
* Similarly, some boxes now lack certain type signatures (e.g., the `Fold` box now lacks the signature for `(^.)`).

RyanGlScott (approved):

no-identd:
Holy shit you actually fixed it.🎉 🥳

PR #1118, opened 2026-06-14
Add lens-partial-iso companion package: inverse prisms and partial isos (#904)

### LINKED ISSUE #904: Turned-around Prisms and partial Isos
When experimenting [with parsing and building using `Prism`s](https://yairchu.github.io/posts/codecs-as-prisms-asts.html) I found a need for "turned-around" `Prism`s (kind of like `re` but not becoming uni-directional) and "partial" `Iso`s (conversion to both sides may fail).

I came up with implementations that seem to work nicely, `Prism`s composed with `InvPrism`s become `PartialIso`s, etc:

```Haskell
class Functor f => Fallback f where
    fallback :: f (Maybe a) -> f a

instance Fallback (Const a) where fallback (Const x) = Const x
instance Fallback Maybe where fallback = join

-- A partial variant of (#) for partial reviews
infixr 8 #?
(#?) :: Optic Tagged Maybe s t a b -> b -> Maybe t
f #? b = Just b & Tagged & f & unTagged

-- A Prism "turned around", i.e a getter but a partial review
type InvPrism s t a b =
    forall p f. (Profunctor p, Fallback f) => Optic p f s t a b

-- A partial Iso, i.e a partial getter and partial review
type PartialIso s t a b =
    forall p f. (Choice p, Applicative f, Fallback f) => Optic p f s t a b

-- Turn APrism around
invPrism :: APrism b a t s -> InvPrism s t a b
invPrism p =
    dimap
    (review (reviewing (clonePrism p)))
    (fallback . fmap (^? getting (clonePrism p)))

-- Create a PartialIso from two partial conversions
partialIso :: (s -> Maybe a) -> (b -> Maybe t) -> PartialIso s t a b
partialIso sma bmt =
    dimap
    (maybe (Left ()) Right . sma)
    (fallback . either (const (pure Nothing)) (fmap bmt)) .
    right'

-- Could this be done by `failing`?
failing' ::
    PartialIso s t a b ->
    Prism s t a b ->
    Prism s t a b
failing' i p =
    prism
    (\b -> fromMaybe (reviewing p # b) (i #? b))
    (\s -> maybe (matching p s) Right (s ^? getting i))

-- Coerce APrism to an Optic'
reviewing ::
    (Choice p, Bifunctor p, Functor f, Settable f) =>
    APrism s t a b -> Optic' p f t b
reviewing p =
    bimap f (fmap f)
    where
        f = runIdentity . unTagged . clonePrism p . Tagged . Identity
```

* `failing'` is a parsing/building combinator - it first tries the given `PartialIso` in both parsing and building, and falls back to the given `Prism`. I called it `failing'` because to me it seems similar to `failing`, though trying to see if the latter can be generalised also to fit this use case seems over my head.
* Note that `InvPrism` doesn't carry all of the same power as `Prism` does, i.e it can't support being an "inverted setter" if such operations existed as it's matching may simply fail and not fall back to type-change the structure.

jozanek:
## Summary

Closes #904 by adding the inverse prisms and partial isos requested there, as a new `lens-witherable` companion package — so `lens` itself takes on no new dependency.

- `InvPrism s t a b` — a `Prism` turned around: the get direction is total, but the build (`review`) may fail.
- `PartialIso s t a b` — an `Iso` where both directions may fail.
- Constructors `invPrism`, `partialIso`; the parse/build fallback combinator `failing'`; and the partial-review operator `(#?)`.

## Background & design discussion

The design history is in #904. As @ekmett noted there:

> the profunctor version of things did lead naturally to 'inverse prisms' and 'inverse lenses', but … the mixed profunctor/functor encoding we wound up using in `lens` got in the way of flipping. … in the profunctor encoding … `re` and `from` and `invPrism` … all collapsed into precisely the same `from` combinator.

lens encodes every optic as `p a (f b) -> p s (f t)`, splitting the two directions across a profunctor `p` (matching) and a functor `f` (building). A `Prism` is `(Choice p, Applicative f)` — matching may fail, building can't. These optics need the opposite asymmetry, which `Applicative` can't express, so they replace it with `Filterable` (a `Functor` you can also drop out of, via `mapMaybe`/`catMaybes`). In the mixed encoding they can't "fall out of `from`" for free, so they're new named types rather than reuses of `re`/`from`.

## Packaging
`Filterable` lives in `witherable`, which `lens` deliberately does not depend on. Following @ekmett's call on #904, these optics ship as a self-contained companion package, `lens-witherable`, so `lens` core is untouched and the change is purely additive: a new in-repo sub-package alongside `lens-properties`, depending only on `lens` + `witherable`.
This was originally named `lens-witherable`, but that package name is already taken on Hackage. The companion package is now named `lens-partial-iso`, which describes the exported API directly while still keeping the `witherable` dependency out of `lens` core.

## Open questions for reviewers

1. `failing'` name: the `'` suffix usually means a stricter variant in lens. Open to a less overloaded name.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,4 @@
 packages: .
           ./examples
           ./lens-properties
+          ./lens-witherable
--- a/lens-witherable/.hlint.yaml
+++ b/lens-witherable/.hlint.yaml
@@ -1,0 +1,4 @@
+- arguments: [--cpp-ansi]
+- ignore: { name: Use camelCase }
+- fixity: "infixr 9 ..."
+- fixity: "infixl 1 &~"
--- a/lens-witherable/CHANGELOG.markdown
+++ b/lens-witherable/CHANGELOG.markdown
@@ -1,0 +1,5 @@
+0.1 [????.??.??]
+----------------
+* Initial release. Provides `Control.Lens.PartialIso` with the `InvPrism` and
+  `PartialIso` optic types, the `invPrism` and `partialIso` constructors, the
+  `failing'` fallback combinator, and the `(#?)` partial-review operator.
--- a/lens-witherable/LICENSE
+++ b/lens-witherable/LICENSE
@@ -1,0 +1,26 @@
+Copyright 2012-2016 Edward Kmett
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright
+   notice, this list of conditions and the following disclaimer in the
+   documentation and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE AUTHORS ``AS IS'' AND ANY EXPRESS OR
+IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED.  IN NO EVENT SHALL THE AUTHORS OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+POSSIBILITY OF SUCH DAMAGE.
--- a/lens-witherable/Setup.hs
+++ b/lens-witherable/Setup.hs
@@ -1,0 +1,2 @@
+import Distribution.Simple
+main = defaultMain
--- a/lens-witherable/lens-witherable.cabal
+++ b/lens-witherable/lens-witherable.cabal
@@ -1,0 +1,70 @@
+name:          lens-witherable
+category:      Data, Lenses
+version:       0.1
+license:       BSD2
+cabal-version: >= 1.10
+license-file:  LICENSE
+author:        Edward A. Kmett
+maintainer:    Edward A. Kmett <ekmett@gmail.com>
+stability:     provisional
+homepage:      http://github.com/ekmett/lens/
+bug-reports:   http://github.com/ekmett/lens/issues
+copyright:     Copyright (C) 2012-2016 Edward A. Kmett
+synopsis:      Inverse prisms and partial isomorphisms for lens
+description:
+  Optics for reversible parsing and building (codecs) that need the
+  @Filterable@ class from @witherable@: \"turned-around\" prisms whose
+  @review@ may fail, and partial isomorphisms where both directions may fail.
+  .
+  These live in a companion package so that @lens@ itself does not have to
+  depend on @witherable@.
+build-type:    Simple
+tested-with:   GHC == 8.0.2
+             , GHC == 8.2.2
+             , GHC == 8.4.4
+             , GHC == 8.6.5
+             , GHC == 8.8.4
+             , GHC == 8.10.7
+             , GHC == 9.0.2
+             , GHC == 9.2.8
+             , GHC == 9.4.8
+             , GHC == 9.6.7
+             , GHC == 9.8.4
+             , GHC == 9.10.3
+             , GHC == 9.12.2
+             , GHC == 9.14.1
+
+extra-source-files:
+  .hlint.yaml
+  CHANGELOG.markdown
+
+source-repository head
+  type: git
+  location: https://github.com/ekmett/lens.git
+
+library
+  build-depends:
+    base       >= 4.9   && < 5,
+    lens       >= 5     && < 6,
+    tagged     >= 0.8.6 && < 1,
+    witherable >= 0.3.5 && < 0.6
+
+  exposed-modules:
+    Control.Lens.PartialIso
+
+  hs-source-dirs: src
+  ghc-options: -Wall
+  default-language: Haskell2010
+
+test-suite hunit
+  type: exitcode-stdio-1.0
+  main-is: hunit.hs
+  hs-source-dirs: tests
+  ghc-options: -Wall -threaded
+  default-language: Haskell2010
+  build-depends:
+    base,
+    lens,
+    lens-witherable,
+    tasty       >= 1.4  && < 1.6,
+    tasty-hunit >= 0.10 && < 0.11
--- a/lens-witherable/src/Control/Lens/PartialIso.hs
+++ b/lens-witherable/src/Control/Lens/PartialIso.hs
@@ -1,0 +1,219 @@
+{-# LANGUAGE RankNTypes #-}
+
+-------------------------------------------------------------------------------
+-- |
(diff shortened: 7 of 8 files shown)


ekmett:
Good stuff. 

Somebody asked me about my thoughts on `witherable` optics at ZuriHac, so I've had a few days to think about this, despite the patch coming in just now.

I'm inclined to say this belongs in a self-contained companion package, avoiding incurring extra dependencies for the main `lens` package, much like we did with the lens zipper code.

A lot of what is in the `lens` package making it as monolithic as it is and not split out is mostly just because I can't implement the current API with a factored out implementation, but that isn't the case here.

jozanek:
@ekmett Thanks! That's what I have pushed. The PR now adds a self-contained `lens-witherable` companion package and reverts every change to `lens` itself, so core takes on no `witherable` dependency. It's purely additive, set up as an in-repo sub-package alongside `lens-properties`.

Two calls I'd leave to you: in-repo sub-package vs. its own repo, and the name `lens-witherable`.

RyanGlScott:
I don't have a strong opinion on where this lives, but it is worth noting that the name `lens-witherable` is already taken: https://hackage.haskell.org/package/lens-witherable

lens-partial-iso/src/Control/Lens/PartialIso.hs
```diff
@@ -0,0 +1,219 @@
+{-# LANGUAGE RankNTypes #-}
+
+-------------------------------------------------------------------------------
+-- |
+-- Module      :  Control.Lens.PartialIso
+-- Copyright   :  (C) 2012-16 Edward Kmett
```
RyanGlScott: Given that you are the author, you may as well list your name here.

Also, please update the current copyright year to 2026.

lens-partial-iso/LICENSE
```diff
@@ -0,0 +1,26 @@
+Copyright 2012-2016 Edward Kmett
```
RyanGlScott: Similar comments here about copyright.

lens-partial-iso/src/Control/Lens/PartialIso.hs
```diff
@@ -0,0 +1,219 @@
...
+-- License     :  BSD-style (see the file LICENSE)
+-- Maintainer  :  Edward Kmett <ekmett@gmail.com>
+-- Stability   :  provisional
+-- Portability :  RankNTypes
+--
+-- A 'Prism' matches one way and builds the other: matching may fail, but
+-- building (a 'review') always succeeds. This module provides two optics with
```
RyanGlScott: For symmetry, it may be worth mentioning the function that performs matching, i.e.,


```suggestion
-- A 'Prism' matches one way and builds the other: matching (a 'preview') may fail, but
-- building (a 'review') always succeeds. This module provides two optics with
```

lens-partial-iso/src/Control/Lens/PartialIso.hs
```diff
@@ -0,0 +1,219 @@
...
+-- the opposite flavor of partiality, useful when describing reversible
+-- parsing\/building (codecs):
+--
+-- * An 'InvPrism' is a 'Prism' \"turned around\": the get direction always
+--   succeeds, but the build direction (the 'review') may fail.
```
RyanGlScott: Perhaps this should say "match direction" instead of "get direction" to more closely resemble the text above?


```suggestion
-- * An 'InvPrism' is a 'Prism' \"turned around\": the match direction (the 'preview') always
--   succeeds, but the build direction (the 'review') may fail.
```

Alternatively, we could say "get" instead of "match" both times---whatever you think is a better description.

Similar comments apply to the Haddocks below.

lens-partial-iso/src/Control/Lens/PartialIso.hs
```diff
@@ -0,0 +1,219 @@
...
+  , partialIso
+  -- * Combinators
+  , failing'
+  , (#?)
+  -- * Filterable
+  --
+  -- | We re-export the 'Filterable' methods used to build these optics, but not
+  -- the rest of the class — in particular @filter@, which would clash with
+  -- "Prelude".'Prelude.filter'. Import "Witherable" directly for the others.
+  , Filterable(mapMaybe, catMaybes)
```
RyanGlScott: Truth be told, I'm not sure how I feel about re-exporting `mapMaybe` or `catMaybes`. While these aren't exported by the `Prelude`, they _are_ exported by `Data.Maybe`, which is a fairly common import. I'd be inclined just to leave it to users to import `Witherable` themselves.

lens-partial-iso/src/Control/Lens/PartialIso.hs
```diff
@@ -0,0 +1,219 @@
...
+-- /Laws:/ a simple @'PartialIso'' s a@ presents two partial functions, a
+-- forward @sma :: s -> 'Maybe' a@ (the get) and a backward @ams :: a -> 'Maybe'
+-- s@ (the build). They must be mutually inverse where defined:
+--
+-- @
+-- sma s ≡ 'Just' a   ⟹   ams a ≡ 'Just' s
+-- ams a ≡ 'Just' s   ⟹   sma s ≡ 'Just' a
+-- @
+--
+-- (For the type-changing @'PartialIso' s t a b@ the laws generalize exactly as
```
RyanGlScott:
```suggestion
-- (For a type-changing @'PartialIso' s t a b@, the laws generalize exactly as
```

lens-partial-iso/lens-partial-iso.cabal
```diff
@@ -0,0 +1,70 @@
...
+version:       0.1
+license:       BSD2
+cabal-version: >= 1.10
+license-file:  LICENSE
+author:        Edward A. Kmett
+maintainer:    Edward A. Kmett <ekmett@gmail.com>
+stability:     provisional
+homepage:      http://github.com/ekmett/lens/
+bug-reports:   http://github.com/ekmett/lens/issues
+copyright:     Copyright (C) 2012-2016 Edward A. Kmett
```
RyanGlScott: Similar comments here about copyright.

lens-partial-iso/lens-partial-iso.cabal
```diff
@@ -0,0 +1,70 @@
+name:          lens-partial-iso
```
RyanGlScott, line 1: I think you'll need to update the CI-related `.yml` files under `.github/workflows` to account for this new package.

lens-partial-iso/src/Control/Lens/PartialIso.hs
jozanek: done

lens-partial-iso/LICENSE
jozanek: done

lens-partial-iso/src/Control/Lens/PartialIso.hs
jozanek: udpated

lens-partial-iso/src/Control/Lens/PartialIso.hs
jozanek: I went with match everywhere

lens-partial-iso/src/Control/Lens/PartialIso.hs
jozanek: Dropped the `mapMaybe`/`catMaybes` re-export, better to leave `Witherable` for users to import.

lens-partial-iso/src/Control/Lens/PartialIso.hs
jozanek: applied

lens-partial-iso/lens-partial-iso.cabal
jozanek: done

lens-partial-iso/lens-partial-iso.cabal
jozanek: I couldn't regenerate `haskell-ci.yml`: haskell-ci doesn't support GHC 9.14 yet (haskell-CI/haskell-ci#818, blocked on `lukko`/`zinza`/`generic-lens-lite`), and the released 0.18.1 caps `base <4.20` so it won't build on my toolchain. Could you regenerate with your pinned generator? I can try to dig into haskell-ci issue in the meantime.

RyanGlScott: I've regenerated the CI.

RyanGlScott (approved):

PR #1120, opened 2026-06-15
Make Control.Lens.Profunctor's from* conversions take the canonical monomorphic optics

### LINKED ISSUE #971: Inconsistency in Control.Lens.Profunctor
`fromLens`, `fromIso`, and `fromPrism` basically require full-fledged `Lens`es, `Iso`s, and `Prism`s. `fromSetter`, on the other hand, will work with `ASetter`, and `fromTraversal` will work with `ATraversal`. From an API standpoint, it's all a bit confusing. From an implementation standpoint, `fromIso` and `fromPrism` are basically `id` (using `#.` appropriately, they could be made actual no-ops), while the rest are not.

So...

1. We should have versions of each of these that take the canonical monomorphic versions.
2. We should make it clear what's doing work.

jozanek:
Closes #971.

`fromSetter` and `fromTraversal` already take the canonical monomorphic optics (`ASetter`/`ATraversal`), but `fromLens`/`fromIso`/`fromPrism` took ad-hoc representations:

```haskell
-- before
fromLens  :: Strong p     => LensLike (Context a b) s t a b -> OpticP p s t a b
fromIso   :: Profunctor p => Optic p Identity s t a b       -> OpticP p s t a b
fromPrism :: Choice p     => Optic p Identity s t a b       -> OpticP p s t a b

-- after
fromLens  :: Strong p     => ALens  s t a b -> OpticP p s t a b
fromIso   :: Profunctor p => AnIso  s t a b -> OpticP p s t a b
fromPrism :: Choice p     => APrism s t a b -> OpticP p s t a b
```

All five `from*` are now uniform.

### Notes

- Backwards compatibility: full polymorphic `Lens`/`Iso`/`Prism` values still typecheck against `ALens`/`AnIso`/`APrism` (they're instantiations), so the common case is unaffected. Only code that passed the previous ad-hoc representations directly needs adjusting.
- `fromLens` stays single-pass: it's reimplemented over `ALens` via `runPretext (k sell s) (Context id)`, pulling the focus and rebuild function out of one traversal — matching the old `Context`-based implementation rather than routing through `withLens`, whose `^#`/`storing` would each re-traverse. `fromIso`/`fromPrism` use `withIso`/`withPrism`, which already return plain functions.
- Docs: dropped the `@ … @` pseudo-signature blocks across the module — they referenced undefined `LensP`/`IsoP`/… names; the now-clean real signatures plus added doctests are self-documenting.
- Tests: added HUnit cases exercising all five `from*` at the function profunctor `(->)`, with inputs bound to explicit `A`-types so the three changed conversions are guarded against regressing to the old signatures.

This is PVP-breaking (changed exported signatures); folded into 5.3.7 for now — let me know if you'd prefer a 5.4 bump.

### DIFF
--- a/src/Control/Lens/Profunctor.hs
+++ b/src/Control/Lens/Profunctor.hs
@@ -30,9 +30,12 @@ import Prelude ()
 
 import Control.Lens.Internal.Prelude
 import Control.Lens.Type (Optic, LensLike)
-import Control.Lens.Internal.Context (Context (..), sell)
+import Control.Lens.Internal.Context (Context (..), runPretext, sell)
 import Control.Lens.Internal.Profunctor (WrappedPafb (..))
-import Control.Lens (ASetter, ATraversal, cloneTraversal, Settable)
+import Control.Lens
+  ( ALens, AnIso, APrism, ASetter, ATraversal
+  , cloneTraversal, withIso, withPrism, Settable
+  )
 import Data.Profunctor (Star (..))
 import Data.Profunctor.Mapping (Mapping (..))
 import Data.Profunctor.Traversing (Traversing (..))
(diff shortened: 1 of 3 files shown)


src/Control/Lens/Profunctor.hs
```diff
@@ -30,63 +30,71 @@ import Prelude ()
...
-fromLens :: Strong p => LensLike (Context a b) s t a b -> OpticP p s t a b
-fromLens l p =
-  dimap
-    (\s -> let Context f a = l sell s in (f, a))
-    (uncurry id)
-    (second' p)
-
--- | Converts a 'Control.Lens.Type.Iso' to a 'Profunctor'-based one.
+-- The result is a 'Strong' profunctor optic: 'second'' carries the rebuild
+-- function alongside the focus, so a single @k 'sell' s@ yields both the focus
```
RyanGlScott: I think mentioning [`sell`](https://hackage-content.haskell.org/package/lens-5.3.6/docs/Control-Lens-Internal-Context.html#v:sell) here in the Haddocks might be a bit distracting, as that mentions implementation details that are somewhat internal.

src/Control/Lens/Profunctor.hs
```diff
@@ -30,63 +30,71 @@ import Prelude ()
...
+fromLens :: Strong p => ALens s t a b -> OpticP p s t a b
+fromLens k p = dimap project (uncurry id) (second' p)
+  where
+    project s = case runPretext (k sell s) (Context id) of
+      Context bt a -> (bt, a)
+
+-- | Converts an 'Control.Lens.Type.Iso' to a 'Profunctor'-based one.
+--
+-- An 'Control.Lens.Type.Iso' is just a pair of mutually inverse functions, so
+-- this is exactly a 'dimap' — it adds no profunctor structure of its own.
```
RyanGlScott: `dimap` isn't internal to `lens`, but again, I feel like this is talking about the implementation details, which don't really belong in the Haddocks.

src/Control/Lens/Profunctor.hs
```diff
@@ -30,63 +30,71 @@ import Prelude ()
...
 
 -- | Converts a 'Control.Lens.Type.Prism' to a 'Profunctor'-based one.
 --
--- @
--- 'fromPrism' :: 'Control.Lens.Type.Prism' s t a b -> PrismP s t a b
--- @
-fromPrism :: Choice p => Optic p Identity s t a b -> OpticP p s t a b
-fromPrism p pab = rmap runIdentity (p (rmap Identity pab))
+-- The result is a 'Choice' profunctor optic: 'right'' runs the profunctor only
+-- on the matching branch, passing a non-match through to @t@ untouched.
```
RyanGlScott: `right'` isn't internal to `lens`, but again, I feel like this is talking about the implementation details, which don't really belong in the Haddocks.

RyanGlScott (commented):
> This is PVP-breaking (changed exported signatures); folded into 5.3.7 for now — let me know if you'd prefer a 5.4 bump.

Let's go ahead and bump the version to 5.4.

src/Control/Lens/Profunctor.hs
jozanek: updated

src/Control/Lens/Profunctor.hs
jozanek: done

src/Control/Lens/Profunctor.hs
jozanek: changed

RyanGlScott (approved):

PR #1122, opened 2026-06-16
Extract Plated into standalone `plated` and `plated-instances` packages (#993)

### LINKED ISSUE #993: Separate Plated into own package
- `Plated` class isn't `lens` specific.
- `optics` (and other lens libraries, e.g. `microlens` users) would benefit from having just one `Plated`
- this would allow to inverse dependency so `lens-aeson` wouldn't need to have orphan `Plated`, but it could be provided by `aeson` `itself.
- This is similar to splitting out indexed traversables

jozanek:
Closes #993.

`Plated` isn't `lens`-specific, and downstream type owners (aeson/optics/microlens ecosystems) currently can't provide `Plated` instances without depending on all of `lens` or shipping orphans. This extracts the class into small packages, mirroring the `indexed-traversable` / `indexed-traversable-instances` split.

## Package layout

- **`plated`** (depends on `base` + `containers`) — the `Plated` class with its `Data`-based default, the `[]` and `Data.Tree.Tree` instances, and the `Generic`-based `gplate`/`gplate1` (`GPlated`/`GPlated1`). `containers` is a GHC boot library, so this is effectively a `base`-only footprint, and it lets the class carry boot-library instances the way `indexed-traversable` does.
- **`plated-instances`** (`plated` + `free` + `comonad`) — the orphan `Plated` instances for the `free`/`comonad` type families (`Free`/`FreeT`/`F`/`Cofree`/`CofreeT`), kept separate so depending on the class doesn't pull in `free`.
- **`lens`** — `Control.Lens.Plated` now re-exports the class + generics from `plated`, pulls the instances in via `Data.Plated.Instances`, and keeps all the combinators (`rewrite`/`transform`/`universe`/…) plus the Template Haskell instances (as `plate = uniplate`, using lens's own optimized `uniplate`).

## Backwards compatibility

`import Control.Lens` / `import Control.Lens.Plated` code is unchanged, **including** empty `instance Plated T` declarations that rely on the `Data` default.

The one behavioral nuance: `plated`'s default `plate` is backed by a portable, `base`-only `uniplate` rather than lens's oracle-optimized `Data.Data.Lens.uniplate`. The two compute the **same** traversal (verified against each other on nested cases — lists, `Maybe`, tuples, newtypes — via `children`, `universe`, and `transform`); the portable one can just be slower on deeply nested non-matching structure. lens's own instances keep using the optimized version, and users can write `plate = Data.Data.Lens.uniplate` explicitly where it matters.

## Open questions for review

- **Package/module names** (`plated`, `plated-instances`; `Data.Plated`, `Data.Plated.Instances`) and the **initial version** (`0.1`) are bikesheds — happy to change.
- **Target-name collision:** `lens` already has a benchmark component named `plated`, so a bare `cabal build plated` is ambiguous (use `cabal build plated:lib:plated`). Might be worth renaming the benchmark.
- **CI:** `.github/workflows/haskell-ci.yml` will need regenerating (`haskell-ci`) to cover the two new packages.
- **Hackage publishing** of `plated`/`plated-instances`, and downstream adoption (e.g. `free`/`containers` providing their own instances), are follow-ups beyond this in-tree carve-out.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,21 @@
 next [????.??.??]
 -----------------
+* Move the `Plated` class, the `[]`/`Tree` instances, and the `Generic`-based
+  `gplate`/`gplate1` (with the `GPlated`/`GPlated1` classes) into a new `plated`
+  package (depending only on `base` and `containers`), and the `Plated`
+  instances for the `free`/`comonad` type families into a companion
+  `plated-instances` package. Both are re-exported from `Control.Lens.Plated`,
+  so existing `import Control.Lens` / `import Control.Lens.Plated` code keeps
+  working, including empty `instance Plated T` declarations that rely on the
+  `Data`-based default. The split lets other libraries provide `Plated`
+  instances for their types depending only on `base` and `containers` (a GHC
+  boot library), instead of all of `lens`.
+  One behavioral note: the `Data`-based default for `plate` is now backed by a
+  portable `uniplate` defined in `plated` rather than the oracle-optimized
+  `Data.Data.Lens.uniplate`. The two compute the same traversal, but the
+  `plated` default can be slower on deeply nested non-matching structure; write
+  `plate = uniplate` (from `Data.Data.Lens`) explicitly to keep the optimized
+  version where it matters.
 * Add `ReifiedReview` to `Control.Lens.Reified`.
 * Reduce the arity of `set`, `set'`, `partsOf`, `partsOf'`, `unsafePartsOf`,
   `unsafePartsOf'`, `mapAccumLOf`, `imapAccumLOf`, `auf`, `both`, and `both1` so
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,5 @@
 packages: .
           ./examples
           ./lens-properties
+          ./plated
+          ./plated-instances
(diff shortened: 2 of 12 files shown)


phadej:
Wow. Interesting to see someone picking this up. When I looked at his long time ago (2021), I dropped the idea because I concluded that `Plated` class just doesn't feel right.

For example consider `Exp`, `Pat`, `Dec`, `Stmt` etc types from `template-haskell`. These types are mutually-recursive. `Plated` defined using `Data` / `uniplate` would allow to look into all `Exp`s inside an `Exp`, `Dec`s inside `Dec` and so forth, even if there are other types in between (e.g. `Exp` inside pattern guard (`Pat`) inside the `Exp`). The `GHC.Generic` derived instances wouldn't be able to do that. And if you would try to write `uniplate`-like instances by hand, you wouldn't be able to use `Plated` instances. You would need functions with types `Traversal' Exp Pat`, `Traversal' Pat Exp` etc. And actually these functions would be quite useful standalone. Sometimes you want to look into all patterns inside an expressions etc.

So TL;DR, I don't think `Plated` is a right thing, and thus I don't see using it. It's fine to exist inside `lens`, and sometimes it may be handy. 

Sorry, I should have commented on my findings in #993 long ago, but as there weren't any discussion I forgot about whole issue.

jozanek:
@phadej fair point. Want me to close this, or keep it as a compat-only extraction?

RyanGlScott:
In my experience, these sorts of extract-a-class-into-another-package changes inevitable give rise to some code churn downstream. As such, I wouldn't want to do this unless we're enthusiastic about the change, and it doesn't sound like we're there.

PR #1124, opened 2026-06-16
Clarify outside Haddock with function-specialised type

jozanek:
Notes the `p ~ (->)` specialisation of `outside`'s `Representable` type in its Haddock. Comment-only.

### DIFF
--- a/src/Control/Lens/Prism.hs
+++ b/src/Control/Lens/Prism.hs
@@ -137,6 +137,9 @@ prism' bs sma = prism bs (\s -> maybe (Left s) Right (sma s))
 
 -- | Use a 'Prism' as a kind of first-class pattern.
 --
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks, but I don't think it's fair to say that this addresses https://github.com/ekmett/lens/issues/1078. The main concern of https://github.com/ekmett/lens/issues/1078 was that the type in the _diagram_ was wrong. In that sense, it's probably more accurate to say that https://github.com/ekmett/lens/pull/1117 is the change that addresses https://github.com/ekmett/lens/issues/1078.

The clarification to the Haddocks is nice, however. If you remove the "addresses #1078" part from the PR description, then I think we could land this.

jozanek:
@RyanGlScott descritpion updated

RyanGlScott (approved):

PR #1130, opened 2026-07-25
Add fromMaybeOf combinator

### LINKED ISSUE #822: add fromMaybeOf combinator
Would it be possible to add a `fromMaybeOf` combinator?  This would generalize `fromMaybe` to work on any `Traversal`.  Or does this already exist in some form?

I imagine it would be implemented like this:

```haskell
import Control.Lens
import Data.Maybe (fromMaybe)
import Data.Monoid (First)

fromMaybeOf :: Getting (First a) s a -> a -> s -> a
fromMaybeOf trav a s = fromMaybe a (s ^? trav)
```

`fromMaybe` could then be implemented like this:

```haskell
fromMaybe :: a -> Maybe a -> a
fromMaybe = fromMaybeOf traverse
```

Here are two examples of using it in GHCi:

```haskell
> fromMaybeOf (_1 . traverse . traverse) "foobar" ([Nothing, Just "hello", Just "bye"], 3)
"hello"
> fromMaybeOf (_1 . traverse . traverse) "foobar" ([Nothing, Nothing, Nothing], 3)
"foobar"
```

Here are some possible reasons not to add it:

- The function name `fromMaybeOf` is a misnomer, since it works on any `Traversal`, instead of just on `Maybe`.
- It might not pass the Fairbairn threshold, since the definition of `fromMaybeOf` is so short.

Here are some reasons to add it:

- I've wanted this function in the past.  Yesterday someone else on the FP Slack was also asking about it.
- Based on the name, it should be immediately recognizable to Haskellers.  I'm assuming most people are already familiar with `fromMaybe`, and `fromMaybeOf` is similar to the other `*Of` functions in lens.
- Maybe its name could be changed to something like `fromTraverseOf`?  (Although this would invalidate the previous point.)

If this `fromMaybeOf` function would be accepted into lens, I would be willing to send a PR adding it as well as documentation.

jozanek:
Closes #822.


RyanGlScott (approved):
Thanks!

PR #1131, opened 2026-07-26
Make TH-generated optics inherit field and constructor Haddocks

### LINKED ISSUE #614: Lenses created by makeLenses should inherit Haddock documentation
Currently the created lenses don't have any documentation, so if we _only_ export the lens accessors for a datatype and hide the constructors, we lose all documentation about the datatype.

I don't know if this is possible with current Haddock, however.

jozanek:
Fixes #614.

On GHC ≥ 9.2 with `-haddock`, optics generated by `makeLenses`, `makeClassy`, `makePrisms`, `makeClassyPrisms`, etc. inherit the Haddock docs of the field/constructor they focus, via a module finalizer using template-haskell 2.18's `getDoc`/`putDoc` (CPP-gated; older GHCs unchanged). Existing documentation is never overwritten.

Inheriting nothing, deliberately: definitions merging several fields, shared field-class methods (`makeFields` & co.), and the quote-based `declare*` variants (GHC drops docs in quotes). Prism support is a small intentional extra beyond the issue's ask — easy to drop if unwanted.


src/Control/Lens/TH.hs
```diff
@@ -422,13 +437,16 @@ makeLensesWith = makeFieldOptics
 -- data Foo = Foo 'Int' 'Int' deriving 'Show'
 -- fooX, fooY :: 'Lens'' Foo Int
 -- @
+--
+-- Cannot inherit field documentation; see 'declareLensesWith'.
```
RyanGlScott: When I first read this comment, I interpreted it as "see `declareLensesWith` for a version of the function that does support inheriting field documentation". A closer inspection reveals that this really means "see the Haddocks for `declareLensesWith` for an explanation of why field documentation can't be inherited". I think this explanation would be clearer---can you change this (and the Haddocks for related functions below) to look more like the latter?

src/Control/Lens/Internal/PrismTH.hs
```diff
@@ -447,19 +456,21 @@ makeClassyPrismClass ::
...
+     methodss <- traverse (mkMethod r) (zip cons cons')
      classD (cxt[]) className (D.plainTV r : vs) (fds r)
        ( sigD methodName methodType
        : map return (concat methodss)
        )
 
   where
-  mkMethod r con =
+  -- srcCon carries the constructor's Name; con has it renamed to the method's
+  mkMethod r (srcCon, con) =
```
RyanGlScott: Do we really need to pass both `srcCon` and `con` here? What if we just passed `srcCon` and then changed the definition of `defName` to something like `defName = prismName (view nconName srcCon)`?

src/Control/Lens/TH.hs
```diff
@@ -141,7 +141,8 @@ simpleLenses f r = fmap (\x -> r { _simpleLenses = x}) (f (_simpleLenses r))
 -- lenses.
 --
 -- Disabling this can be useful if you want to provide a more restricted type
--- signature or if you want to supply hand-written haddocks.
+-- signature or if you want to supply hand-written haddocks, which are
+-- never overwritten by inherited field documentation (see 'makeLenses').
```
RyanGlScott: Is the second part of this sentence still relevant? It claims that you might want to disable generating a type signature if you want to supply hand-written Haddocks, but it appears that you can just hand-write Haddocks regardless.

RyanGlScott (commented):
Sorry for the wait on this one!

src/Control/Lens/TH.hs
jozanek: Good catch. Changed all six to `Cannot inherit field documentation; 'declareLensesWith' explains why.`

src/Control/Lens/Internal/PrismTH.hs
jozanek: Done. Also drops `cons'` and the `zip`.

Note `computeOpticType` now gets the real constructor, so its `List.delete` actually deletes. That only changes `s`/`t`/`a`, which we discard. `-ddump-splices` is unchanged.

Also removed the `Control.Lens.Setter` import, since `over` was its last use.

src/Control/Lens/TH.hs
jozanek: Tested: a doc comment before the splice attaches to only one lens (with `_mOne`/`_mTwo` it hit `mTwo`), and writing your own signature is a duplicate-signature error.

So it's still needed for documenting a single optic. Reworded to "hand-written haddocks for individual optics", happy to drop it instead.

RyanGlScott (approved):

Issue #799, opened 2018-03-22
makeFields instances violate coverage condition


bergey:
`makeFields` generates instances with `TF foo ~ a` constraints for data families.  These require UndecidableInstances, but the more obvious version works without that extension.

I'm using `lens-4.15.4` and GHC 8.2.  I haven't yet tested with other versions.  

I suspect this behavior was introduced in https://github.com/ekmett/lens/issues/754 (Based on reading that thread, not trying earlier versions of lens.) Maybe it's possible to distinguish type families from data families, and be less eager to use the equality constraint?  I'm at the limit of my knowledge of TH, unfortunately.

Short example below.  In my actual use, I have the `Key` family from `persistent`, as the data family causing me trouble.

```
module Experiment where

class HasFoo s a | s -> a where
    foo :: s -> a

data family DF a
newtype instance DF Int = FooInt Int

data Bar = Bar {
    _foo :: DF Int
    }

-- | Type checks
instance HasFoo Bar (DF Int) where
    foo = _foo

-- | Does not typecheck without UndecidableInstances

-- instance a ~ DF Int => HasFoo Bar a where
--     foo = _foo

--     • Illegal instance declaration for ‘HasFoo Bar a’
--         The coverage condition fails in class ‘HasFoo’
--           for functional dependency: ‘s -> a’
--         Reason: lhs type ‘Bar’ does not determine rhs type ‘a’
--         Un-determined variable: a
--         Using UndecidableInstances might help
--     • In the instance declaration for ‘HasFoo Bar a’
--    |
-- 20 | instance a ~ Foo Int => HasFoo Bar a where

```

RyanGlScott:
Yes, good catch. Indeed, the code which currently decides whether to emit an equality constraint or not:

https://github.com/ekmett/lens/blob/9d1d2e6236c06dd48fadbf199e871b1f64e61123/src/Control/Lens/Internal/FieldTH.hs#L386-L389

Does not distinguish between type families and data families. We really should exclude data families, just as we exclude normal data types.

bergey:
Thank you for the impressively fast turnaround on that patch & test!

RyanGlScott:
I've uploaded [`lens-4.16.1`](http://hackage.haskell.org/package/lens-4.16.1) to Hackage with a fix.

PR #1132, opened 2026-09-07
Add makeConstructors: overloaded constructor prisms

### LINKED ISSUE #934: Have prisms with ad-hoc classes for every field (equivalent to how `makeFields` generates lenses)
Perhaps this is not possible for some reason (or I'm blind), but would it make sense to have prisms generated with ad-hoc classes for every field -- similar to how `makeFields` does it?

I.e right now if two data types (with multiple data constructors for prisms to make sense) have identically named data constructors then the prisms need to be generated in different namespaces and also qualified at use sites. (The `makeClassyPrisms` makes a class per set of prisms, but that doesn't fit for when only some data constructor names match.)

jozanek:
Fixes #934.

`makeConstructors ''T` makes, for each constructor `Con`, `class AsCon s a | s -> a` with method `_Con :: Prism' s a`, and `T`'s instance of it. No top-level optics.

It only declares the class if no class of that name is in scope. So two types with a constructor of the same name share one class — the same way `makeFields` shares its `Has*` classes.

Some things the issue didn't settle. I picked these; happy to change any:

- **Existential constructors get nothing.** The payload type isn't fixed by the type, so the functional dependency's coverage condition rejects the instance — the liberal one too.
- **Operator-named constructors get nothing.** There is no way to spell `AsCon` for them. Operator-named types are fine.
- **A lone constructor gets a `Prism'`**, not the `Iso` `makePrisms` would make. The class method is always a simple `Prism'`; `makePrisms` remains the way to get the `Iso` or a type-changing `Prism`.
- **Payloads with a type family** go behind an equality constraint, as in #799. Splices like that need `UndecidableInstances`.
- **`makeClassInstance` sits in `FieldTH`** (commit 1). `Internal.TH` would be a cycle through `Language.Haskell.TH.Lens`.
- **The name `makeConstructors`** is the one from the issue thread.

No `makeConstructorsWith` or `declareConstructors` yet; see the review discussion.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -74,6 +74,11 @@ next [????.??.??]
   visibly the `Review` box, which still used the long-removed four-parameter
   form — and add `images/Hierarchy.dot` as an editable Graphviz source for the
   diagram (regenerated via `scripts/hierarchy`).
+* Add `makeConstructors` to `Control.Lens.TH`: the `makePrisms` counterpart of
+  `makeFields`. Each constructor `Con` of a type `T` gets the optic `makePrisms`
+  would generate, as `_TCon`, and an instance of a per-constructor class
+  `AsCon s a | s -> a` with method `_Con :: Prism' s a`; the class is declared
+  only if not already in scope, so types with same-named constructors share it.
 
 5.3.6 [2026.01.10]
 ------------------
--- a/lens.cabal
+++ b/lens.cabal
@@ -337,6 +337,7 @@ test-suite templates
     BigRecord
     T799
     T917
+    T934
     T972
   if impl(ghc >= 9.2)
     other-modules:
--- a/src/Control/Lens/Internal/FieldTH.hs
+++ b/src/Control/Lens/Internal/FieldTH.hs
@@ -33,6 +33,7 @@ module Control.Lens.Internal.FieldTH
   , makeFieldOpticsForDec
   , makeFieldOpticsForDec'
   , makeFieldOpticExp
+  , makeClassInstance
   , HasFieldClasses
   ) where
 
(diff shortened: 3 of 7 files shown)


RyanGlScott:
Thanks for working on this, @jozanek! I haven't had a chance to review the implementation itself thoroughly, but here are some thoughts on the overall idea:

* I think most of the design choices are reasonable.
* One design choice I think we should change: I would prefer to just generate the classes and their instances, without the top-level functions. That is, generate `AsCon` with a `_Con` method, but don't generate a `_TCon` function. The latter is subsumed by the former, but more importantly, generating fewer definitions keeps the code simpler and provides fewer opportunities for name clashes.
* Have you verified that if you call `makeConstructors` on two different data types with the same constructor names, then only one class definition will be generated? For instance, if you call `makeFields` twice, e.g.,

  ```hs
  data T1 = MkT1 { _t1X :: Int, _t1Y :: Int }
  data T2 = MkT2 { _t2X :: Int, _t2Y :: Int }

  $(makeFields ''T1)
  $(makeFields ''T2)
  ```

  Then the TH machinery is smart enough to recognize that only the first `$(makeFields ...)` declaration needs to generate `Has{X,Y}` class definitions. I am envisioning a similar scenario with `makeConstructors` like [this one](https://github.com/ekmett/lens/issues/934#issuecomment-663488542):

  ```hs
  module Other where

  data Type2
    = One Double Int String
    | More Int
  ```
  ```hs
  module Main where

  import qualified Other

  data Type1
    = One Int String
    | Two String

  $(makeConstructors ''Type1)
  $(makeConstructors ''Other.Type2)
  ```

  We'd want to make sure that only the first `$(makeConstructors ...)` declaration generates an `AsOne` class declaration.
* As you've noted, we'll definitely want a `makeConstructorsWith` function (which takes a `LensRules` argument) for symmetry with the rest of the TH machinery. Arguably, we'd also want `declareConstructors` and `declareConstructorsWith` as well to allow users to supply TH `Dec`s instead of `Name`s as arguments. This might require some non-trivial amount of additional work, however, so I'd be happy to defer these tasks to separate issues if you don't want to address them as part of this PR.

jozanek:
Thanks!

Dropped the top-level optics; only classes and instances now. Description updated.

Sharing works with your exact example, and it's now a test.

`makeConstructorsWith` isn't symmetry: `makePrisms` takes no `LensRules`, and none of its fields fit constructors, so it needs a new rules type. `declareConstructors` has its own wrinkle with classes declared in the same quote. I'd leave both to a follow-up issue if there's demand.

